PR #7, opened 2026-08-19
Plan and begin realtime browser architecture

yongkyuns:
## Scope

This draft PR is the implementation branch for Noon's realtime/browser redesign.

The architecture is a renderer-independent semantic/compiler/runtime pipeline with a persistent WebGPU/browser player and Python/Pyodide as an optional authoring/control plane:

`Python/Rust/JS -> SceneDefinition/ScenePatch -> CompiledScene -> SceneInstance -> FrameState -> FramePreparer -> wgpu/WebGPU`

## Core principles

- Python-first ergonomics without Python in the normal frame-critical path
- language-neutral scene and patch representation
- stable semantic IDs externally, dense runtime indices internally
- deterministic direct seek and forward playback
- cached/precompiled geometry work rather than per-frame tessellation
- structural/numerical correctness tests first; visual regression at the renderer boundary
- strict legacy Noon/Manim API compatibility is not required

## Current implementation

Implemented foundations include:

- stable semantic objects, styles, transforms, paths, tracks, and ordered patches
- deterministic compiler/runtime with incremental dirty tracking and playhead-preserving live reconciliation
- atomic full-scene reconciliation for arbitrary Python source reruns; structural changes compile/rebuild/seek once rather than once per patch
- analytic instanced circles/rectangles/lines plus cached Lyon vector paths
- WebGPU canvas playback, profiling, release wasm packaging, and Python/Pyodide worker authoring
- path reveal with precomputed arc-length metadata
- deterministic path correspondence with feature-preserving resampling
- fixed-topology GPU path morphing with no steady retessellation
- Manim-like `scene.play(Transform(...))` authoring lowered to a first-class atomic language-neutral Transform track
- same-kind analytic geometry Transform for circle radius, rectangle size, and line endpoints with no path conversion or tessellation
- shared semantic stroke join/cap policy across static Lyon paths and fixed-topology path Transform: `round` / `miter` / `bevel` joins and `round` / `butt` / `square` caps
- bounded fixed-topology filled-path Transform for one continuously certifiable closed contour
- renderer cache identity/correctness coverage including fill presence, stroke width, join, and cap topology

### Generic Transform

`Transform` is represented as one semantic track:

```text
Property::Transform
TrackValues::Object { from: ObjectSnapshot, to: ObjectSnapshot }
```

Detached target objects carry geometry/transform/style but no scene identity. The source `ObjectId` remains stable.

The compiler selects an explicit geometry plan:

```text
Static
Circle { from_radius, to_radius }
Rectangle { from_size, to_size }
Line { from_start, from_end, to_start, to_end }
PathPair(prepared_geometry)
```

Circle/rectangle/line geometry interpolates directly in semantic `FrameObjectState.geometry`; the analytic renderer consumes those values through packed instance records with no path conversion or tessellation.

Path->path geometry changes compile to one fixed source/target render pair. Semantic `FrameObjectState.geometry` remains separate from renderer-prepared morph geometry and reaches exact source/target endpoints. Steady path Transform frames reuse cached geometry and dirty only instance state.

### Stroke join/cap fidelity

Stroke policy is semantic and shared by static and morph paths:

- `StrokeJoin::{Round, Miter, Bevel}`
- `StrokeCap::{Round, Butt, Square}`
- older serialized styles remain compatible because both fields serde-default to round
- Python exposes and validates `stroke_join=` / `stroke_cap=`
- static paths lower directly into Lyon join/cap options
- morph paths use deterministic fixed topology: segment quads plus fixed join/cap primitives
- both left and right join slots exist and the inactive side collapses, so turn direction may change during morphing without changing topology
- path cache keys include stroke width, join, cap, and fill presence

### Bounded filled-path Transform

Filled path Transform is supported for a deliberately bounded class rather than falling back to per-frame triangulation.

The supported contract requires:

1. exactly one closed contour at each endpoint
2. simple, nondegenerate resampled endpoint contours
3. fill presence unchanged across the geometry-changing Transform
4. stroke width/join/cap topology unchanged
5. one center-fan triangulation that remains valid for the complete interpolation interval

The planner reuses deterministic feature-preserving correspondence, canonicalizes winding, cyclically aligns the target without reversing winding, computes endpoint area centroids, and builds a fixed center-fan topology. Every fan triangle is certified over the full continuous interpolation interval: signed triangle area is quadratic in time, so the exact endpoints and any interior critical point are checked. Unsafe/self-intersecting/non-certifiable cases are rejected with an explicit compile error.

This supports useful concave/star-shaped cases, including rounded-loop -> concave-star morphs, while intentionally not claiming arbitrary polygon triangulation. Fill and stroke share one cached prepared path mesh and one morph progress value; there is no per-frame fill correspondence, triangulation, or tessellation.

Current explicit safety boundaries remain:

- general filled polygons outside the certified center-fan class
- multiple filled contours / holes
- fill-presence changes during a geometry-changing path Transform
- path stroke-width changes during geometry-changing Transform
- join/cap topology changes during geometry-changing Transform
- unsupported cross-kind geometry Transform

The browser playground includes stroke path, filled path, and analytic primitive Transform examples.

See `docs/generic-transform.md` for the detailed contract, proofs/validation strategy, limitations, and continuation plan.

## Validation discipline

The normal draft-PR CI is read-only and runs:

```bash
cargo fmt --all -- --check
cargo check --workspace --all-targets --all-features
cargo clippy --workspace --all-targets --all-features -- -D warnings
cargo test -p noon-geometry --test tessellation_correctness
cargo test -p noon-geometry --test stroke_style_parity
cargo test -p noon-geometry --test filled_morph
cargo test -p noon-geometry --test filled_morph_interval
cargo test --workspace --all-features
cargo check -p noon-render-wgpu --target wasm32-unknown-unknown --no-default-features --features web
cargo check -p noon-web --target wasm32-unknown-unknown
bash scripts/build-web-demo.sh
```

Filled-path verification includes endpoint topology/area checks, fill-only and fill+stroke meshes, unsafe/self-intersecting/open-contour rejection, a regression where both endpoints are valid but a fan triangle would invert only inside the animation interval, direct-seek/forward runtime parity, renderer cache identity, cold geometry reuse, and instance-only steady morph updates.

The final repository state uses read-only CI and contains no temporary migration/self-patching scripts.

## Immediate next milestone

Move to **Transform lifecycle semantics** without weakening the current geometry guarantees:

1. `ReplacementTransform`
2. `TransformFromCopy`
3. define lifecycle/identity semantics explicitly in the language-neutral IR/runtime
4. add direct-seek/forward and live-reconciliation tests for object creation/removal/replacement behavior
5. then build matching-shape correspondence on top of those lifecycle primitives

Cross-kind geometry interpolation, general filled polygons/holes, and path stroke-width interpolation remain separate future milestones.

This PR remains draft and must not be merged to `master` without explicit authorization.


yongkyuns:
Implementation is intentionally paused before Milestone 1 code because the required GitHub Actions gate is not starting. The CI workflow has been landed on `master` via PR #8, and a subsequent feature-branch commit was pushed, but GitHub reports no workflow run/check for the PR head. Per `docs/implementation-plan.md`, architectural code will not advance until the CI gate actually executes and is green.

Issue #176, opened 2026-08-25
[Parity V0] Make ManimCE v0.21.0 the canonical raster + timeline oracle for supported examples


yongkyuns:
## Goal
Build the missing end-to-end differential lane that renders the **same supported scene** through pinned ManimCE v0.21.0 and Noon, then compares both timeline samples and pixels.

This is the bridge between #57 (semantic Manim differential tests) and #109 (Noon renderer goldens). It should **reuse** those harnesses rather than replacing either one.

## Why this is needed
The current Manim tutorial gallery is not suitable as a visual oracle. `web/python/examples/manim_tutorial_manifest.json` explicitly describes the ready entries as `original-noon-adaptation`, and those adaptations materially change values that affect output.

For example, upstream ManimCE quickstart uses essentially:
```py
circle = Circle()
circle.set_fill(PINK, opacity=0.5)
self.play(Create(circle))
```
while Noon currently uses:
```py
Circle(radius=1.0, color=BLUE).set_fill(BLUE, opacity=0.35).set_stroke(WHITE, 0.06)
scene.play(..., run_time=1.0)
scene.wait(0.25)
```
Likewise the current Noon square-to-circle adaptation changes dimensions/timing and omits upstream rotation/FadeOut. A screenshot difference between these programs cannot tell us whether Noon is correct.

Reference: https://docs.manim.community/en/stable/tutorials/quickstart.html

## Scope
- Pin ManimCE **0.21.0** exactly, matching #57.
- Add a small canonical corpus whose logical source/options are identical between Manim and Noon, with only the minimum import/Scene wrapper adaptation required by the frontend.
- Start with the supported quickstart subset: CreateCircle, SquareToCircle, SquareAndCircle/positioning, AnimatedSquareToCircle, DifferentRotations where supported.
- Define one deterministic render profile: pixel dimensions, frame width/height, frame rate, renderer (prefer Cairo as the initial reference), background, transparency, and deterministic fonts/resources where applicable.
- Render Manim reference frames at exact semantic times, not by comparing unrelated encoded-video frame numbers.
- Render Noon at the exact same semantic times through the browser renderer.
- Produce `reference / noon / diff` artifacts plus machine-readable measurements for every sampled time.
- Compare semantic observables (center/bounds/style/scene membership/timing) alongside raster output so failures identify whether the problem is authoring semantics, timeline interpolation, camera mapping, or rasterization.
- Keep explicit `unsupported` entries; do not normalize current Noon output into the reference.

[...]

Issue #81, opened 2026-08-24
[Roadmap B3/C1] Indication, tracing, updater, number, and speed-modifier animations


yongkyuns:
## Roadmap

B3 deterministic/native animation breadth plus C1 host-dynamic cases under #954/#955. Architecture authority: `docs/architecture.md`.

## Goal

Cover indication, tracing, updater-driven, numeric and speed-modifier helpers while classifying each behavior into deterministic timeline, native reactive or host-dynamic execution—without forcing every live update through authored semantic mutation.

## Execution contract

- deterministic effects lower from semantic animation intent;
- native-reactive effects write effective Runtime state through the shared dependency graph;
- arbitrary callbacks use C1/#70's coherent staged result contract: ordinary live property/style output is an effective driver write, while persistent content/structure edits use semantic mutations;
- transient dynamic content/resource output uses C4/#368's versioned effective lane where appropriate;
- persistent final results reconcile to authored/base state only when semantics require it.

A host-dynamic object must not make unrelated deterministic/native content host-driven.

### Deterministic/native where expressible
- Indicate/FocusOn/Flash/Circumscribe;
- ShowPassingFlash/Wiggle/ApplyWave and similar deterministic helpers;
- deterministic portions of TracedPath/AnimatedBoundary;
- ChangeSpeed/time remapping;
- numeric animations after B4 numeric text exists.

### Host-dynamic when arbitrary user code is required
- UpdateFromFunc/UpdateFromAlphaFunc and equivalent arbitrary callable semantics;
- updater/tracing cases whose behavior cannot be lowered natively.

## Architecture constraints

- arbitrary updater transform/style output does not create semantic revision/relowering every frame;
- tracing/content accumulation uses semantic mutation only when the observable model is persistent authored structure; transient regenerated content uses effective resource state;
- numeric text driven each frame does not automatically perform semantic `ReplaceContent`; use native/effective content updates where semantics permit;
- callback ordering/backpressure/replay classification follows #70;
- time remapping composes semantic animation intent rather than rebuilding timelines;
- execution class is visible in diagnostics.

## Acceptance

- representative supported indication/tracing/updater/number/time-remap cases match pinned observable behavior;
- deterministic/native cases require no Python after authoring;
- host-backed cases preserve updater order and explicit replay/seek semantics;
- ordinary updater property writes stay on effective runtime lane;
- persistent structural/content mutations remain local and atomic;
- transient dynamic content does not churn semantic revision;
[...]

PR #286, opened 2026-08-26
Add retained Manim Blink compatibility

yongkyuns:
Tracks #81.

## Scope
- adds ManimCE v0.21 `Blink` for one leaf 2D `VMobject`
- mirrors upstream `Blink(Succession)` phase structure: opacity-on `UpdateFromFunc`, opacity-off `UpdateFromFunc`, repeated `blinks` times, plus the final on phase unless `hide_at_end=True`
- lowers those deterministic callbacks to constant retained ObjectSnapshot tracks, so fill/stroke alpha switches at the exact child boundary without waking Python during playback
- reuses Noon's shared composition scheduler for intrinsic timing, `run_time` scaling, outer rate functions, nested composition time maps, seek/rewind, and exact final-state selection
- keeps arbitrary user updaters on the existing host-callback path; Blink itself does not create a callback session
- adds a focused CPython regression covering defaults, phase timings, fill/stroke alpha, final state handoff, `hide_at_end`, multiple blinks, play-level runtime scaling, and unsupported boundaries

## Exact retained subset
- one leaf 2D `VMobject`
- positive finite `time_on` / `time_off`
- positive integer `blinks`
- canonical Succession ordering (`lag_ratio=1`)
- no concurrent user mobject updater on the Blink target
- no same-object overlap with another generic Transform

Group/VGroup family opacity, zero-duration/count edge cases, updater-order interactions, and overlapping same-object generic transforms remain explicitly partial rather than approximated.

## Architecture
Manim implements Blink through `UpdateFromFunc`, but the callback ignores alpha and only calls `set_opacity(1/0)`. Noon therefore does not route Blink through the host callback protocol. Each child is represented by a constant retained full-style snapshot because Manim `VMobject.set_opacity` writes fill and stroke alpha independently.

Mapped composition is safe: runtime selection ignores mapped tracks whose child interval has not begun and chooses the latest begun phase; at the exact root endpoint Noon settles to the final authored target, matching Blink cleanup state.

## Qualification / gallery
The branch is based on current master and intentionally does not touch shared parity manifests while several transform-family parity branches are active. After implementation CI is green, attach a canonical source-equivalent geometry fixture to #176/#185 with unchanged tolerances. Do not classify Blink as parity-qualified until semantic, Cairo raster/timeline, direct-seek/incremental-playback, WebGPU, and WebGL gates pass.

No public gallery entry is added here because #273 requires a ready entry to be literal checked-in upstream Manim source with import-only substitution; the public Blink documentation example depends on currently unsupported exact text coverage.

### DIFF
--- a/web/python/_manim_updaters.py
+++ b/web/python/_manim_updaters.py
@@ -315,7 +315,7 @@ def _positive_phase_time(value: object) -> float:
 
 
 class Blink(_composition.Succession):
-    """Blink one leaf mobject without running a Python callback during playback."""
+    """Blink one leaf VMobject without running a Python callback during playback."""
 
     def __init__(
         self,
@@ -330,8 +330,14 @@ class Blink(_composition.Succession):
             raise NotImplementedError(
                 "Blink(Group/VGroup) requires retained family opacity semantics and remains partial"
             )
-        if not isinstance(mobject, _base.Mobject):
-            raise TypeError("Blink target must be a Mobject")
+        if not isinstance(mobject, _compat.VMobject):
+            raise NotImplementedError(
+                "Blink currently supports one leaf 2D VMobject; non-vector Mobjects remain partial"
+            )
+        if _updaters(mobject):
+            raise NotImplementedError(
+                "Blink with user mobject updaters has callback-order semantics and remains partial"
+            )
         if isinstance(blinks, bool) or not isinstance(blinks, int):
             raise TypeError("Blink blinks must be an integer")
         if blinks <= 0:
--- a/web/python/_manim_updaters.py
+++ b/web/python/_manim_updaters.py
@@ -3,17 +3,26 @@
 The Python callable stays in Pyodide. During playback the main thread sends one
 coherent runtime snapshot for the callback phase; all getter/setter traffic stays
 inside this module and only one PatchBatch crosses back to WASM.
+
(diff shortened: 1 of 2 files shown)


yongkyuns:
Static exactness audit after implementation:

- ManimCE Succession switches children on `current_time >= active_end`; `next_animation()` finishes the old child and `begin()`s the next child immediately. Therefore at the exact `time_on` boundary Blink has already executed the opacity-off UpdateFromFunc. Noon's later constant retained phase is likewise selected at that exact timestamp.
- Runtime mapped-channel selection scans backward for the latest mapped track whose child interval has begun, so nested/nonlinear compositions do not let future Blink phases override the active phase.
- At the exact root endpoint, mapped tracks intentionally settle to the final authored target, matching Blink finish/cleanup state.
- Composition wrapper restoration only restores Python scene/object bindings; Blink lowering does not mutate wrapper style, so authoring state does not leak across nested scheduling.
- API coverage requires no shared inventory edit: a newly exported but unclassified upstream symbol resolves to `partial` until an explicit behavioral override is added.

The PR remains draft because connector-authored PR/open/ready/synchronize events still produce no Actions check suite on this head. Externally-authored current PRs do have fresh Actions, so this is a trigger-path limitation rather than evidence of a pass.

Issue #363, opened 2026-08-27
[Arch C3.2] Add bounded paged glyph/raster/outline cache residency


yongkyuns:
Parent: #361
Related: #65, #114, #339-#342

## Goal
Give retained text rendering an explicit bounded residency policy suitable for long-running interactive scenes, zooming, editing and hot reload.

## Current problem
The current architecture correctly caches rasterized glyphs and GPU atlas entries, but residency is not yet a complete product policy:
- fixed atlas extent can become full;
- atlas allocation is append-only shelf packing;
- CPU raster cache can grow with every unique glyph/size/variation;
- outline/stroked-outline caches can grow with every path-level request;
- effective raster size currently changes at integer pixel boundaries, so long zoom ranges can accumulate many copies of the same glyph.

## Scope
Design and implement one coherent residency manager covering:
- paged/growable atlas allocation or equivalent bounded multi-atlas scheme;
- LRU/clock/generation-based eviction/reuse policy with deterministic behavior;
- bounded CPU raster cache;
- bounded lazy outline/stroked-outline cache;
- explicit raster-size bucket policy with hysteresis or geometric steps to avoid churn during smooth zoom;
- pinning semantics for resources actively referenced by the current resident scene/frame;
- renderer/device-generation invalidation;
- counters for resident bytes, pages, entries, hits/misses, evictions, rerasterization and outline regeneration.

Keep mask and color glyph planes separate where required by texture format/shader semantics.

## Requirements
- eviction must not change layout/glyph identity, only raster residency;
- font bytes remain immutable resources outside frame state;
- cache keys retain exact font face/variation/glyph identity;
- translation must not participate in raster identity;
- camera/DPR/scale policy must not visibly undersample text;
- no synchronous global cache rebuild when one page is evicted;
- atlas/page changes must integrate with retained per-object preparation from #362.

## Tests / acceptance
Add deterministic stress tests:
1. cycle through far more unique glyphs than one atlas can hold;
2. animate camera scale across a large range and back repeatedly;
3. alternate bounded working sets and verify steady-state reuse;
4. exercise color glyphs plus alpha-mask glyphs;
5. repeatedly trigger/release reveal/stroke outlines;
6. destroy/recreate renderer generation.

Acceptance:
- no `AtlasFull` product failure for a workload within configured total residency budget;
- memory reaches a documented plateau rather than monotonically increasing;
- returning to a recent working set shows cache reuse;
[...]

Issue #361, opened 2026-08-27
[Arch C3] Converge text/font/Typst/TeX on one canonical retained scene and scalable renderer


yongkyuns:
## Objective
Consolidate Noon's text/font/math implementation around the architecture that has now proven sound in #65/#304/#306/#312/#319/#324/#326/#329/#330/#339-#351, while removing transitional authoring/transport complexity and closing the remaining scalability/dynamic-content gaps before `Text` and real `Tex`/`MathTex` broaden the surface.

This is an architecture/convergence issue, not a replacement for public parity issue #83. #83 continues to own the Manim-facing APIs and fixtures. This issue owns the shared implementation direction those APIs must consume.

## Reviewed current architecture
The core retained model is sound and should be preserved:

```text
source language -> layout backend -> TextResource -> retained scene/runtime -> shared renderer

Native Text/MarkupText --+
Typst/MathTypst ----------+-> TextResource
Tex/MathTex (real LaTeX) -+
                              |- GlyphRun -> atlas for steady state
                              |- TextVectorItem -> shared geometry resources
                              `- lazy glyph outlines -> shared path renderer when required
```

### Decisions to keep
- `TextSourceKind` remains distinct from `TextLayoutBackendKind`; real LaTeX is not translated to Typst.
- `TextResource` remains the backend-neutral normalized representation.
- Exact OpenType bytes are retained outside frame snapshots and glyph IDs resolve against immutable font identity.
- Variable-font coordinates remain explicit rasterization inputs rather than being reconstructed from a string key.
- Ordinary glyphs remain atlas-backed; do not persist one vector path per glyph.
- Fraction rules/radicals/backend-authored shapes remain shared immutable vector geometry.
- stroked/reveal/morph/path-level glyph behavior uses lazily cached outlines through the existing path renderer.
- geometry and text share one semantic `ObjectId` and one global painter-order stream.
- Python must never own glyph arrays/font buffers/SVG/atlas state.

These choices already give the correct backend-independent architecture for native text, Typst and real LaTeX, and for Rust/Python/JavaScript frontends.

### 1. Retained execution is incremental, render preparation is not
`RetainedSceneInstance` and the retained execution delta path can identify changed objects, but the renderer still prepares text largely as a whole-frame operation:
- scan all present retained objects;
- rebuild all visible glyph quad records;
- copy prepared text arrays into a second snapshot buffer;
- rebuild scratch geometry for vector/outline lanes;
- reconstruct the mixed painter-order stream.

[...]

PR #386, opened 2026-08-27
Stabilize glyph raster size residency

yongkyuns:
## Summary
- make the long-lived retained text renderer actually use #375's bounded CPU raster cache, with a finite default of 8,192 entries / 64 MiB image payload
- expose renderer-level raster cache limits and runtime tightening
- replace integer-per-pixel raster identity with conservative 12.5% geometric size buckets
- always round raster size upward so cache stabilization never undersamples the active transform
- keep translation out of raster identity and preserve exact font/glyph/variation keys
- add focused tests for finite renderer defaults, no undersampling, nearby-scale reuse, and bounded bucket count over a 1..1000 zoom range

## Why
#375 introduced the cache policy primitive but deliberately kept `GlyphRasterCache::new()` unbounded for standalone compatibility. `RetainedTextQuadPreparer` was still using that unbounded constructor, and continuous camera/object zoom still selected every integer pixel size. This slice closes both product-level gaps without changing shaping or authoring semantics.

## Policy
- default CPU raster residency: 8,192 entries / 64 MiB
- raster bucket ratio: 1.125x
- a request always selects the first bucket >= requested device-pixel size
- no hysteresis state is added in this slice; after first residency, crossing a bucket boundary only switches between a small bounded set of cached sizes

## Invariants
- no layout/glyph/source identity changes
- no frontend or worker protocol changes
- no translation-dependent raster keys
- bucket selection is conservative and cannot reduce requested resolution

## Follow-up
- paged/bounded GPU atlas eviction/reuse
- optional per-object hysteresis if bucket-boundary oscillation remains measurable after bounded residency

Tracks #363. Architecture parent #361.

### DIFF
--- a/crates/noon-text-render-wgpu/src/lib.rs
+++ b/crates/noon-text-render-wgpu/src/lib.rs
@@ -697,16 +697,7 @@ mod tests {
 
     #[test]
     fn raster_size_buckets_never_undersample_requested_size() {
-        for requested in [
-            0.0,
-            1.0,
-            1.01,
-            10.0,
-            100.0,
-            256.1,
-            1_000.0,
-            100_000.0,
-        ] {
+        for requested in [0.0, 1.0, 1.01, 10.0, 100.0, 256.1, 1_000.0, 100_000.0] {
             let bucket = raster_size_bucket(requested);
             assert!(bucket >= requested.max(1.0));
             assert!(bucket.is_finite());
--- a/crates/noon-text-render-wgpu/src/lib.rs
+++ b/crates/noon-text-render-wgpu/src/lib.rs
@@ -20,8 +20,18 @@ use noon_text_atlas::{
     DEFAULT_GLYPH_ATLAS_EXTENT,
 };
 use noon_text_raster::{
-    GlyphRaster, GlyphRasterCache, GlyphRasterError, GlyphRasterKey, GlyphRasterStats,
+    GlyphRaster, GlyphRasterCache, GlyphRasterCacheLimits, GlyphRasterError, GlyphRasterKey,
+    GlyphRasterStats,
 };
+
+pub const DEFAULT_GLYPH_RASTER_CACHE_MAX_ENTRIES: usize = 8_192;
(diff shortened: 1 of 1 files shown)


yongkyuns:
Raster oracle follow-up: the current 1.125x policy is intentionally **not mergeable**. The failure is real (`webgpu/hello-typst` differing ratio 0.008001543 > 0.008), even though it is only a couple pixels over the ratchet.

Correction direction:
- keep the pre-existing `requested.ceil()` raster identity exactly through ordinary display sizes;
- activate residency bucketing only above an explicit high-zoom threshold;
- keep bucket outputs integer-valued and always >= the requested size;
- preserve the finite 8,192-entry / 64 MiB renderer CPU-raster default from this PR;
- add tests that ordinary sizes are byte-identical to legacy integer selection and that high zoom still collapses raster identities substantially;
- do **not** loosen any Manim raster ratchet.

This separates visual parity from the cache-churn optimization instead of accepting a steady-state raster change.

yongkyuns:
Fixed the raster regression on head `aa4975e3` without relaxing any ratchet.

The original 1.125x policy changed ordinary glyph raster identity and produced the observed 2-pixel overage. The corrected policy now:
- keeps the finite 8,192-entry / 64 MiB CPU raster cache;
- preserves the pre-#386 `requested.ceil().max(1.0)` identity through 256 physical pixels;
- uses conservative 1.125x geometric buckets only above 256 px;
- always rounds high-zoom requests upward, so bucketing cannot undersample;
- adds explicit tests for exact ordinary-size identity, high-zoom sharing, no undersampling, and bounded high-zoom bucket count.

The existing raster oracle remains unchanged; this head must pass it before merge.

yongkyuns (commented):
The corrected raster policy is now clean through rustfmt/clippy/workspace tests and has passed the retained Typst raster ratchet again. No tolerance/fixture changes were made; ordinary <=256px raster identity remains the legacy integer ceil path, with geometric bucketing only above the threshold.

Issue #369, opened 2026-08-27
[Roadmap B4] Shared text compiler registry and compiled-resource cache


yongkyuns:
## Roadmap
Phase B4 child of #954. Architecture authority: `docs/architecture.md`.

## Goal
Provide one backend-neutral text compilation boundary and content-addressed compiled-resource cache so native Text, Typst and LaTeX do not become separate authoring/runtime pipelines.

## Target contract
```text
TextSpec
  source kind/source
  layout/backend options
  template/preamble where applicable
  font/resource inputs affecting output
        |
        v
TextCompilerRegistry
  Native / Typst / LaTeX
        |
        v
CompiledTextArtifact
  normalized retained text/vector resources
  deterministic fingerprint
```

### B4.1 Compiler interface
Define a shared compiler/normalization interface. Backend-specific code ends at the normalized artifact boundary; Runtime/Renderer never branch on the source language.

### B4.2 Cache identity
Include every immutable input that can affect normalized output: source/kind, backend/compiler version, templates/environment, font/resource fingerprints, shaping/layout/features/variations and deterministic configuration. Exclude presentation-only transform/style inputs when they do not affect layout.

### B4.3 Resource deduplication
Repeated identical text may share compiled immutable resources while keeping independent semantic `NodeId`s and presentation state.

### B4.4 Lifetime/accounting
Define cache lifetime/eviction/accounting compatible with long-running authoring sessions and localized content replacement (#368). Failed compiles do not install/poison partial cache entries.

### B4.5 Diagnostics
Expose deterministic compile/cache hit/miss/time/retained-byte counters and typed backend errors.

## Acceptance
- many identical Text/Typst/LaTeX objects compile a backend once or a documented minimal count and share immutable resources;
- presentation-only changes do not trigger recompilation;
- source/backend/template/font-layout changes produce distinct artifact identity;
- failed compilation leaves no installable partial artifact;
- runtime/renderer consume only the normalized retained representation;
- no backend-specific authoring/resource insertion path bypasses this boundary.

## Dependencies
Phase A semantic/resource ownership. #364 and #365 consume this interface; #368 consumes its versioned replacement semantics; #362 consumes normalized retained resources.

yongkyuns:
[...]

PR #390, opened 2026-08-27
Add shared text compiler cache contract

yongkyuns:
## Summary
- add backend-neutral `noon-text-compile` infrastructure for #369
- define a complete content/layout `TextCompileKey` that intentionally excludes presentation-only object state
- define one normalized `CompiledTextArtifact` containing `TextResource`, exact font dependencies, vector/geometry dependencies, deterministic artifact identity, and retained-byte accounting
- define a `TextCompiler` trait with an associated error type so native/Typst/LaTeX diagnostics stay typed
- add a bounded deterministic LRU `TextCompileCache` with finite defaults, immediate limit tightening, oversized-admission rejection, and hit/miss/compile/error/eviction counters
- ensure compile failures are never cached or partially admitted
- add focused tests for identical-content compile-once reuse, identity separation, recent-working-set LRU behavior, oversized artifacts, failure recovery, and runtime limit tightening

## Architecture
This is deliberately the additive contract/cache foundation only. It does **not** rewire the already-stable Typst path or couple native/LaTeX implementations yet. Follow-up adapters can implement `TextCompiler` independently, then a heterogeneous registry/dispatcher can be layered on the same contract without changing retained runtime/renderer resource shape.

## Invariants
- no renderer/per-frame work
- no backend-specific payloads in the normalized artifact
- transform/color/opacity/z-order/reveal/morph do not participate in compile identity
- source/backend version/template/layout/font/config inputs can never alias accidentally
- immutable exact font and geometry dependencies remain outside mutable frame state
- cache memory is bounded by entries and deterministic retained-byte accounting

Tracks #369. Architecture parent #361.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "crates/noon-web",
     "crates/noon-typst",
     "crates/noon-text-native",
+    "crates/noon-text-compile",
     "crates/noon-text-raster",
     "crates/noon-text-atlas",
     "crates/noon-text-render-wgpu",
--- a/crates/noon-text-compile/Cargo.toml
+++ b/crates/noon-text-compile/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "noon-text-compile"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+description = "Backend-neutral text compiler registry and compiled artifact cache for Noon"
+
+[dependencies]
+noon-core = { path = "../noon-core" }
--- a/crates/noon-text-compile/src/lib.rs
+++ b/crates/noon-text-compile/src/lib.rs
@@ -1,0 +1,504 @@
+#![forbid(unsafe_code)]
+
+//! Backend-neutral text compilation and compiled-artifact caching for Noon.
+//!
+//! This crate sits between authoring/source backends and the retained runtime. It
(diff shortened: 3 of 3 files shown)


yongkyuns (commented):
The foundation direction matches #361/#369, but the current contract does not yet enforce two acceptance-critical invariants before cache admission:

1. `CompiledTextArtifact::new()` calls `TextResource::validate()`, which validates internal spans/render-item/range structure but does not prove dependency closure. Every `TextVectorItem.geometry` must resolve in the supplied `GeometryResourceArena`, and every `GlyphRun.font` must resolve in the supplied `FontResourceArena` (same checks currently performed by retained resource transport).
2. `get_or_compile()` accepts any artifact returned by a compiler under the requested `TextCompileKey`. The artifact should be validated against the key before admission: source, source kind, and layout backend/version (and relevant layout/template fingerprint identity) must not disagree with the key. Otherwise a buggy backend can cache a normalized artifact under the wrong content identity and create exactly the cross-backend/template aliasing #369 is intended to prevent.

Please make cache admission validate dependency closure + key/artifact identity, return a typed shared contract error (wrapping backend errors as needed), and add negative tests proving missing font/vector dependencies and mismatched source/kind/backend cannot be admitted. Compile failures/validation failures should leave resident entries/bytes unchanged.

Issue #400, opened 2026-08-27
[Roadmap B2] Expose shared geometry through thin Python semantic-handle adapters


yongkyuns:
## Roadmap

Phase B2 of #954. Depends on Phase A #953, especially A3/#61.

## Goal

Expose the common geometry families implemented in shared Rust through the Manim-compatible Python facade **using the authoritative Semantic Scene handle API**. Do not preserve or extend pre-reset snapshot/SceneDocument adapter paths.

JavaScript is not a completion dependency for this issue; future JS/TS exposure is tracked by #259.

## Scope
As shared geometry semantics become ready, expose supported families including:
- `Arc`, `ArcBetweenPoints` and arc queries;
- `AnnularSector`, `Sector`, `Annulus`;
- `Polygon`, `Polygram`, `RegularPolygon`, `RegularPolygram`, `Triangle`, `Star`;
- `RoundedRectangle`;
- `SurroundingRectangle`, `BackgroundRectangle`, `Cross`, `Underline`;
- existing Dot/Ellipse/Circle/Rectangle/Square behavior through the same ownership model.

## Architecture requirements
- Python owns constructor/signature/type adaptation only;
- geometry construction/query/bounds/path semantics remain shared Rust behavior;
- wrappers hold semantic handles, not authoritative geometry snapshots;
- ordinary queries/mutations use typed handle calls, not JSON scene round-trips;
- no new renderer primitive solely to expose a public class;
- no frontend reconstruction of path/vertex/arc math.

## Implementation cases
1. expose semantic-handle constructors for each geometry family;
2. expose typed shared queries needed by the public API;
3. make generic transform/style/layout operations work through the normal Mobject handle path;
4. remove obsolete geometry-specific Python snapshot/fallback code as each family moves;
5. update public namespace/compatibility metadata only after the actual adapter exists.

## Acceptance
- representative source-equivalent Manim snippets construct/query each supported family through Python;
- Python and idiomatic Rust observe the same semantic state/results;
- no Python geometry math is authoritative;
- no SceneDocument/JSON round-trip is needed for ordinary geometry operations;
- #401 exact semantic/raster qualification gates supported families before `parity-qualified` status.

## Relationship
Parent feature scope: #76. Exact-output qualification: #401 / #185.

yongkyuns:
[...]

Issue #76, opened 2026-08-24
[Roadmap B2] Common 2D geometry primitives


yongkyuns:
## Roadmap
Phase B2 geometry parent under #954. Architecture authority: `docs/architecture.md`.

## Goal
Cover the common ManimCE v0.21 2D geometry vocabulary using shared semantic geometry and retained immutable resources, without Python-owned shape math or feature-specific renderer architecture.

### Arc/circle family
- Dot/Ellipse/Arc/ArcBetweenPoints;
- Annulus/AnnularSector/Sector;
- common arc parameters and start/end/angle/center queries.

### Polygon/polygram family
- Polygon/Polygram/RegularPolygon/RegularPolygram/Triangle/Star;
- stable vertex/order/query semantics.

### Rectangle/matcher family
- RoundedRectangle;
- SurroundingRectangle/BackgroundRectangle/Cross/Underline where B1/text dependencies permit.

Existing Circle/Rectangle/Square must use the same semantic interfaces.

## Architecture constraints
- preserve analytic content where profitable;
- lower to immutable path resources when path-level semantics require it;
- expose canonical path/query information required by Create, morphing and matching transforms;
- transform/style-only edits do not rebuild immutable geometry;
- geometry/bounds/query semantics live in shared Rust;
- Python adapters are thin semantic-handle wrappers (#400);
- no geometry class exists only as a renderer special case.

## Acceptance
- representative source-equivalent v0.21 constructors/style/query behavior works through shared handles;
- Create/Transform integrate through B3 without a second geometry animation path;
- bounds/centers/vertices/start/end/arc queries pass focused semantic qualification;
- compatibility stroke scaling follows #62;
- curved/filled/stroked/transparent/overlap cases pass #401/#185 B7 qualification;
- local transform/style edits retain content resources and bounded dirty work.

## Dependencies
A1/B1/#53/#62 establish semantic identity/resources/values. #400 exposes the Python facade; #401 owns exact geometry qualification.

yongkyuns:
### Acceptance criteria amendment — exact Manim output parity

This breadth issue must use the exact-output gate from #185, not only API/semantic/example coverage. For **every slice promoted to supported**:

- add a source-equivalent ManimCE **v0.21.0** fixture to the canonical #176 raster + timeline corpus (no Noon-only geometry/style/timing tuning; only the minimum import/browser wrapper adaptation);
[...]

PR #405, opened 2026-08-27
Expose shared Arc constructors through the authoring store

yongkyuns:
## Summary
First isolated implementation slice for #400.

- add a `noon-web` bridge that constructs `Arc` and `ArcBetweenPoints` by calling the shared Rust `noon` authoring implementation
- attach those canonical snapshots through the existing `WasmAuthoringStore::createMobject` path, preserving the current semantic-store identity/transform/style ownership
- expose separate default-radius and explicit-radius `ArcBetweenPoints` constructors instead of encoding optional radius with a sentinel
- reject non-finite / f32-overflowing JS inputs before narrowing
- add native tests proving the bridge retains shared `VectorPath` geometry and exact endpoints

## Architecture
This PR intentionally does **not** implement arc/path math in Python or JavaScript. The browser bridge only marshals scalar/vector inputs into shared Rust and returns the same `WasmAuthoringMobjectHandle` used by Circle/Rectangle/Line.

No renderer, text/font, worker protocol, or retained execution changes are involved.

## Why this is a separate first slice
The existing Python semantic-handle layer already attaches typed Circle/Rectangle/Line constructors through one long-lived `WasmAuthoringStore`. This PR first certifies that the same store can be extended from an isolated module for geometry breadth. Once green, the next commit/slice can expose the Python names and then add typed Arc query access without changing object identity or reconstructing geometry in Python.

## Validation
- focused native bridge tests
- normal workspace fmt/Clippy/tests
- normal Web package build exercises the generated wasm-bindgen surface

Tracks #400. Parent #76.

### DIFF
--- a/crates/noon-web/src/lib.rs
+++ b/crates/noon-web/src/lib.rs
@@ -11,6 +11,7 @@ mod host_player;
 #[path = "legacy.rs"]
 mod legacy;
 mod lifecycle;
+mod manim_geometry_bridge;
 mod reactive_authoring_facade;
 mod reactive_player;
 mod retained_authoring;
--- a/crates/noon-web/src/manim_geometry_bridge.rs
+++ b/crates/noon-web/src/manim_geometry_bridge.rs
@@ -152,7 +152,7 @@ mod wasm {
 
 #[cfg(test)]
 mod tests {
-    use noon_core::{GeometryRef, ObjectSnapshot, PathCommand};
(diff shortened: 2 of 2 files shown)


yongkyuns:
Superseded by the landed shared-geometry adapter foundation and the current Arc stack (#412 for Rust/WASM metadata/query bridging, #418 for the thin Python facade). This draft now overlaps a `manim_geometry_bridge` module already present on master and is intentionally not being rebased. Closing to keep #400's implementation direction unambiguous.

PR #408, opened 2026-08-27
Fix Rotating arrow family motion

yongkyuns:
## Summary
- fix the visible Arrow tip/shaft separation in the public Manim `RotatingDemo`
- keep the existing exact centered-leaf `Rotating` fast path unchanged
- lower candidate z-axis family/external-pivot rotation into deterministic <=5° transform intervals so retained children follow the same circular motion to sub-pixel tolerance instead of one long chord
- preserve global rate-function timing by sampling the shared deterministic rate function at interval boundaries
- add a regression that verifies the Arrow shaft/tip family uses continuous short intervals and remains attached at every interval endpoint

## Root cause
`Arrow` is retained as a family (shaft + triangular tip). The broad `Rotating` compatibility path was lowering each leaf to one generic start/end `Transform`. For a 180° rotation, the tip translation therefore travelled across a straight diameter while the shaft rotation followed its own transform interpolation, producing the visible detachment/collapse.

## Scope
This addresses the z-axis rotations that cause the broken rotating-arrow/circle demo while retaining deterministic seek/rewind semantics. The existing x/y 3D projection portions of upstream `RotatingDemo` remain explicitly candidate parity until native 3D intermediate-frame representation lands.

### DIFF
--- a/web/python/_manim_updaters.py
+++ b/web/python/_manim_updaters.py
@@ -377,8 +377,11 @@ def _install_rotating_breadth() -> None:
 
     The qualified centered-leaf implementation stays in ``_manim_rotate``. This final
     bootstrap wrapper handles retained groups, external pivots, and the 180-degree x/y
-    projections used by the upstream documentation scene. It intentionally remains a
-    parity candidate until curved z-pivot and 3D intermediate-frame tracks are native.
+    projections used by the upstream documentation scene. Z-axis family rotation is
+    subdivided into short deterministic transforms so independently retained children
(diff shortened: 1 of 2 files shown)


yongkyuns:
Superseded by #410 and #425, both now merged. #410 landed the deterministic segmented z-axis family motion for the Arrow shaft/tip and circle collapse defect; #425 completed the principal x/y projection path, exact segment handoffs, semantic-handle commit-back, f32 principal-axis tolerance, and full RotatingDemo regression. Keeping this older branch open would duplicate now-integrated behavior.

PR #412, opened 2026-08-27
Bridge Arc metadata and snapshot queries to WASM

yongkyuns:
## Summary
- add a thin `noon-web` bridge for shared Rust `Arc` and `ArcBetweenPoints` construction
- keep immutable constructor metadata in `ArcMetadata`
- expose snapshot-derived start/end/center/stop-angle queries through the same Rust functions used by typed Arc authoring
- keep `ObjectSnapshot` as the only mutable geometry state
- add focused tests for transformed snapshot queries and negative explicit-radius metadata

## Architecture
The bridge deliberately does not add frontend geometry math or a second mutable Arc representation. Explicit query calls can consume the current semantic-handle snapshot JSON and immediately evaluate it in Rust; rendering and mutations remain on the existing shared handle path. A later optimization can add direct handle methods without changing semantics.

Stacked on #406. Tracks #76 and #61.

### DIFF
--- a/crates/noon-web/src/manim_geometry_bridge.rs
+++ b/crates/noon-web/src/manim_geometry_bridge.rs
@@ -1,4 +1,7 @@
-use noon::{Dot, IntoSnapshot, Triangle};
+use noon::{
+    arc_center_from_snapshot, arc_end_from_snapshot, arc_start_from_snapshot,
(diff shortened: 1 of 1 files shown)


yongkyuns:
Superseded by #611, which replayed this Arc bridge onto current master, preserved the newer geometry bridge work, passed the full exact-head CI/raster/browser/release matrix, and merged as dad4d38321ea80bd0d08decc8d36b6827345f8c9. Closing this stale stack root to avoid duplicate review/CI scope.

PR #418, opened 2026-08-27
Route Python Arc semantics through shared Rust

yongkyuns:
## Summary
- expose Manim-compatible `Arc` and `ArcBetweenPoints` through the existing shared-geometry Python adapter
- construct browser Arc geometry from the Rust/WASM Arc spec and attach it to the ordinary shared semantic mobject handle
- keep `radius`, `start_angle`, `angle`, and `num_components` as immutable constructor metadata supplied by Rust
- delegate `get_start`, `get_end`, `get_arc_center`, and `stop_angle` to Rust against the current semantic-handle snapshot
- make `move_arc_center_to` a translation derived from the shared Rust center query
- free temporary WASM Arc spec/query objects in the worker and add source guards for that wiring
- retain a native-CPython-only cubic fallback for local compatibility tests; browser tests make that fallback throw to prove it is not authoritative

## Architecture
The browser path has no Python-owned Arc geometry or center/endpoint math. `ObjectSnapshot` in the semantic handle remains the only mutable Arc state. Python keeps only constructor metadata and forwards explicit queries to the same Rust snapshot functions used by typed Rust Arc authoring. No worker protocol or renderer primitive is added.

Stacked on #412. Tracks #76 and #61.

### DIFF
--- a/crates/noon-web/src/lib.rs
+++ b/crates/noon-web/src/lib.rs
@@ -11,6 +11,7 @@ mod host_player;
 #[path = "legacy.rs"]
 mod legacy;
 mod lifecycle;
+mod manim_arc_bridge;
 mod manim_geometry_bridge;
 mod reactive_authoring_facade;
 mod reactive_player;
@@ -35,6 +36,7 @@ pub use execution_transport::*;
 pub use host_player::*;
 pub use legacy::*;
 pub use lifecycle::*;
+pub use manim_arc_bridge::*;
 pub use manim_geometry_bridge::*;
 pub use reactive_authoring_facade::*;
 pub use reactive_player::*;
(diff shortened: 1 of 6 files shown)


yongkyuns:
Roadmap triage after #952: closing this Arc facade PR. The shared Rust Arc semantics and query cases remain Phase B2 requirements under #954/#400/#76, but this branch is based on the old snapshot/adapter model (`ObjectSnapshot` as mutable authority). Replay it after Phase A using the authoritative Semantic Scene handle API.

yongkyuns:
Superseded by Phase A (#953) after #952 established one authoritative Semantic Scene. Keep the Arc semantic/query tests and adapter evidence, but replay the public facade on the consolidated typed semantic-handle ABI rather than extending this pre-consolidation adapter stack.

Issue #362, opened 2026-08-27
[Roadmap C3] Dirty-object incremental retained text rendering


yongkyuns:
## Roadmap

Phase C3 child of #955. Architecture authority: `docs/architecture.md`.

## Goal

Make text/mixed renderer preparation proportional to dirty objects/member state rather than total visible glyph/object count, while consuming coherent effective Runtime publications rather than reconstructing semantic truth.

## Input/identity contract

The renderer consumes Runtime `FrameState`/change sets for a specific coherent `FrameEpoch` plus derived execution/resource identity. It does not read the Semantic Scene as current live authority and must not invalidate everything merely because `SceneRevision` or `FrameEpoch` changed.

Retained records are keyed by the derived facts that affect them, for example:

```text
execution slot + generation
immutable ResourceVersion/content handles
painter/topology position
prepared glyph/instance ranges
vector/outline bindings
family-member bindings
last projected effective mutable state
```

### C3.1 — incremental prepared state
- keep glyph atlas/path/resource residency across frames;
- update only objects/ranges named by Runtime dirtiness;
- effective presence/style/transform changes update mutable records without rebuilding immutable content;
- structural/persistent content/resource changes invalidate only affected records/dependencies;
- stale change sets from an older execution/frame context are rejected rather than applied to a newer retained state.

### C3.2 — painter ordering
Ordinary effective property changes do not rebuild global mixed painter order. Structural/order/topology changes update only the required ordering structures.

### C3.3 — text lane locality
- static glyph runs remain resident;
- effective transform/style/appearance changes update only affected instance records;
- outline/path realization is lazy and versioned;
- camera/DPR changes invalidate only representation genuinely dependent on them;
- old resource versions referenced by in-flight submissions remain valid until safe retirement.

### C3.4 — family animation
#835 reuses the same retained/local renderer path. Family progress is effective mutable member state and must not become a parallel scratch-frame authority that rematerializes equivalent glyph/path content every frame.

### C3.5 — instrumentation
Expose deterministic counters for:
- objects/member records prepared;
- glyph/path materializations;
- order rebuilds;
- resource cache misses/versions;
- stale change-set rejections;
- GPU ranges/bytes uploaded;
- explicit full rebuilds/maintenance.

## Acceptance

Use a 10k+ glyph mixed scene and assert after warmup:
- a clean frame performs effectively zero text preparation/upload;
- one effective transform/style change touches only that object's mutable records;
[...]

PR #442, opened 2026-08-28
Reuse unchanged retained parent preparation

yongkyuns:
## Summary
- cache a complete successful parent retained-preparation generation: copied text snapshot, mixed painter-order stream, parent stats, and device metrics
- on empty changes with a reused scratch generation and unchanged metrics, keep the child text preparer's O(1) reuse call for validation/accounting but skip parent snapshot copies and mixed-order reconstruction
- reuse geometry through the existing empty incremental path
- invalidate parent generation validity before fallible preparation and restore it only after full success
- add regressions proving unchanged same-metric parent reuse and metric-change fallback

## Invariants
- no stale parent snapshot can survive a failed child preparation
- unchanged parent reuse does not mutate painter order or copied text arrays
- device metric changes bypass the parent fast path
- child text reuse accounting remains intact

## Scope
One retained-renderer commit rebased directly onto current master. The implementation blob is unchanged from the previously reviewed head; only the parent was refreshed so current browser/runtime/lifecycle/editor/Arc fixes are included.

Current clean head: `d25334c1`.

Tracks #362. Architecture parent #361.

### DIFF
--- a/crates/noon-render-wgpu/src/gpu/retained_text.rs
+++ b/crates/noon-render-wgpu/src/gpu/retained_text.rs
@@ -545,6 +545,10 @@ pub struct RetainedFramePreparer {
     snapshot_color_quads: Vec<GlyphQuadInstance>,
     snapshot_text_items: Vec<PreparedTextItem>,
     snapshot_text_stats: RetainedTextPrepareStats,
+    snapshot_prepare_stats: RetainedPrepareStats,
+    snapshot_metrics: Option<TextDeviceMetrics>,
+    prepared_generation_ready: bool,
+    prepared_generation_reuses: u64,
 }
 
 impl Default for RetainedFramePreparer {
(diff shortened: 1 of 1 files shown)


yongkyuns:
Revalidation note against current master (`a2160889`): the only historical failure on head `16f8443d` was `Playground cross-browser matrix` on Chromium timing out waiting for an applied scene. That run predates the deferred-start playground contract; current `scripts/playground-browser-matrix-smoke.mjs` explicitly asserts deferred startup, clicks `#replace-scene`, then waits for the applied scene. Also, `crates/noon-render-wgpu/src/gpu/retained_text.rs` has not changed in the 78 master commits since this PR's base (`6c5916c5`), so the one-file renderer patch has no product-code conflict to resolve. Do not treat the old browser failure as a renderer regression; refresh/revalidate this PR on current master before merge.

Issue #108, opened 2026-08-24
[Test T5] Add versioned scene/patch serialization contract and compatibility tests


yongkyuns:
## Goal
Treat serialized scene, patch, callback, and browser transport formats as explicit compatibility contracts rather than incidental serde/JSON output.

## Current gap
Noon relies on JSON across Python/JS/WASM/Rust boundaries and supports scene replacement/reconciliation/patching, but there is no dedicated compatibility corpus proving round-trip stability, old-fixture readability, unknown-field behavior, or schema evolution rules across releases.

## Scope
- Inventory every persisted/transmitted payload that crosses a language/process/version boundary:
  - scene documents;
  - patch batches/transactions;
  - callback frame/commit payloads;
  - authoring worker results and version/revision envelopes;
  - any retained animation/reactive definitions intended to survive reload.
- Define explicit versioning/evolution rules for each contract: required fields, defaults, unknown fields, enum expansion, ID width/range, numeric encoding, ordering guarantees.
- Add canonical fixture files checked into the repo for current and representative older versions.
- Test Rust -> JSON -> Rust round trips and Python/JS-produced payload -> Rust/WASM -> normalized output round trips.
- Verify serialization is deterministic/canonical where stable text output is relied upon for identity/diff/cache behavior.
- Add negative fixtures for future version numbers, malformed payloads, unsupported variants, duplicate IDs, and missing required fields.
- Add a migration/upgrade test path if schema versions require explicit conversion.

## Acceptance
- Every cross-language/version payload is listed in one schema/contract document or machine-readable manifest.
- Checked-in fixtures prevent accidental incompatible field/enum changes.
- Current readers can load all compatibility fixtures they are expected to support.
- Unsupported future/invalid formats fail with clear version/context diagnostics rather than panicking or partially applying state.
- Cross-language round-trip tests run in normal CI.
- Intentional breaking schema changes require an explicit fixture/version update.

## Non-goals
- Do not freeze internal Rust-only data structures that never cross a boundary.
- Do not duplicate #73's Manim public API manifest.

## Dependencies / parallelism
**Wave T1; start immediately.** Align new worker envelopes with #60 and hot-reload state with #64 as those land.

Issue #105, opened 2026-08-24
[Test T2] Add model-based/property tests for semantic, patch, timeline, and reactive invariants


yongkyuns:
## Goal
Use generative/model-based tests to protect the stateful engine invariants that example-based tests are least likely to exhaust.

## Current gap
There are good focused unit/integration tests for semantic handles, transactional patches, transforms, lifecycle, and reactive execution, but most fixtures use hand-authored short sequences. The engine is increasingly stateful: deletion/reuse, family aliasing, track edits, nested composition, reactive dependencies, and local transactions create a combinatorial state space.

## Scope
Build deterministic seeded property/state-machine suites for:

### Semantic store
- insert/remove/reinsert with stale-handle rejection;
- source identity uniqueness and reassignment;
- arbitrary DAG family membership/aliasing;
- cycle rejection and parent/member symmetry;
- unrelated IDs remain stable after local mutations.

### Scene patches / transactions
- random valid/invalid patch batches;
- atomic rollback on every failing suffix position;
- equivalence of batched and sequential successful mutations;
- create/remove/track dependency invariants;
- fast-path and conservative-path semantic equivalence.

### Timeline/composition
- random non-overlapping/overlapping tracks, zero-boundary samples, nested time maps and rate functions;
- direct evaluation agrees with a simple reference evaluator at sampled times;
- track insertion/removal/replacement cannot corrupt unrelated state.

### Reactive/compute graph
- random acyclic dependency graphs compared with a simple interpreter/reference model;
- dirty-closure locality and stable results independent of valid topological ordering;
- cycle/type/invalid-reference rejection where applicable.

Use reproducible seeds and print minimized/replayable failures.

## Acceptance
- Property tests run in normal CI with bounded deterministic case counts.
- Every failure reports a seed/minimized operation sequence that can become a fixed regression test.
- At least one simple reference model/oracle exists for each stateful subsystem above.
- Structural-locality counters are asserted where already exposed, without duplicating the full performance gate in the separate testing lane.
- The suites remain backend-independent and fast enough for normal PR CI.

## Dependencies / parallelism
**Wave T1; start immediately.** Extend as #51/#52/#55/#58/#63 evolve. This issue tests their contracts; it does not own those implementations.

PR #461, opened 2026-08-28
Reject non-finite transported object state

yongkyuns:
## Summary
- reject non-finite geometry, transform, style and color values at the shared `SceneDefinition` transport/mutation boundary
- validate vector-path commands and nested morph targets without introducing range/positivity policy
- keep direct property patches and property/structural transactions atomic when validation fails
- preserve existing unknown/duplicate/ID-exhaustion error precedence
- add focused tests for direct patch rollback, property-transaction rollback, structural-transaction rollback, bulk scene admission, and nested path/morph validation

## Why
Timeline tracks now reject non-finite endpoints, but transported object state could still admit `NaN`/`±∞` through `SceneDefinition::from_parts`, `CreateObject`, `SetGeometry`, `SetTransform`, or `SetStyle`. Once admitted, those values can poison bounds, runtime state and renderer inputs and are difficult to diagnose downstream.

The shared semantic boundary should reject malformed numeric state once rather than relying on Python/JS/renderer sanitization.

## Architecture / performance
Validation stays in `noon-core`, alongside stable-ID and timeline validation. Only the changed/admitted payload is inspected: analytic geometry is O(1); a vector path is O(path commands), including its optional morph target. Transactions still preflight compact metadata and never clone the scene or heavy geometry for rollback.

The policy is intentionally narrow: **finite values only**. Negative scales, zero dimensions, out-of-range-but-finite colors, etc. are left to their existing semantic/compatibility rules rather than being silently constrained here.

## Scope / parallelism
Touches only `crates/noon-core/src/patch.rs` and `patch/transaction_preflight.rs`. It does not overlap the active timeline validation (#458), renderer recovery/profiling (#450/#459), playback controls, retained-text work, or mixed-scene files.

## Remaining boundary
The infallible direct Rust authoring helpers such as `SceneDefinition::add` can still be handed non-finite values by trusted in-process callers. Changing those APIs to return `Result` is a separate public-API design decision; this PR hardens the fallible transport/patch admission boundary without forcing that redesign.

Related: #105, #108.

### DIFF
--- a/crates/noon-core/src/patch.rs
+++ b/crates/noon-core/src/patch.rs
@@ -2,8 +2,8 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 
 use crate::{
-    validate_track_definition, GeometryRef, ObjectDefinition, ObjectId, SceneDefinition, Style,
-    TimelineError, TrackDefinition, TrackId, Transform2D,
+    validate_track_definition, Color, GeometryRef, ObjectDefinition, ObjectId, PathCommand,
+    SceneDefinition, Style, TimelineError, TrackDefinition, TrackId, Transform2D, Vec2, VectorPath,
 };
 
 mod transaction_preflight;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Fresh required CI exposed two `clippy::field_reassign_with_default` violations in the new non-finite-state tests. Fixed both by constructing the invalid `Style` values directly with struct update syntax; production code and test semantics are unchanged. New head: `1fa3be445efcdee0bf3558855f17a674797d1c96`. CI/coverage/API/differential jobs have been retriggered on that exact head. The separate WebGPU painter-order/recovery failure remains in the actively owned renderer-recovery lane and is intentionally not worked around here.

Issue #113, opened 2026-08-24
[Test T10] Add deterministic structural performance regression gates for architecture invariants


yongkyuns:
## Goal
Convert Noon's scalability goals into deterministic CI assertions that catch accidental O(scene-size) work without relying on noisy wall-clock thresholds on shared runners.

## Current gap
`docs/performance.md` has useful named-machine benchmarks for scene replacement/patching, frame preparation/upload, browser scene transport, and manual WebGPU profiling. #68 defines the desired complexity invariants and broader workload set. However, normal CI does not yet provide one systematic suite that fails when a change reintroduces full-scene scans, repacks, uploads, clones, callback work, or draw/resource churn for local/static edits.

This issue is testing infrastructure for #68; it does not own the architecture implementations.

## Scope
Build deterministic workload fixtures plus counters/assertions for:

### Static / mostly-static scenes
- 1k/10k/100k objects;
- unchanged frame: zero semantic/runtime dirty work where expected, zero instance repack/upload, stable cache counts;
- one-object transform/style change: work proportional to affected slots/resources.

### Sparse long timelines
- very large historical track/event count with zero/few active tracks;
- assert evaluation work tracks events crossed + active CPU channels, not total track count;
- seek vs steady playback counters separated.

### Reactive graphs
- large graph, single leaf/input change;
- assert evaluated nodes are limited to dirty dependency closure.

### Geometry/resources
- many shared resources and many unique paths;
- assert local edits do not clone/rebuild unrelated geometry;
- once #59 lands, assert unique-path packing/draw behavior structurally.

### Host/browser bridge
- mostly native scene plus a small host-dynamic subset;
- assert host payload/callback slots scale with host-relevant state rather than total scene size.

### Structural/live edits
- local create/remove/track edit on a large scene;
- assert touched slots/resources/indices remain local according to #58/#66 contracts.

## CI policy
- Gate deterministic counters/operation counts in normal PR CI.
- Keep p50/p95 wall-clock benchmarks as diagnostic/manual or dedicated named-runner measurements.
- Store benchmark output in a machine-readable format so dedicated hardware can compare trends without mixing that noise into correctness gates.

## Acceptance
- A single structural-performance suite maps every target invariant in #68 to one or more measurable counters.
- Deliberately replacing an incremental path with a full scan/repack/upload causes a deterministic CI failure.
- Required assertions cover 100k mostly-static scenes and long sparse timelines without excessive runtime.
[...]

Issue #68, opened 2026-08-24
Architecture scalability reset: parallel workstream tracker


yongkyuns:
## Objective
Restructure Noon now—before broad Manim API expansion—so a large realtime interactive web scene pays for active/dirty/visible work rather than total historical scene complexity, while supporting live authoring without making Python/editor state authoritative runtime scene state.

Target complexity invariants:

```text
static/paused clean frame    ~ O(0) meaningful CPU/render work
timeline CPU work            ~ O(events crossed + CPU-active channels)
reactive CPU work            ~ O(dirty dependency closure)
property edit                ~ O(affected object/slots)
structural edit              ~ O(local dependencies)
render preparation           ~ O(dirty resident state + visible projection)
render upload                ~ O(changed slots/resources)
draw submission              ~ O(visible batches/instances)
hit test                     ~ O(index query + candidates)
host bridge                  ~ O(host-relevant state)
resource regeneration        ~ O(actual content/resource changes)
```

The key product target is not merely high FPS. A paused clean scene should be able to sleep; input/editor/camera changes wake only the necessary execution/presentation path and do not implicitly advance authored timeline time.

See:
- `docs/renderer-locality-and-caching.md` for retained renderer/cache policy;
- `docs/interactive-runtime-and-live-authoring.md` for the ManimCE/ManimGL/Noon reference model and session-state architecture.

## Upstream reference policy

Use upstream Manim implementations for different purposes:

```text
ManimCE v0.21
    -> semantic/API/visual/timing compatibility oracle

current 3b1b/manim
    -> interaction/live-authoring capability reference
    -> renderer optimization ideas to benchmark, not copy blindly

Noon
    -> authoritative execution architecture
    -> language-neutral semantics
    -> deterministic retained execution
    -> worker isolation
    -> O(active + dirty + visible)
```

Do not make current ManimGL source APIs a second compatibility target. Borrow capabilities and measured implementation ideas while preserving Noon's ownership boundaries.

## Wave 1 — start in parallel now
These are deliberately partitioned so separate PRs can progress concurrently.

### Semantic foundation
- #51 — stable authoritative semantic store, family graph, identity
- #62 — semantic numerics/style/bounds/2.5D compatibility model

### Runtime/timeline
- #52 — event-driven timeline scheduler; eliminate all-track-group frame scans

### Resource ownership
- #53 — immutable geometry/resource arena; eliminate heavy geometry cloning

### Renderer correctness
[...]

PR #471, opened 2026-08-28
Make execution-slot transaction preflight sparse

yongkyuns:
## Summary
- replace full `ExecutionSlotTable` cloning during transaction preflight with a sparse touched-slot/object shadow
- preserve create/remove ordering, free-list reuse, duplicate/unknown-object checks, and generation-exhaustion semantics without mutating durable state
- report `slots_indexed` as the unique execution slots actually materialized by structural preflight instead of total historical slot capacity
- add a 100,000-slot regression proving a one-object structural preflight materializes one slot, plus assert non-structural transactions index zero execution slots

## Why
#113 requires local structural work to scale with affected slots/resources rather than historical scene capacity. The previous preflight cloned the entire slot vector and object map for every transaction, making a one-object edit O(total execution-slot capacity).

## Architecture
The durable `ExecutionSlotTable` remains the single committed identity/free-list authority. The preflight shadow borrows it immutably and stores only touched slot records plus touched object-binding overrides, so transaction atomicity is preserved without introducing a second persistent allocator or changing commit behavior.

## Validation
The branch changes one runtime file and includes focused deterministic counter/state-preservation tests. This environment cannot clone GitHub locally, so required Rust/workspace CI is the compile/test authority for the exact head.

Tracks #113 and #68.

### DIFF
--- a/crates/noon-runtime/src/execution_slots.rs
+++ b/crates/noon-runtime/src/execution_slots.rs
@@ -193,6 +193,101 @@ pub struct ExecutionSlotTable {
     last_mutation: ExecutionSlotMutationStats,
 }
 
+struct ExecutionSlotPreflight<'a> {
+    base: &'a ExecutionSlotTable,
+    free_head: Option<u32>,
+    next_new_slot: usize,
+    touched_slots: HashMap<u32, ExecutionSlot>,
+    object_overrides: HashMap<ObjectId, Option<ExecutionSlotId>>,
+}
+
+impl<'a> ExecutionSlotPreflight<'a> {
+    fn new(base: &'a ExecutionSlotTable) -> Self {
+        Self {
+            base,
+            free_head: base.free_head,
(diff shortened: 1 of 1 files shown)


yongkyuns:
CI status on exact head `3b246a28`: Manim differential, raster differential, API coverage, coverage, cross-browser matrix, web package, and all browser jobs passed. The only failure is `cargo fmt --check`; rustfmt requests exactly one mechanical change at `execution_slots.rs:1064`, collapsing the two-line `MutationTransaction::from_mutations([ScenePatch::RemoveObject(object)])` binding to one line. This connector exposes only whole-file replacement for edits, and the runtime file is large/hot, so I am not risking a full-file rewrite for a one-line formatter-only delta. No semantic/test failure has been observed; clippy/workspace tests are blocked only because CI stops after format. The next safe action is to apply that exact rustfmt diff from a local/edit-capable environment and rerun the Rust job.

PR #491, opened 2026-08-28
Fix demo startup latency and scroll memory pressure

yongkyuns:
## What changed
- defer Python/Pyodide + GPU execution startup until the first explicit Run or example selection instead of warming/autoplaying on page load
- load CodeMirror/Ruff only when the source editor is first focused
- cap gallery DOM/image residency to 18 thumbnails per page with Previous/Next navigation, low-priority image fetches, and `content-visibility`
- replace per-animation-frame metrics work with a 500 ms visibility-aware poll
- stop metrics and runtime resources cleanly on page hide
- make the in-flight runtime-start promise authoritative so concurrent first-run callers cannot observe a half-started execution client
- clear stale restart state after a successful fresh runtime start and keep both gallery pager directions disabled while scene work is busy

## Why
The public demo previously did several high-cost/high-memory tasks together at startup: Pyodide and render workers, editor/linter WASM, first-scene autoplay, plus an effectively unbounded thumbnail corpus available for decoding while scrolling. On memory-constrained browsers this increases initial latency and lets gallery browsing push the tab toward its memory limit.

The new initial state is intentionally lightweight: source/gallery UI is interactive, but Python, execution/render workers, editor enhancement, playback controls, and metrics polling are not started until they are actually needed.

## Architecture / invariants
- lazy startup still uses the existing shared authoring/execution/reconciliation paths; no alternate demo-only renderer or Python execution path is introduced
- once requested, Noon WASM, Pyodide, and the content-addressed Python compatibility bundle retain their parallel startup boundary from #489
- `runtimeStartPromise` is checked before the published `player`, so a second caller waits for the same initialization rather than racing a partially initialized player
- gallery paging limits live thumbnail elements rather than relying only on `loading=lazy`
- metrics work is skipped while hidden, busy, rerunning, or recovering
- page teardown terminates the Python authoring client and execution player and destroys playback/metrics resources

## Validation contract
Source tests require:
- no eager Python/GPU runtime construction or startup autoplay
- no eager CodeMirror/Ruff import
- bounded 18-card gallery rendering
- low-priority/offscreen thumbnail behavior
- 500 ms visibility-aware metrics polling with no metrics RAF loop
- safe concurrent first-run startup and pager busy-state behavior
- preservation of #489's parallel Python/WASM/bootstrap startup

Dedicated browser suites now explicitly prove the new product flow:
1. load the public playground and assert runtime startup is `deferred` with no renderer/playback allocation;
2. assert gallery DOM residency is bounded;
3. start execution through the real Run control;
4. run the existing playback, rerun-race, authoring recovery, resize, lifecycle, and Chromium/Firefox/WebKit workflow assertions unchanged after startup.

The authoring-recovery smoke also focuses the textarea first and waits for the real CodeMirror editor, proving editor enhancement is lazy rather than merely absent.

### DIFF
--- a/scripts/playground-authoring-error-recovery-smoke.mjs
+++ b/scripts/playground-authoring-error-recovery-smoke.mjs
@@ -74,6 +74,7 @@ async function snapshot(page) {
     const patch = document.querySelector("#patch-status");
     return {
       runtimeState: status?.dataset.state ?? "",
+      runtimeStartup: status?.dataset.runtimeStartup ?? "",
       rendererBackend: status?.dataset.rendererBackend ?? "",
       executionMode: status?.dataset.executionMode ?? "",
       presentedFrames: Number(status?.dataset.presentedFrames ?? "0"),
@@ -89,9 +90,21 @@ async function snapshot(page) {
 }
 
 async function waitForInitialScene(page) {
+  await page.waitForFunction(() => window.__noonExampleGallery !== undefined);
+  const deferred = await snapshot(page);
+  assert.equal(deferred.runtimeStartup, "deferred", "authoring runtime must stay deferred on page load");
+  assert.equal(deferred.rendererBackend, "", "deferred authoring page must not initialize the renderer");
+  assert.equal(deferred.presentedFrames, 0, "deferred authoring page must not render frames");
+
+  const sourceTextarea = page.locator("#python-scene-source");
+  await sourceTextarea.focus();
   await page.waitForSelector("#scene-editor-panel .python-code-editor[data-editor-ready='true'] .cm-content", {
     timeout: 30_000,
   });
+
+  const runButton = page.locator("#replace-scene");
+  assert.equal(await runButton.isEnabled(), true, "Run must remain available after lazy editor startup");
+  await runButton.click();
   await page.waitForFunction(
     () => {
       const status = document.querySelector("#status");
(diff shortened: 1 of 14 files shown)


yongkyuns:
Found a lifecycle race that should be fixed before merging this memory-pressure hardening PR.

`ensureRuntimeReady()` publishes the outer `AuthoringExecutionClient` before awaiting `nextPlayer.start()`, so `pagehide` can call `nextPlayer.terminate()` while startup is still pending. But `AuthoringExecutionClient.#startLegacy()` keeps the actual `ExecutionWorkerClient` only in a local variable until `await player.start(...)` resolves; `AuthoringExecutionClient.terminate()` currently only terminates `this.#player`, which is still null during that await. A successful late start can therefore assign the private player after page teardown, then let `ensureRuntimeReady()` continue creating playback state on a runtime that was supposed to be gone. This directly undermines the new pagehide cleanup/memory invariant.

Please make in-flight startup cancellation authoritative (either in `AuthoringExecutionClient` with a startup/termination generation/owned pending player, or with an equally strong top-level teardown epoch that terminates a late completion) and add a browser regression that triggers pagehide/visibility teardown while startup is deliberately pending and proves no worker/player/playback/metrics resources are resurrected afterward. The existing lifecycle smoke starts the deferred runtime to completion before exercising lifecycle behavior, so it does not cover this race.

Issue #114, opened 2026-08-24
[Test T11] Add resource lifecycle, cache eviction, and long-running churn/leak tests


yongkyuns:
## Goal
Catch retained CPU/GPU/browser resource leaks and unbounded cache growth under the edit/reload/churn patterns of an interactive authoring tool.

## Current gap
The renderer has focused cache lifecycle tests, but there is no end-to-end stress suite proving that repeated scene replacement, patching, geometry churn, worker reruns, renderer recreation, and scene switching return resources to a bounded steady state. Short correctness tests can all pass while memory/caches grow every edit.

### Native/resource counters
Expose or reuse test-only counters for:
- live semantic/execution slots;
- geometry/resource arena entries/refcounts once #53 lands;
- renderer path/analytic caches, GPU buffers/textures/pipelines where trackable;
- retired/reusable slots and generations;
- callback/worker registrations.

### Stress scenarios
- load -> render -> replace with unrelated scene, repeated hundreds/thousands of times;
- create/remove the same bounded working set repeatedly;
- alternate between two scenes/resources to verify reuse vs leak;
- repeated local path/style/transform edits;
- rapid example switching and authoring reruns;
- renderer/device/context teardown and recreation;
- failed scene/patch loads followed by successful recovery;
- large temporary geometry followed by a small scene.

### Browser memory diagnostics
Where stable enough, record JS/WASM heap/memory growth diagnostics in a non-flaky dedicated test/profile. Required CI should primarily gate deterministic live-resource counts and bounded retained capacities rather than browser GC timing.

## Required invariants
- live resource counts return to baseline or a documented cache plateau after references disappear;
- caches have explicit retention/eviction policies and cannot grow with every unique transient edit forever;
- stale generations are not retained solely by renderer/worker bookkeeping;
- failed transactions do not leak partially created resources;
- teardown/restart releases old renderer-generation resources.

## Acceptance
- Deterministic churn tests execute at least 1,000 bounded edits/replacements without monotonic live-resource growth.
- Test-only resource snapshots identify which resource class leaked on failure.
- Large temporary scenes do not permanently pin proportional live resources unless a documented cache budget explicitly permits it.
- Browser/WASM memory profiling guidance exists for deeper manual/named-runner checks.
- New resource systems from #53/#59/#65 plug into the same accounting rather than adding opaque caches.

## Dependencies / parallelism
**Wave T1/T2; start with existing caches now.** Expand accounting as #53/#59/#60/#65 land.

[...]

PR #502, opened 2026-08-28
Enforce shared object-state validation in compiled patches

### LINKED ISSUE #499: [Bug] Compiled live object patches bypass finite-state validation
## Problem

`SceneDefinition` rejects non-finite geometry/transform/style state through the shared validators in `noon-core::patch`, but `CompiledScene::apply_patch_with_stats` directly installs `CreateObject`, `SetGeometry`, `SetTransform`, and `SetStyle` payloads without running those validators.

This is observable through `SceneInstance::apply_patch`: a `SetTransform` containing `Vec2::new(f32::NAN, 0.0)` succeeds and can place non-finite state into the live compiled/runtime scene, even though applying the same semantic patch to `SceneDefinition` fails with `PatchError::InvalidObjectState`.

This was exposed while adding #497's sustained failed-patch churn gate: Rust CI failed at iteration 0 because the invalid transform was accepted.

## Architectural fix

Do not duplicate finite checks in `noon-compile`. Make the authoritative object/property validation contract reusable across the core/compile boundary, then validate before any compiled mutation:

- `CreateObject`: validate the complete `ObjectDefinition` before appending a compiled slot.
- `SetGeometry` / `SetTransform` / `SetStyle`: validate the incoming object state before touching the compiled object.
- compiled transaction preflight should enforce the same contract for structural/property mutations.
- map the shared invalid-object-state error into an explicit `CompilePatchError` variant rather than misclassifying it as a timeline or identity error.

The validation must remain O(payload), preserve property-edit O(1) scene locality, and happen before mutation so rejection is transactional.

## Acceptance

- equivalent invalid object-state patches are rejected consistently by `SceneDefinition`, `CompiledScene`, and `SceneInstance`;
- non-finite geometry, transform, style, and invalid created-object state are covered;
- rejection leaves compiled/runtime state, dirty indices, patch stats, slot/accounting state unchanged;
- no duplicate finite-validation implementation is introduced in `noon-compile`.

Related: #114, #497.

yongkyuns:
## Summary
- expose the existing `noon-core` whole-object and property-patch validators as the shared semantic admission contract
- enforce that contract in `CompiledScene` for `CreateObject`, `SetGeometry`, `SetTransform`, and `SetStyle`
- enforce the same contract during compiled transaction preflight
- map non-finite object payloads to an explicit `CompilePatchError::InvalidObjectState` without duplicating geometry/transform/style validation rules
- add compiled-state and live-runtime regressions for transactional rejection and recovery

## Invariants
- duplicate/unknown identity and object-capacity errors keep their existing precedence
- validation happens before any compiled slot, payload, live-count, frame, dirty-state, or patch-stat mutation
- object-local property edits remain O(patch payload); no scene scan or alternate mutation path is introduced
- `SceneDefinition`, `CompiledScene`, and `SceneInstance` now share one authoritative finite-state contract

## Regression coverage
- non-finite geometry, transform, and style patches preserve compiled state exactly
- invalid object creation allocates no slot and changes no live accounting
- a later invalid property in transaction preflight rejects atomically
- the exact NaN live transform that exposed the bug leaves frame state, dirty indices, and `RuntimePatchStats` untouched, then a valid local patch still succeeds

Fixes #499. Related to #114.

### DIFF
--- a/crates/noon-compile/src/lib.rs
+++ b/crates/noon-compile/src/lib.rs
@@ -268,7 +268,10 @@ impl PartialEq<Vec<CompiledTrack>> for CompiledTracks<'_> {
 pub enum CompileError {
     TooManyObjects(usize),
     UnknownObject(ObjectId),
-    DiscontinuousPresence { previous: TrackId, next: TrackId },
+    DiscontinuousPresence {
+        previous: TrackId,
+        next: TrackId,
+    },
     UnsupportedTransformGeometry(TrackId),
     PathTransformRequiresRetessellation(TrackId),
     UnsafeFilledPathTransform(TrackId),
--- a/crates/noon-compile/src/lib.rs
+++ b/crates/noon-compile/src/lib.rs
@@ -268,10 +268,7 @@ impl PartialEq<Vec<CompiledTrack>> for CompiledTracks<'_> {
 pub enum CompileError {
     TooManyObjects(usize),
     UnknownObject(ObjectId),
-    DiscontinuousPresence {
-        previous: TrackId,
-        next: TrackId,
-    },
+    DiscontinuousPresence { previous: TrackId, next: TrackId },
     UnsupportedTransformGeometry(TrackId),
     PathTransformRequiresRetessellation(TrackId),
     UnsafeFilledPathTransform(TrackId),
@@ -325,7 +322,10 @@ pub enum CompilePatchError {
         field: ObjectStateField,
     },
     InvalidTrack(TimelineError),
-    DiscontinuousPresence { previous: TrackId, next: TrackId },
+    DiscontinuousPresence {
+        previous: TrackId,
+        next: TrackId,
+    },
     UnsupportedTransformGeometry(TrackId),
     PathTransformRequiresRetessellation(TrackId),
     UnsafeFilledPathTransform(TrackId),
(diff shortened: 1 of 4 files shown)


yongkyuns:
Exact-head Rust CI reached `cargo fmt --all -- --check` and failed only on rustfmt; Clippy/tests were skipped. Required formatter output is mechanical:

- `CompilePatchError::DiscontinuousPresence { previous: TrackId, next: TrackId }` expands to the standard multiline struct variant after the new multiline `InvalidObjectState` variant.
- `object_state_patch_validation.rs` wraps the long `use noon_core::{...}` import after `ObjectStateField` and expands the long `assert_eq!(compiled, before, ...)` onto multiple lines.

No semantic failure was observed. Please apply `cargo fmt --all` exactly, then refresh/rebase onto current master (which now includes #500 and #503) and let the full Rust gate run. I reviewed the semantic diff already: shared validators, error precedence, pre-mutation validation, atomic preflight, and runtime recovery all look sound.

yongkyuns:
The latest formatting commit fixed `object_state_patch_validation.rs`, but the other rustfmt diff from the failed Rust job is still present on head `a14f898`: `CompilePatchError::DiscontinuousPresence` remains the one-line struct variant in `crates/noon-compile/src/lib.rs`.

The failed formatter required exactly:
```rust
    DiscontinuousPresence {
        previous: TrackId,
        next: TrackId,
    },
```

Please apply that formatter-only change as well. No semantic change is needed; Clippy/workspace tests never ran on the earlier head because format stopped the job first.

Issue #254, opened 2026-08-26
[Roadmap B7/D4] Example Gallery — Camera and 3D scenes


yongkyuns:
## Roadmap

B7/D4 camera/3D acceptance under #954/#956. Architecture authority: `docs/architecture.md`.

## Goal

Use the pinned ManimCE v0.21 Special Camera Settings examples as product-level acceptance for shared B6 2D camera and D1-D4 3D behavior. The examples validate the architecture; they do not define another camera/scene model.

## Required cases
- `FollowingGraphCamera` — B6/#89 plus plotting/animation dependencies;
- `MovingZoomedSceneAround` — B6/#89 plus image/text/compositing dependencies;
- `FixedInFrameMObjectTest` — D4 mixed fixed-in-frame 2D/3D;
- `ThreeDLightSourcePosition` — D3 lighting/material/surface capability;
- `ThreeDCameraRotation` — D4 camera helpers;
- `ThreeDCameraIllusionRotation` — D4 or explicit intentional divergence;
- `ThreeDSurfacePlot` — D2/D3 surface/depth/material capability.

## Camera/state policy

- authored camera identity/projection parameters belong to the Semantic Scene;
- timeline/reactive camera motion is observed as effective Runtime camera state at the sampled `FrameEpoch`;
- native/web renderers consume the same effective camera through the shared execution/session path;
- camera movement must never be emulated by Python-moving every object;
- viewport/backend projection details may differ internally but must preserve the same observable camera result;
- no separate 3D scene/runtime/canvas or frontend projection authority.

## Qualification

For each promoted case require:
- source-equivalent behavior against pinned ManimCE v0.21;
- authored scene/family/camera declarations correct where relevant;
- effective camera/object state correct at begin/intermediate/end and discontinuity-sensitive times;
- exact lifecycle/timing/rate-function semantics;
- direct seek == forward playback where deterministic;
- reviewed raster/depth/compositing equivalence on supported backends;
- explicit replay classification for any host-dynamic behavior;
- no target-specific camera semantics or demo-only compensation.

## Acceptance

Every required case remains represented and is parity-qualified or explicitly blocked/deferred by its current B6/D owner. Camera qualification distinguishes authored/base state from current effective Runtime state.

## Dependencies

B6/#89 for 2D camera, D1/#698 and D2/#699 for 3D substrate, later D3/D4 under #956/#90, and B7 evidence #185/#109.

Issue #185, opened 2026-08-25
[Roadmap B7] ManimCE v0.21 exact output convergence


yongkyuns:
## Roadmap

Phase B7 qualification tracker under #954. Architecture authority: `docs/architecture.md`.

## Goal

For the pinned ManimCE v0.21 subset Noon claims to support, require the same source-level scene to produce equivalent authored structure, live/effective observables, timing and reviewed raster output within explicit tolerances.

“Runs” or “looks similar” is not compatibility. Qualification must test the shared engine rather than compensate in examples/frontends.

## Qualification contract

For every source-equivalent fixture:
- pin ManimCE version/reference renderer, viewport/DPR/FPS/background/resources;
- compare exact authored times rather than loosely corresponding video frames;
- compare **authored semantic declarations/structure** where observable: identity/membership/order/source/options/lifecycle declarations;
- compare **effective Runtime observables** at coherent sampled states: transforms, bounds/centers, style, content/resource result, camera, visibility and other live getters;
- sample begin/intermediate/final/end and discontinuity-sensitive times;
- ensure each sampled effective result comes from one coherent frame/publication context rather than mixed revisions;
- compare lifecycle/segment completion semantics, not merely final pixels;
- use semantic numeric tolerances only where mathematically required and explicit raster tolerances only for renderer differences;
- retain Manim, Noon backend and diff artifacts on failure;
- classify failures by owning B1-B6/D/C layer rather than adding fixture-specific constants.

## Determinism/replay

- deterministic/seekable cases require direct seek == forward playback for effective observables;
- host-dynamic/external cases must declare replay/seek limitations explicitly and use recorded/replayed inputs where deterministic qualification is claimed;
- `play()`/`wait()` qualification checks the logical completion barrier before querying final effective state;
- no test may accidentally read authored/base values where the compatibility API promises current effective values.

## Initial corpus

At minimum maintain source-equivalent pinned Quickstart cases:
1. `CreateCircle`;
2. `SquareToCircle`;
3. `SquareAndCircle`;
4. `AnimatedSquareToCircle`;
5. `DifferentRotations` once supported.

Expand through #91/#251-#256 as capabilities land.

## `parity-qualified`

A supported fixture is parity-qualified only when applicable checks prove:
- same authored duration/play/wait/lifecycle boundaries;
- same relevant authored semantic structure;
- same coherent effective state at representative/discontinuity-sensitive times;
- same effective camera/frame composition;
[...]

Issue #90, opened 2026-08-24
[Roadmap D] ThreeDScene, 3D mobjects, and vector-space capability


yongkyuns:
## Roadmap
Phase D capability parent under #956. Architecture authority: `docs/architecture.md`.

## Goal
Track ManimCE 3D/vector-space product breadth after the shared D1/D2 substrate exists. This issue does not define a separate 3D architecture.

### 3D scene/camera
- `ThreeDScene` public surface;
- orientation/movement/ambient rotation helpers;
- fixed-in-frame/fixed-orientation behavior;
- perspective/depth/occlusion semantics.

### 3D mobjects
- `ThreeDAxes`;
- `Surface` and generated solids such as sphere/cube/prism/cylinder/cone/torus/polyhedra;
- representative 3D point/line helpers.

### Vector-space scenes
- `VectorScene`;
- `LinearTransformationScene` and related matrix/vector helpers.

## Architecture constraints
All work uses the same:
- Semantic Scene identity/family/lifecycle;
- high-precision world transforms;
- animation/reactive/mutation model;
- Execution Plan and Runtime;
- retained resource ownership;
- renderer device/surface.

A mesh/depth/material lane is a renderer capability, not another scene engine. Fixed-in-frame content is a compositing domain within one scene/presentation surface. Frontends do not own camera projection, mesh generation, lighting or depth ordering.

## Sequencing
1. #698 — D1 semantic world/camera/mesh resource substrate.
2. #699 — D2 depth-aware retained mesh renderer and first qualified 3D fixture.
3. D3 — surfaces/solids/materials/lighting.
4. D4 — mixed 2D/3D composition and camera helpers; gallery cases in #254.
5. D5 — vector-space and broader advanced scene APIs.

Do not expose broad compatibility names before the underlying shared semantics and renderer capability are real and qualified.

## Acceptance
- supported 3D APIs compose the D1/D2 substrate instead of creating specialized engines;
- direct seek and forward playback agree for deterministic camera/object animation;
- static mesh resources remain retained and local edits remain local;
- mixed 2D/3D uses one authored scene and presentation surface;
- every promoted compatibility slice has semantic/timing/raster evidence through B7-style qualification;
- unsupported APIs are explicit rather than approximated.

## Non-goal
No 2D roadmap work waits for this capability parent.

yongkyuns:
### Acceptance criteria amendment — exact Manim output parity for any supported 3D/vector-space slice

[...]

Issue #492, opened 2026-08-28
[Arch] Preserve one render surface across authoring mode switches


yongkyuns:
## Goal
Remove the remaining reload-like canvas remount when Python authoring switches between legacy execution and mixed retained execution.

#453 now preserves the live render worker / OffscreenCanvas when only the execution engine fails. The remaining normal-path surface break is `AuthoringExecutionClient` mode transition: both `#rebuildRetained()` and `#rebuildLegacy()` call `#replaceTransferredCanvas()`, which terminates the player, clones/replaces the HTML canvas, reconnects `ResizeObserver`, and boots a different render worker.

## Current architecture
`AuthoringExecutionClient` selects between two complete worker pairs:
- `ExecutionWorkerClient` -> `execution-engine-worker.js` + `execution-render-worker.js`
- `RetainedExecutionWorkerClient` -> retained engine + retained render worker

Both render workers own the transferred `OffscreenCanvas`, so the current code cannot hand the same transferred canvas from one render worker to the other. The DOM clone is therefore an architectural consequence, not a UI-only bug.

The renderer implementations are also much closer than the worker split suggests:
- legacy `ExecutionCanvasRenderer` and retained `RetainedExecutionCanvasRenderer` both own a WGPU surface/device and drive the shared GPU renderer;
- retained execution is already a **mixed geometry + retained-text** path and uses `GpuRenderer` + `RetainedFramePreparer`;
- retained resource installation is immutable and separate from ordinary per-frame deltas.

That makes a persistent surface owner preferable to teaching the UI to hide canvas replacement.

## Target architecture
One browser render-surface owner should acquire the `OffscreenCanvas` exactly once and survive:
- legacy -> retained authoring transitions;
- retained -> legacy transitions;
- execution-engine replacement;
- ordinary reruns/reconciliation;
- pause/seek/restart-playback state changes.

Only actual unrecoverable render-worker/surface loss should require replacing the HTML canvas.

Preferred direction:

```text
HTMLCanvasElement
       |
       | transfer once
       v
Persistent RenderSurfaceWorker
       |
       +-- generation/mode-neutral render transport
       |
       +-- attach legacy engine
       |        or
       +-- attach mixed-retained engine
```

Avoid a main-thread proxy that copies rendered frames between canvases and avoid keeping two GPU surfaces hidden behind one UI element.

### 1. Separate surface lifetime from execution-mode lifetime
[...]

PR #534, opened 2026-08-29
Define staged Manim 3D parity architecture

yongkyuns:
## Summary
- choose the #90 compatibility target: a source-compatible common ManimCE v0.21 3D subset, expanded only after exact qualification
- define shared semantic camera/world-transform/mesh-resource boundaries instead of a second 3D scene engine
- keep one render surface/device and add depth-aware world + fixed-overlay compositing domains
- define explicit staged policies for depth, transparency, lighting/materials, mixed 2D/3D scenes, backend parity, dirty-work performance, and exact-output qualification
- provide a five-stage implementation sequence beginning with numeric camera/mesh contracts before broad API exposure

## Key decisions
- supported 3D APIs remain Manim-oracle compatible; unsupported behavior is deferred rather than approximated
- immutable mesh resources carry topology/heavy data; transforms/material/camera remain retained mutable state
- opaque world rendering is depth-resolved; fixed-in-frame objects remain the existing 2D overlay domain
- arbitrary translucent intersection/order-independent transparency is not claimed by the first tranche
- WebGPU and WebGL consume the same semantic camera/mesh/material contract
- no Python-owned projection/mesh sampling, second canvas, second scene graph, or 3D-special worker topology

## Why
#90 explicitly requires the compatibility/renderer decision before broad 3D implementation. This keeps 3D from distorting the current common-2D architecture while giving later implementation PRs concrete boundaries and exit criteria.

## Validation
Documentation-only architectural slice. The branch is directly based on current master and adds one new document; no product/runtime files are changed. GitHub CI remains the repository-wide validation authority.

Tracks #90. Related: #254, #185, #176.

### DIFF
--- a/docs/three-d-parity-architecture.md
+++ b/docs/three-d-parity-architecture.md
@@ -1,0 +1,294 @@
+# Manim 3D parity architecture
+
+This document records the architecture decision required by #90 before broad 3D implementation begins. It deliberately defines contracts and staging only; it does not add 3D behavior to the current 2D renderer.
+
+## Decision
+
+Noon will target a **source-compatible common ManimCE v0.21 3D subset**, with unsupported APIs explicitly classified as deferred or intentionally divergent until their observable semantics are implemented and qualified.
+
+This is option 2 from #90. It avoids two failure modes:
+
+- claiming full Manim 3D compatibility before camera, depth, lighting, transparency, and surface semantics are proven;
+- inventing a Noon-native 3D object/camera model that later makes source compatibility unnecessarily difficult.
+
+For every API Noon marks supported, Manim-visible semantics remain the oracle under #176/#185. The implementation is free to use a GPU-native retained architecture internally; source compatibility does not imply reproducing Manim's internal renderer.
+
+The first supported tranche is intentionally narrow:
+
+1. `ThreeDScene` camera state and deterministic camera animation;
+2. `ThreeDAxes`;
+3. one immutable indexed mesh/surface representation;
+4. one basic shaded `Surface` fixture;
+5. fixed-in-frame 2D overlay composition.
+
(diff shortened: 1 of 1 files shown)


yongkyuns:
Architecture review blocker before merge: the document says the proposed `WorldTransform` is capable of ordinary affine 3D transforms, but `translation + quaternion + Vec3 scale` only represents TRS. It cannot represent shear/general linear transforms (relevant to Manim matrix/apply_matrix-style semantics), and non-uniform/general linear transforms also require an explicit inverse-transpose normal transform policy.

Please resolve the contract one of two ways before implementation builds on it:
1. explicitly define the first 3D tranche as TRS-only and classify arbitrary affine/matrix transforms as deferred; or
2. make the canonical linear component general enough (e.g. deterministic 3x3 linear transform + translation, optionally with ergonomic TRS decomposition/adapters), with renderer matrices and normal transforms derived from it.

The rest of the one-scene/one-device, retained mesh resource, world-vs-fixed-overlay, depth, transparency staging, and WebGPU/WebGL parity direction looks sound.

yongkyuns (commented):
Architecture direction is sound. One wording correction before merge: the invariant “one render-surface owner and one GPU device lifetime” is stronger than Noon’s existing recovery contract and can be read as forbidding device recreation after loss. Please phrase this as one active render-surface/device ownership domain (or one active device generation) with controlled device recreation on recovery. 3D should reuse that same recovery/generation boundary rather than creating a second device owner; it should not require an immortal GPU device.

yongkyuns (commented):
Architecture is sound overall, but please correct the invariant `one render-surface owner and one GPU device lifetime`. Noon already supports device-loss/recovery generations, so the invariant should be one active GPU surface/device owner per renderer generation, with explicit replacement on recovery. Otherwise this memo contradicts the existing recovery contract and the #492/#546 direction.

PR #537, opened 2026-08-29
Preflight authoring engines before renderer handoff

yongkyuns:
## Summary
- stage the next legacy/retained engine against an unattached `MessageChannel` before mutating the live render owner
- wait for candidate engine readiness, which both engine implementations emit only after their initial transport/resources/snapshot have been queued
- keep the old engine and renderer fully authoritative during that preflight window, including serving in-flight state/metrics requests
- only after successful preflight publish the candidate engine, retire the old engine, and send `switch_engine` / `rebuild_engine` to the persistent render worker
- if candidate initialization fails before handoff, terminate only the candidate and leave the old engine, renderer, mode, canvas, session, and recovery state intact
- keep the existing conservative full-surface recovery boundary once renderer handoff has actually begun
- cancel unpublished candidates through the existing lifecycle generation on `terminate()` without double-terminating workers

## Why
#517 removed DOM canvas/render-worker replacement from normal legacy ↔ retained authoring transitions, but it still retired the active engine/renderer before replacement engine bootstrap completed. This removes engine/WASM/resource/snapshot preparation from that presentation-risk window and adds a real rollback boundary before the renderer is touched.

The staging mechanism uses the browser's existing `MessageChannel` queue rather than a second render transport or frame proxy. For retained execution, `retained_resources`, optional shared `transport_setup`, and the first complete snapshot are posted before the engine's main-thread `ready` message. Legacy execution likewise posts its initial snapshot before `ready`.

## Validation contract
New deterministic worker-client tests prove:
- the render worker receives no `switch_engine` before candidate readiness
- the old engine remains live and can answer requests during candidate preflight
- old engine retirement happens only after candidate readiness
- a candidate bootstrap failure leaves old engine/render/canvas/mode authoritative and usable
- `terminate()` cancels the unpublished candidate and does not resurrect either generation

The existing real dual-transport authoring mode-switch browser oracle, authoring router/lifecycle races, renderer recovery, playback, resize, and cross-browser suites remain the end-to-end gates.

Tracks #492.

### DIFF
--- a/scripts/build-web-demo.sh
+++ b/scripts/build-web-demo.sh
@@ -69,6 +69,7 @@ node --test scripts/manim-reference-inventory.test.mjs
 node --test web/example-gallery.test.mjs
 node --test web/execution-transport.test.mjs
 node --test web/execution-worker-client.test.mjs
+node --test web/execution-worker-preflight.test.mjs
 node --test web/render-gpu-diagnostics.test.mjs
 node --test web/execution-worker-startup.test.mjs
 node --test web/authoring-client.test.mjs
--- a/web/execution-worker-client.js
+++ b/web/execution-worker-client.js
@@ -20,6 +20,8 @@ const LIFECYCLE_CANCELLED_MESSAGE =
 export class ExecutionWorkerClient {
   #canvas;
   #engineWorker = null;
+  #candidateEngineWorker = null;
+  #candidateEngineReject = null;
   #renderWorker = null;
   #nextRequestIds = { engine: 0, render: 0 };
   #pending = new Map();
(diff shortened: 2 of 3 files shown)


yongkyuns:
Concurrency blocker found in the new preflight state.

`#transitionEngine()` does not increment `#lifecycleGeneration`, and `#mode` is not updated until the transition completes. Therefore two overlapping public transitions (e.g. two `switchToRetained()` calls while the first candidate is still booting) can both enter preflight with the same lifecycle generation.

The second call overwrites the single `#candidateEngineWorker` and `#candidateEngineReject`. The first candidate's readiness/error listeners then see `#candidateEngineWorker !== firstCandidate` and ignore its events, while its cancellation callback has also been overwritten. That first `await candidateReadyPromise` can remain pending indefinitely and its candidate can be stranded.

Please make transition ownership explicit before merge. Good options are:
- serialize engine transitions behind one in-flight transition promise; or
- introduce a per-transition generation/token that supersedes and terminates/rejects the previous candidate before installing the next one (while keeping the committed old engine authoritative until the winning candidate is ready).

Add a deterministic test that starts transition A, starts transition B before A emits `ready`, then proves A settles as cancelled, only B can commit, no candidate leaks, and `terminate()` still settles the winner exactly once.

The pre-ready message-order assumption itself checks out: both engine implementations send initial transport/resources/snapshot over the render port and emit only main-thread `ready` afterward.

PR #572, opened 2026-08-29
Add bounded live glyph atlas pages

yongkyuns:
## Summary
- move live `GpuGlyphAtlas` ownership from one texture per plane to bounded page vectors backed by the existing deterministic page allocator
- preserve `GpuGlyphAtlas::new(extent)` as the existing one-page-per-plane policy
- add explicit `with_page_limit(extent, max_pages_per_plane)` opt-in without changing the retained renderer's default GPU memory budget
- preserve `texture_view(plane)` as page-0 compatibility and add page-aware view lookup/counts
- make retained text GPU bind groups page-aware and select the bind group from `PreparedTextItem::GlyphBatch.page`
- add noop-GPU regressions that force page 1 allocation and submit a draw using that page

## Invariants
- default production behavior still allocates at most one texture per plane
- page allocation remains deterministic and bounded; no eviction/reuse policy is introduced here
- textures remain lazy and are allocated only when a page is reached
- nonzero pages can no longer be rejected merely because the renderer has a single bind group
- current-frame pinning/eviction, metadata plateau counters, and product multi-page defaults remain the next isolated #363 slice

Tracks #363. Built directly on merged #562 (`bc1a8986`).

### DIFF
--- a/crates/noon-text-atlas/src/lib.rs
+++ b/crates/noon-text-atlas/src/lib.rs
@@ -508,11 +508,7 @@ impl GpuGlyphAtlas {
             image.placement.height,
             bytes_per_pixel,
         )?;
-        let allocation = self.allocate(
-            plane,
-            image.placement.width,
-            image.placement.height,
-        )?;
+        let allocation = self.allocate(plane, image.placement.width, image.placement.height)?;
         let extent = self.extent;
         let state = self.ensure_page(device, plane, allocation.page)?;
         let bytes_per_row = allocation.outer_size[0]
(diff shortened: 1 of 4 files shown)


yongkyuns:
Validation update after the first main-CI run:

- head `a49cfc9e…` passed Web package, both browser rendering backends, browser authoring/reactive/editor, coverage, Manim API/semantic/raster gates, cross-browser, authoring mode-switch, and platform/release validation;
- the only failing job was Rust, which stopped at `cargo fmt --all -- --check` before Clippy/tests;
- the reported diff was exactly three rustfmt-only hunks in `noon-text-atlas/src/lib.rs` and `noon-text-render-wgpu/src/gpu.rs`;
- current head `09adfc35…` is one commit ahead, `Apply rustfmt to live atlas paging`, and its diff is exactly those three formatter changes with no semantic code change.

Fresh exact-head workflows are now queued/running. Keep this unmerged until required CI completes on `09adfc35…`; no functional workaround or tolerance change is needed.

Issue #569, opened 2026-08-29
[Roadmap C3] Spatial queries for interaction and renderer culling


yongkyuns:
## Roadmap

Phase C3 child of #955. Architecture authority: `docs/architecture.md`.

## Goal

Use one execution-owned retained spatial index for interaction broad phase and renderer viewport candidates while keeping coherence and locality explicit.

## Architecture

```text
retained live execution state
        +
execution-owned spatial index
        |
        +--> interaction candidates
        +--> viewport candidates
```

Do not add frontend bounds scans, renderer-owned spatial trees or another semantic/object index.

## Revision contract

Spatial/candidate results are derived runtime data. Each result must be valid for the exact context it was computed from:

- `ExecutionRevision` or equivalent derived-execution revision;
- `FrameEpoch` for the effective runtime state;
- relevant semantic/execution identity generations;
- camera/viewport state incorporated into that frame context.

Do not use one ambiguous global “generation” value for all of these domains. Stale candidates must be rejected rather than applied to a newer frame.

### C3.1 — interaction candidates
- point/region queries for hover/selection;
- painter-correct candidate order;
- precise geometry hit testing only on candidates;
- no full semantic/frame scan for ordinary pointer events.

### C3.2 — renderer viewport candidates
- derive visible execution slots from the same retained index;
- conservative bounds are broad phase only; renderer clipping remains final;
- preserve painter order among candidates;
- offscreen live objects keep healthy immutable resource residency.

### C3.3 — local effective updates
Timeline/reactive/host/editor effective transform changes refit only affected entries. They may advance `FrameEpoch` without pretending the Semantic Scene changed or requiring semantic relowering.

### C3.4 — camera changes
Camera movement changes candidate projection/query results without marking every object semantically mutated.

### C3.5 — diagnostics
Expose candidate count, live count, refit count, submitted instances/batches and explicit full-index rebuilds.

## Acceptance

With a 100k mostly-static scene and a small visible fraction:
- viewport projection/submission scales with candidates rather than all-live objects;
- one effective object transform refits bounded spatial state;
- selection/hover uses the same index;
- candidate order is painter-correct;
- stale execution/frame/identity contexts are rejected;
- no duplicate frontend/renderer spatial index exists;
- WebGPU/WebGL/native consumers agree on candidate semantics;
- any deliberate full rebuild is explicit maintenance, not a hidden local-edit fallback.

## Dependencies

[...]

Issue #66, opened 2026-08-24
[Arch I1] Add retained spatial index for hit testing and viewport culling


yongkyuns:
## Goal
Make interaction and visibility queries scale sublinearly with scene size.

## Scope
- Add renderer-independent execution bounds suitable for spatial indexing.
- Build a retained BVH/R-tree or equivalent for static objects plus incremental updates for dynamic leaves.
- Use the index for native hit testing, selection/hover candidates, and viewport culling.
- Separate cheap conservative culling bounds from layout-accurate semantic bounds.
- Preserve painter/z ordering among returned hit candidates.
- Add counters for index rebuilds, leaf updates, and candidate counts.

## Acceptance
- Pointer hit test in a 100k-object scene does not scan 100k objects.
- Transforming a small number of dynamic objects updates only affected leaves.
- Viewport changes query the index without semantic scene scans.
- Correctness tests compare indexed results to a brute-force reference corpus.

## Dependencies
**Wave 2. Consume stable execution slots from #58 and bounds semantics from #62. Can be implemented independently of browser transport.**

yongkyuns:
Current master now satisfies most of the retained-index mechanics, but this issue should remain open.

Completed evidence:
- `ExecutionSpatialIndex` is retained and incrementally updated from `SlottedSceneInstance`; transform/timeline changes refit affected leaves without a full rebuild.
- `point_query_in_hundred_thousand_objects_does_not_scan_scene` gates the 100k-object sublinear hit-test requirement (`cells_visited == 1`, bounded candidates, no full-scan fallback).
- viewport queries are cell-local and preserve painter order; update/query counters are exposed.
- #550 added a 1,000-generation slot-reuse regression so stale execution generations cannot alias a replacement spatial leaf.

Remaining acceptance/scope I could not find on current master:
1. No brute-force reference corpus compares indexed hit/viewport results against an independent oracle across varied geometry/transforms.
2. Repository search finds `hit_test` / `query_viewport` consumption only inside `noon-runtime`; browser selection/hover and renderer viewport-culling paths do not appear to consume the retained index yet.

So the next #66 work should be split into (a) renderer-independent brute-force differential correctness tests, then (b) real interaction/culling consumers, rather than adding more index internals.

yongkyuns:
Follow-up split is now explicit:

- #566 covers the remaining renderer-independent brute-force differential oracle acceptance.
[...]

PR #580, opened 2026-08-29
Expose retained spatial query JSON from ScenePlayer

yongkyuns:
## Summary
- add `ScenePlayer::hit_test_json` and `ScenePlayer::query_viewport_json` as the browser-facing serialization seam for the existing retained spatial index
- serialize generation-safe execution slot identity as `{slot, generation}` without converting back to semantic IDs or frame-row indices
- include retained query metrics (`cells_visited`, `candidates_tested`, `results`, `full_scan_fallbacks`) so interaction/renderer consumers can prove broad-phase locality
- preserve painter-order result ordering from `ExecutionSpatialIndex`
- add focused native tests for overlapping painter order, generation fields, viewport results, and no full-scan fallback

## Architecture
This does not create another spatial index or another player. The methods live on the existing authoritative `ScenePlayer` and directly serialize `SpatialQueryResult` from `SlottedSceneInstance`.

Serialization is isolated in a small internal module rather than embedded in the large GPU/WASM wrapper. The next thin #569 slice can expose these two methods on the existing `ScenePlayer` / `NoonCanvasPlayer` wasm-bindgen classes without duplicating slot/metric mapping.

## Scope
No renderer culling behavior or frontend pointer UI is changed here. This is the reusable query boundary that both consumers need.

Tracks #569 and #66.

### DIFF
--- a/crates/noon-web/src/lib.rs
+++ b/crates/noon-web/src/lib.rs
@@ -25,6 +25,7 @@ mod retained_execution_transport;
 mod retained_resource_transport;
 mod retained_typst_canvas;
 mod semantic_snapshot;
+mod spatial_query;
 
 pub use authoring_facade::*;
 pub use authoring_mobject::*;
--- a/crates/noon-web/src/spatial_query.rs
+++ b/crates/noon-web/src/spatial_query.rs
@@ -1,0 +1,100 @@
+//! Browser-facing serialization for retained spatial-index queries.
+//!
+//! Inputs are Noon world coordinates. Results preserve retained painter order and
+//! opaque execution-slot generations so browser consumers never need a scene scan or
+//! semantic-ID/frame-row remapping.
+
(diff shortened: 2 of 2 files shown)


yongkyuns:
Rebased cleanly onto current master `ab8de4712430546d11d0d0dd949d03909b46806b` after #585/#579 advanced the base. I verified the intervening master changes touched only `noon-core` resource/reactive files and the retained text locality test; neither `crates/noon-web/src/lib.rs` nor the new `spatial_query.rs` overlapped.

The original two-file behavior was replayed unchanged, with one module-level contract comment clarifying that inputs are Noon world coordinates and outputs preserve painter order plus opaque slot generations. Current exact head is `7781b724adf2c5bde524f2991e96651668cb37ca` and fresh PR workflows are now queued/pending on that head. Keeping unmerged until exact-head CI completes.

PR #585, opened 2026-08-29
Gate retained outline object locality

yongkyuns:
## Summary
- extend the existing #362 large-scene text locality gate with the remaining resident outline-path case
- warm 10,000 retained text objects sharing one immutable Typst resource, with exactly one object already in the outline lane via partial reveal
- update only that outline object's translation/opacity for 128 logical frames
- require exactly one outline run throughout, one object-local update per frame, zero fallback rebuilds, and bit-for-bit unchanged raster/atlas accounting after warmup

## Architecture
This is a deterministic structural regression only. It does not add renderer behavior, timing thresholds, cache policy, worker transport, or backend-specific tolerances. It specifically protects the already-supported case where an object remains resident in the outline lane while presentation state changes.

Changing `reveal`/`morph` so an object crosses glyph ↔ outline lane is still structurally incompatible with `RetainedTextQuadPreparer::can_update_objects` and intentionally falls back today; that remains substantive #362 work rather than being hidden by this gate.

Tracks #362 and #68.

### DIFF
--- a/crates/noon-text-render-wgpu/tests/retained_object_update_locality.rs
+++ b/crates/noon-text-render-wgpu/tests/retained_object_update_locality.rs
@@ -1,7 +1,7 @@
 use noon_core::{ObjectContentRef, ObjectId, Style, TextResourceArena, Transform2D, Vec2};
 use noon_runtime::{FrameChanges, RetainedFrameObjectState, RetainedFrameState};
 use noon_text_render_wgpu::{
-    RetainedTextIncrementalStats, RetainedTextQuadPreparer, TextDeviceMetrics,
+    PreparedTextItem, RetainedTextIncrementalStats, RetainedTextQuadPreparer, TextDeviceMetrics,
 };
 use noon_typst::{compile_typst_resource, TypstMode};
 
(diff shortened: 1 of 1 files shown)


yongkyuns (commented):
Follow-up design note: the remaining glyph↔outline lane transition is intentionally not folded into this test-only PR. `RetainedTextQuadPreparer` currently stores global `items`/quad arrays plus per-object ranges; a lane transition can change an object's item/quad cardinality and therefore invalidate subsequent ranges. A correct incremental implementation should first move toward retained per-object prepared records (or an equivalent indirection/range allocator) so one object's structural text-lane change does not require shifting/reindexing the whole frame. Avoid an in-place vector-splice workaround here; that would preserve O(total trailing records) behavior and work against #362's target model.

yongkyuns (commented):
Merged exact green head `f9c9bedde4168f26fe8712d95fc1e6840ff50127` as `94788e8708434a814678f2b95b0b2530ee47d19b`. This closes the resident-outline locality acceptance slice only; glyph↔outline lane transitions still require object-local prepared-record ownership rather than global array splicing.

Issue #573, opened 2026-08-29
[Arch C3.4] Add localized resource mutation transaction transport


yongkyuns:
Parent: #368
Architecture parent: #361
Depends on: #570 core version-aware mutation primitive; generation-safe text IDs from #567
Related: #362, #64, #114

## Goal
Carry one-object retained resource/content replacement across the execution/runtime boundary as an explicit transaction, without resending or reconstructing unrelated scene/resources.

## Scope
- define a narrow retained resource mutation transaction for add/replace/remove resource operations plus the affected semantic/content handle
- carry exact resource ID generation and version across the boundary
- apply dependency installs before content-handle switch, then retire old dependencies only after commit
- reject stale/out-of-order expected handles deterministically
- preserve rollback if any compile/decode/install/validation step fails
- distinguish content-dirty from ordinary presentation-dirty updates at the runtime boundary
- keep ordinary transform/style frame deltas resource-payload-free

## Invariants
- one text source edit touches only the affected content slot plus newly required dependencies
- unrelated text/font/vector handles and retained runtime state remain bit-for-bit stable
- failed transactions leave the previous live scene/resources intact
- stale transactions cannot overwrite a newer resource generation/version
- no full-scene resource bundle resend is required after initial session install
- transport remains generic over immutable retained resources; text is the first consumer, not a special-case mutation engine

## Acceptance
- focused transaction tests for successful one-resource replacement, stale expected-version rejection, malformed/new-resource failure rollback, and dependency-install failure rollback
- mixed-scene regression proving unrelated resource handles and object content handles do not change
- serialized round-trip preserves full opaque resource IDs and handle versions
- repeated replacements cooperate with #114 churn/lifetime guarantees

## Non-goals
- no shaping/layout redesign
- no cache eviction policy (#363)
- no authoring-scene migration (#367)
- no compatible-part semantic reconciliation beyond exposing the committed content identity for #64

This should remain a protocol/runtime slice; renderer prepared-record invalidation can be a following independent tranche once the committed transaction boundary is stable.

yongkyuns:
#579 has merged as `ab8de4712430546d11d0d0dd949d03909b46806b`, closing the core atomic transaction foundation for this issue.

Completed boundary:
- all-or-nothing multi-replace/remove transactions over retained text resources
- generation/version-aware stale-target rejection
- duplicate-target rejection
- complete payload/dependency preflight before mutation
- unchanged live state/accounting on every rejection path
[...]

Issue #368, opened 2026-08-27
[Roadmap C4] Localized dynamic content/resource replacement


yongkyuns:
## Roadmap

Phase C4 child of #955. Architecture authority: `docs/architecture.md`.

## Goal

Support localized content/resource replacement without conflating persistent authored edits with transient runtime-produced content.

## Contract

Persistent source/content edits use the shared semantic `ReplaceContent` mutation and versioned immutable resources:

```text
persistent source/content edit
  -> validate / compile / prepare
  -> semantic ReplaceContent
  -> local impact + lowering/runtime update
  -> coherent resource/render publication
```

Transient host/updater output may instead publish a versioned **effective runtime content/resource result**. It must not rewrite authored/base content or advance semantic revision every frame merely because the effective result changes.

If product semantics require a transient result to persist, reconcile it into authored/base state through the normal semantic mutation path.

### C4.1 — persistent replacement
Provide one content-type-independent semantic replacement operation. Preserve semantic `NodeId` unless the operation explicitly replaces the node itself.

### C4.2 — effective dynamic content
Allow host/native dynamic producers to install a new effective content/resource version for the current runtime state without turning each update into authored semantic mutation/relowering.

### C4.3 — generation/version safety
Every async/host compile or resource-preparation result is tied to the exact source/content/revision context that produced it. Stale or out-of-order results cannot overwrite newer state.

### C4.4 — fail-coherent publication
Preparation failure leaves the previous coherent authored/effective resource active. Semantic state, execution state, spatial/bounds state and renderer-facing deltas publish coherently rather than in partially updated phases.

### C4.5 — locality
Only affected execution slots/dependencies/spatial entries/resources/renderer ranges update. Transform/style-only changes never regenerate immutable content.

### C4.6 — text integration
Text/math replacement consumes #369 compiled artifacts and the same versioned content path. No text-specific scene, runtime or sidecar.

## Acceptance

- replacing one persistent object resource in a large scene does not rebuild/resend unrelated state;
- transient `always_redraw`-class content can update effective runtime/resource state without semantic revision churn;
- stale async/host results are rejected;
- failed preparation leaves the prior coherent version active;
- unrelated semantic/execution/resource identities remain stable;
- bounds/spatial/resource dependents update correctly and locally;
- transform/style-only edits do not regenerate content;
- #64 hot reload and #70 callbacks reuse this same resource/version machinery.

## Dependencies

[...]

PR #587, opened 2026-08-29
Add retained resource mutation wire transport

yongkyuns:
## Summary
- add a dedicated retained resource mutation transport channel/version
- carry affected `ObjectId`, exact expected text handle, replacement text handle, and one-resource `RetainedResourceBundle`
- reuse the existing resource bundle schema for localized font/vector/text dependencies rather than introducing a second payload model
- add binary encode/decode with protocol validation
- add a session/sequence guard that deterministically drops stale transactions and rejects gaps/new-session nonzero sequences
- prove opaque 64-bit resource IDs and full handle versions survive serialized round-trip exactly

## Architecture
This is the first wire-format tranche of #573, built directly on merged #579. It intentionally does not mutate `InstalledRetainedResources` yet: serialization/version sequencing and install/rollback mechanics remain separate reviewable boundaries.

`RetainedResourceBundle` is reused as the localized dependency payload, so initial install and hot mutation share one immutable resource serialization format. Ordinary retained frame deltas remain resource-payload-free.

The next tranche will apply a validated envelope through the existing `InstalledRetainedResources` owner, staging dependencies before the content-handle switch and leaving prior state untouched on any failure.

Tracks #573 and #368. Architecture parent #361.

### DIFF
--- a/crates/noon-web/src/lib.rs
+++ b/crates/noon-web/src/lib.rs
@@ -22,6 +22,7 @@ mod retained_authoring_scene;
 mod retained_execution_canvas;
 mod retained_execution_resources;
 mod retained_execution_transport;
+mod retained_resource_mutation_transport;
 mod retained_resource_transport;
 mod retained_typst_canvas;
 mod semantic_snapshot;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Superseded by #629, which replays this same transport-only implementation directly onto current master while preserving the evolved `noon-web` module/export list. Fresh exact-head CI will qualify the current-master replay before merge.

Issue #364, opened 2026-08-27
[Roadmap B4] Native Text/MarkupText retained resource backend


yongkyuns:
## Roadmap
Phase B4 child of #954. Architecture authority: `docs/architecture.md`.

## Goal
Implement native `Text`/`MarkupText` shaping/layout as semantic content that compiles into the shared retained text-resource representation. Remove temporary geometry-backed/frontend-owned text behavior for supported cases.

## Architecture
- `Text` is one semantic object unless public semantics explicitly require semantic children;
- shaping/layout/source-cluster identity is shared Rust behavior;
- immutable font/text resources are content resources owned below semantic authoring;
- ordinary glyphs use retained atlas rendering;
- outlines are produced lazily only for path-level semantics such as Create/Write/morph/matching;
- Python is a thin handle facade per #61;
- this feature does not wait for JS/TS D6/#259.

### Text
Common ManimCE v0.21 behavior for source, font/font size, weight/style/slant where supported, color/range styling, multiline spacing/alignment, bounds/layout and stable cluster/source identity.

### MarkupText
Support the required Pango-like markup subset with explicit diagnostics for unsupported constructs and preserved semantic/source part identity.

### Font policy
- deterministic font selection for qualification paths;
- immutable exact font-resource identity;
- deterministic fallback/diagnostics;
- no hidden host-font dependence in canonical parity fixtures.

## Acceptance
- supported `Text`/`MarkupText` examples author through shared semantic handles;
- no Python-owned glyph geometry or rectangle placeholder remains for supported cases;
- Python and Rust observe the same semantic text/bounds/query behavior;
- steady-state text uses retained glyph resources with no Python frame work;
- substring/range styling, multiline layout and font selection have focused tests;
- animation-member/source identity integrates with #705/#715/#737/#741;
- representative cases pass B7 semantic/raster qualification.

## Dependencies
Phase A semantic handles/resource ownership; #369 shared compiler/cache; #362 renderer locality. Real LaTeX remains #365.

yongkyuns:
Implementation status / next order:

- #374 active: native Swash/OpenType compiler foundation normalizing directly into `TextResource` + `FontResourceArena`; single-line LTR is intentionally the first deterministic slice.

After #374 is green/merged, proceed backend-first in this order:
1. multiline line/run layout with stable UTF-8 source offsets, bounds/baseline, alignment and explicit line-spacing semantics;
2. script/direction itemization and bidi handling;
[...]

PR #590, opened 2026-08-29
Expose Rust-owned native Text layout queries

yongkyuns:
## Summary
- derive native `Text` authoring bounds from the canonical Rust `noon::Text` compiler/font policy rather than reimplementing measurement in Python
- cache only the intrinsic scene-space `Rect` on the WASM source handle; shaped glyphs, font bytes, SVG, geometry, and atlas state remain Rust-owned and absent from the authoring wire
- expose center/width/height/critical-point scalar queries that update from ordinary retained transforms without reshaping
- wire Python `Text.width`, `Text.height`, `get_center`, and `get_critical_point` to those Rust-owned queries
- exercise Manim-style `Text.next_to(...)` and width scaling in the browser authoring smoke

## Architecture
The measurement path deliberately authors one temporary native `Text` through the same public Rust path introduced by #582, reads its normalized `TextResource.bounds`, converts to scene units, and drops the temporary resource arenas. Python never measures text or receives glyph/font resources.

#586 is merged. This branch has been replayed as one commit directly on current master because none of its three target files changed across the intervening 29 commits. The browser/Python blobs are byte-identical to the previously qualified implementation. The only Rust adjustment cfg-gates the layout helpers and `Rect` import to `wasm32 || test`, fixing the native-library Clippy dead-code failure without suppressing lints.

## Validation
The browser smoke creates `Square(2)` plus `Text("Native Noon")`, asserts positive native metrics, sets `label.width = 2`, then uses `label.next_to(box, RIGHT, buff=0.25)` and requires the retained text center to land at x=2.25 with zero placeholder text geometry.

Fresh exact-head CI is required before merge.

Related: #364, #361, #582, #586

### DIFF
--- a/crates/noon-web/src/retained_authoring.rs
+++ b/crates/noon-web/src/retained_authoring.rs
@@ -1,5 +1,7 @@
 use std::collections::HashSet;
 
+#[cfg(any(target_arch = "wasm32", test))]
+use noon_core::Rect;
 use noon_core::{Color, ObjectId, Transform2D, Vec2, WHITE};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


yongkyuns:
Scope note from final review: these authoring queries intentionally expose the canonical native compiler's current `TextResource.bounds`, which is a centered conservative line-box bound owned by Rust. This PR does not introduce Python measurement and does not claim exact glyph-ink bounds for arbitrary rotated text. Exact native Text bounds/layout and canonical raster parity remain owned by #364; improving the native backend's bound definition there will automatically flow through this handle surface rather than requiring another frontend implementation.

PR #593, opened 2026-08-29
Gate retained text resource churn

yongkyuns:
## Summary
- add a 50,000-cycle retained text-resource churn regression under `noon-core`
- mix versioned in-place replacement with remove/reinsert generation recycling
- assert the physical arena remains at a one-slot high-water mark throughout
- assert stale version/generation handles never regain access
- assert retained-byte/live-resource accounting returns to zero after removals

## Why
This is an additive #114 resource-lifetime hardening slice following the newly merged generation/version mutation foundation. It uses deterministic structural counters rather than timing thresholds and does not overlap active text authoring, visibility/culling, renderer, or transport PRs.

## Risk
Test-only change. Main risk is runtime cost; the payload is intentionally empty of glyph/vector content so 50k cycles exercise arena bookkeeping/identity without expensive shaping or rendering.

Tracks #114 and validates the resource identity foundations used by #573/#368.

### DIFF
--- a/crates/noon-core/tests/text_resource_churn.rs
+++ b/crates/noon-core/tests/text_resource_churn.rs
@@ -1,0 +1,78 @@
+use std::sync::Arc;
+
+use noon_core::{
+    remove_text_if_current, replace_text_if_current, Rect, TextResource, TextResourceArena,
+    TextSourceKind, Vec2,
+};
+
+const CHURN_CYCLES: usize = 50_000;
+
+fn text(source: impl Into<Arc<str>>) -> TextResource {
+    TextResource {
+        source: source.into(),
+        kind: TextSourceKind::Plain,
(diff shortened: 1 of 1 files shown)


yongkyuns:
Exact-head validation checkpoint for `06b94b5ef92a43b7be7853235284f11134baa1d5`: rustfmt passed, Clippy passed, and the full Rust workspace test suite passed, including the new 50,000-cycle retained text-resource churn regression. Coverage inventory/semantic-ownership validation also passed. Remaining CI work is runner-queued web-package validation plus Rust and Python/JavaScript coverage. Master advanced only via #580 (`noon-web`), which does not overlap this one-file `noon-core` test change; PR remains mergeable. Keeping unmerged until those exact-head tails complete.

PR #604, opened 2026-08-29
Derive engine visibility from evaluated camera

yongkyuns:
## Summary
- add `Camera2DState::viewport_bounds(aspect)` as the shared semantic camera-to-world-bounds rule
- expose `EngineScenePlayer::viewport_visibility_json(aspect)` and `viewportVisibilityJson` in WASM
- derive the query from the authoritative post-evaluation camera and retained `ScenePlayer` spatial index
- reject non-positive/non-finite render aspects explicitly
- add regressions for aspect-sensitive visibility and an animated camera moving an object into view

## Architecture
The execution engine remains the only spatial-query owner. The render worker supplies only render aspect; it does not reconstruct semantic camera bounds or maintain a second player/index. Camera world width uses the same contract as the renderer: `height * aspect`.

This is stacked on #601. The next slice can carry render aspect to the engine worker, return a same-frame `noon.execution.visibility` packet, then resolve it through #601 and submit candidate-only batches through #591.

Tracks #569 and #66.

### DIFF
--- a/crates/noon-core/src/camera.rs
+++ b/crates/noon-core/src/camera.rs
@@ -1,6 +1,30 @@
-use crate::{Camera2DState, GeometryRef, Transform2D};
+use crate::{Camera2DState, GeometryRef, Rect, Transform2D, Vec2};
 
 impl Camera2DState {
+    /// Derive axis-aligned world bounds for a render target aspect ratio.
+    ///
+    /// Camera height is the semantic vertical span. The horizontal span follows
+    /// the render target aspect exactly, matching the renderer's camera contract.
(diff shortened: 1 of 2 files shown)


yongkyuns:
Do not merge current head `2d436ffd…` yet: downstream Rust CI already exposed a deterministic rustfmt failure in the WASM wrapper. `viewport_visibility_json` must be formatted as:

```rust
self.inner
    .viewport_visibility_json(aspect)
    .map_err(js_error)
```

No semantic change is needed; this is the only known #604-specific blocker. Please keep the one-commit normalized diff and rerun exact-head CI after the formatter-only correction.

PR #605, opened 2026-08-29
Consume retained viewport visibility in web renderer

yongkyuns:
## Summary
- pair legacy execution deltas with authoritative retained `noon.execution.visibility` packets after visibility activation
- carry render aspect from the render worker to the engine so viewport queries use the post-evaluation semantic camera
- resolve visibility through `ExecutionFrameMirror` and submit only the ordered rows through #591 while keeping packed GPU caches resident
- gate presentation after activation until matching visibility arrives, including control/host-callback deltas and no-delta resize ticks
- bind visibility delivery to the existing execution `{session, sequence}` identity, publish it only after the paired delta is acknowledged, and reject visibility whose pair does not match the last execution envelope actually accepted by the renderer
- serialize visibility-bearing execution frames to one in flight so shared and transferable transports have the same ordering contract
- reuse the last mirror-compatible visibility on unchanged ticks; first activation or an aspect change with no object delta synchronizes one snapshot before computing new visibility
- defer surface resize while an old-aspect visibility frame is in flight so candidate bounds and GPU projection cannot mix
- expose candidate/result/live-set renderer metrics and add focused worker protocol regressions

## Architecture
The execution engine remains the only spatial-index owner. The render worker supplies only aspect and consumes generation-safe candidate identity; it does not scan bounds or build another index.

Correctness does not rely on MessagePort/FIFO timing across the execution transport and visibility messages. The engine admits one visibility-bearing frame at a time, derives the existing execution session/sequence from that delta, waits for the render transport acknowledgement, then publishes visibility for the now-current mirror frame. The render worker independently tracks metadata from the execution envelope it actually applied and requires an exact pair match before accepting visibility. Reconnects/transitions clear that pairing state. This is frame pairing, not protocol-version negotiation; Rust and JS still share one current application schema.

The initial bootstrap frame remains uncullled so renderer construction does not depend on a second bootstrap payload. After activation, legacy presentation is visibility-gated and never falls back to ordinary preparation for a pending execution frame.

Stacked on #604. Tracks #569 and #66.

### DIFF
--- a/crates/noon-web/src/execution_canvas.rs
+++ b/crates/noon-web/src/execution_canvas.rs
@@ -14,14 +14,14 @@ mod wasm {
     use std::mem;
 
     use noon_core::Vec2;
-    use noon_render_wgpu::{Camera2D, FramePreparer, GpuRenderer};
+    use noon_render_wgpu::{Camera2D, GpuRenderer, VisibilityFramePreparer};
     use noon_runtime::FrameChanges;
     use wasm_bindgen::prelude::*;
     use web_sys::OffscreenCanvas;
 
     use crate::{
         gpu_diagnostics::{install_wgpu_error_handler, GpuDiagnosticMailbox},
-        ExecutionFrameMirror, TransportApplyOutcome,
+        ExecutionFrameMirror, ExecutionVisibilityEnvelope, TransportApplyOutcome,
     };
 
     use super::{MANIM_DEFAULT_CAMERA_HEIGHT, MANIM_DEFAULT_CLEAR_COLOR};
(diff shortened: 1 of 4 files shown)


yongkyuns:
CI diagnosis on current consumer design: the two browser failures share one transition scheduling bug. `authoring-render-worker.js::frame()` continues emitting ticks after `beginRendererTransition()` attaches the replacement engine and while `transitionMode != null` / replacement `bootstrapPromise` is active. On retained -> legacy this sends a timestamp-only retained-shaped tick to the new legacy engine (`invalid viewport aspect`); after the replacement snapshot commits, the same loop can request legacy visibility while `ExecutionCanvasRenderer.create()` is still awaiting (`visibility before renderer bootstrap`).

Fix at the render-owner scheduling boundary, not by relaxing engine validation or staging stale visibility: continue draining transition transport so the replacement snapshot can arrive, but suppress new tick emission while `transitionMode !== null || bootstrapPromise !== null || renderer === null`. The existing RAF loop should keep scheduling itself; once bootstrap commits, ticks resume using the now-authoritative `mode`. Add a protocol/source regression for this transition gate. Also apply the two rustfmt diffs already reported in `execution_canvas.rs` and #604's `execution_transport.rs` before merge.

PR #612, opened 2026-08-29
Add dirty-range text GPU upload primitive

yongkyuns:
## Summary
- add `TextGlyphGpuRenderer::upload_ranges` for caller-provided mask/color instance ranges
- preserve the existing full upload API unchanged
- automatically fall back to a full plane write whenever instance-buffer reallocation occurs
- report actual bytes written through the existing `TextGpuUploadStats`
- add noop-device regressions for partial range writes and reallocation fallback

## Architecture
This is the GPU primitive for the next #362 locality slice. It deliberately does not yet change parent retained-generation plumbing. The caller must only use partial ranges when the GPU state is known to contain the immediately preceding prepared generation; missed/new generations will remain full-upload fallbacks in the parent integration.

Tracks #362 and #68.

### DIFF
--- a/crates/noon-text-render-wgpu/src/gpu.rs
+++ b/crates/noon-text-render-wgpu/src/gpu.rs
@@ -1,4 +1,4 @@
-use std::mem::size_of;
+use std::{mem::size_of, ops::Range};
 
 use bytemuck::{Pod, Zeroable};
 use noon_core::Vec2;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Superseded by #621, which replays this exact two-file retained-text upload primitive onto current master after #619 and requires fresh exact-head CI before merge.

Issue #77, opened 2026-08-24
[Roadmap B2] Lines, arrows, tips, and dashed objects


yongkyuns:
## Roadmap
Phase B2 child under #954. Architecture authority: `docs/architecture.md`.

## Goal
Implement the common ManimCE v0.21 line/arrow/dash vocabulary as shared semantic geometry with retained dependent resources reusable by diagrams, plotting and Graph edges.

## Scope
- DashedLine;
- Arrow, Vector, DoubleArrow;
- Elbow, TangentLine and common simple line-derived helpers;
- arrow tips, placement/scaling, endpoint shortening and buff semantics;
- DashedVMobject where compatible with generic path semantics;
- start/end/angle/length/query helpers.

## Architecture constraints
- tips/dashes are semantic/resource dependencies, not frontend reconstruction;
- moving/resizing a line updates dependent tip/dash state locally;
- style inheritance and painter ordering come from shared B1 semantics;
- ordinary line/arrow content participates in Create/GrowArrow/Transform through B3;
- Graph/DiGraph directed edges reuse this representation;
- no graph-specific or renderer-only arrow primitive.

## Acceptance
- representative source-equivalent Line/DashedLine/Arrow/Vector/DoubleArrow cases work with common constructor options;
- endpoint/buff/tip/dash geometry passes focused semantic/raster qualification;
- repeated-arrow scenes retain/batch immutable content efficiently;
- transform/style changes do not rebuild unrelated resources;
- WebGPU/WebGL produce equivalent supported output through B7 gates.

## Dependencies
A1/B1/#53/#62 and B2 path semantics where needed. B3 consumes this for GrowArrow/Create; B5 Graph/plotting reuse it.

yongkyuns:
### Acceptance criteria amendment — exact Manim output parity

This breadth issue must use the exact-output gate from #185, not only API/semantic/example coverage. For **every slice promoted to supported**:

- add a source-equivalent ManimCE **v0.21.0** fixture to the canonical #176 raster + timeline corpus (no Noon-only geometry/style/timing tuning; only the minimum import/browser wrapper adaptation);
- compare semantic state **and rendered pixels** against real Manim at the canonical profile, with explicit/ratcheted raster tolerances and categorized diff artifacts;
[...]

PR #628, opened 2026-08-30
Expose shared Manim Elbow through WASM

yongkyuns:
## Summary
- expose the newly-landed shared Rust `Elbow` semantics through a focused `noon-web` WASM bridge
- serialize the ordinary retained `ObjectSnapshot`; frontend callers do not reconstruct the three-corner path or rotation
- validate JS numeric inputs at the bridge boundary and preserve Manim-compatible zero/negative widths
- add focused native tests for exact retained path/style defaults, edge-case widths, and invalid numeric inputs

## Architecture
This follows the existing sector/geometry bridge pattern. `noon::Elbow` remains the sole owner of observable geometry and Manim defaults; `noon-web` only validates f64→f32 compatibility, serializes the snapshot, and exposes `manimElbowSnapshotJson`. No renderer primitive, worker protocol, Python path math, or persistent Elbow-specific state is introduced.

## Dependency
#630 identified and fixes a real shared-Rust constructor representation bug: Manim rotates Elbow points during construction, while the previous Noon implementation stored the constructor angle in `ObjectSnapshot.transform.rotation`, producing incorrect canonical bounds. This bridge must not merge before #630 lands. After that merge, this branch will be refreshed and its old internal transform-rotation assertion replaced with public rotated path/bounds observations.

## Validation
Fresh exact-head CI is required after the #630 refresh. The normal Rust/web-package jobs will exercise native tests plus wasm-bindgen export generation.

## Remaining boundary
This establishes the browser semantic bridge only. Thin Python/JS compatibility adapters and exact Manim raster/demo qualification can follow independently without adding geometry semantics.

Tracks #77 and #400.

### DIFF
--- a/crates/noon-web/src/lib.rs
+++ b/crates/noon-web/src/lib.rs
@@ -14,6 +14,7 @@ mod host_player;
 #[path = "legacy.rs"]
 mod legacy;
 mod lifecycle;
+mod manim_elbow_bridge;
 mod manim_geometry_bridge;
 mod manim_path_query_bridge;
 mod manim_sector_bridge;
@@ -44,6 +45,7 @@ pub use execution_visibility::*;
 pub use host_player::*;
 pub use legacy::*;
 pub use lifecycle::*;
+pub use manim_elbow_bridge::*;
 pub use manim_geometry_bridge::*;
 pub use manim_path_query_bridge::*;
 pub use manim_sector_bridge::*;
--- a/crates/noon-web/src/manim_elbow_bridge.rs
+++ b/crates/noon-web/src/manim_elbow_bridge.rs
@@ -1,0 +1,112 @@
+use noon::{Elbow, IntoSnapshot};
+use noon_core::ObjectSnapshot;
+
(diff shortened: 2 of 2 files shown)


yongkyuns:
Blocking merge on #630. The pinned ManimCE semantic oracle exposed a real shared-Rust `Elbow` constructor-bounds mismatch for rotated/wide and negative-width cases. #630 now fixes constructor-time rotation ownership in `noon::Elbow` and is requalifying at unchanged tolerance. This WASM bridge should be refreshed onto that corrected shared primitive after #630 lands rather than expose the known-wrong representation.

yongkyuns:
Refreshed after #630 merged. Current head `04c70806ddbf0456f490e16bb8ff901f749c5f51` is one commit directly on corrected master `6ef59ca229555bb8bd50a4be96b2936686ec3cd4`. The stale constructor-storage regression was replaced with public serialized center/width/height checks for the rotated Elbow, while also asserting no residual constructor rotation remains in transform state. Current exact-head Rust CI has passed rustfmt and Clippy; workspace tests and the remaining qualification matrix are running.

PR #630, opened 2026-08-30
Gate shared Elbow semantics against ManimCE

yongkyuns:
## Summary
- add a renderer-independent semantic oracle for shared Rust `Elbow`
- compare Noon against pinned ManimCE v0.21.0 for default, rotated/wide, zero-width, and negative-width cases
- qualify public start/end/center/width/height observations at the existing `2e-5` tolerance
- fix the real mismatch exposed by that oracle: bake Manim's constructor-time rotation into the retained path points so canonical snapshot bounds observe the rotated VMobject points rather than a rotated local AABB
- keep subsequent `.rotate()` authoring operations on the ordinary retained `ObjectSnapshot` transform

## Architecture
Manim constructs `Elbow` by creating the three-corner path, scaling it about the origin, then rotating the VMobject's points about the origin. The previous Noon implementation represented that constructor angle as `ObjectSnapshot.transform.rotation`; canonical bounds therefore rotated the path's local AABB and were conservatively wrong for rotated and negative-width Elbows.

The corrected shared Rust constructor performs only the constructor-time point rotation in geometry. It adds no renderer special case, frontend geometry, alternate bounds implementation, or persistent Elbow state. Later transforms continue through the shared retained transform path.

## Regression
Focused Rust coverage pins the rotated/wide and negative-width public center/width/height values that failed against Manim and verifies later rotation still remains a retained transform. The pinned-Manim differential covers default, rotated/wide, zero-width, and negative-width start/end/center/width/height observations.

Fresh exact-head CI is required before merge. No tolerance changes are included.

Tracks #77 and #185.

### DIFF
--- a/.github/workflows/manim-differential.yml
+++ b/.github/workflows/manim-differential.yml
@@ -43,4 +43,5 @@ jobs:
           python scripts/manim-differential.py
           python scripts/manim-dot-ellipse-differential.py
           python scripts/manim-sector-differential.py
+          python scripts/manim-elbow-differential.py
           python scripts/manim-frame-sampling-differential.py
--- a/crates/noon-web/examples/manim_elbow_oracle.rs
+++ b/crates/noon-web/examples/manim_elbow_oracle.rs
@@ -52,12 +52,7 @@ fn observation(snapshot: ObjectSnapshot) -> Value {
     })
 }
 
-fn insert(
-    observations: &mut Map<String, Value>,
-    name: &str,
-    width: f32,
-    angle: f32,
-) {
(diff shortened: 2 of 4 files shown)


yongkyuns:
The first pinned-Manim run exposed a broader shared-core bounds defect rather than an Elbow path/orientation defect. Elbow start/end semantics match; `ObjectSnapshot::world_bounds()` currently rotates the local AABB for `VectorPath`, which inflates rotated open-path bounds and shifts `center()`. I opened #631 with the shared-core fix/acceptance criteria. Keeping this differential unchanged as the executable regression—no tolerance change or Elbow-specific workaround.

PR #637, opened 2026-08-30
Use analytic world bounds for transformed circles

### LINKED ISSUE #636: Use analytic world bounds for transformed Circle geometry
## Problem

#631 tightened transformed `Line` and `VectorPath` world bounds but intentionally left `Circle` separate. `GeometryRef::world_bounds()` still transforms the four corners of a circle's local square AABB. Under nonuniform scale plus rotation, the mathematical shape is an oriented ellipse, so a tighter analytic AABB is possible.

## Validation result

PR #637 implemented that analytic bound, but full workspace CI exposed an important semantic mismatch: `Ellipse` is represented as an affinely stretched analytic `Circle`, while Noon's public Manim-compatible bounds intentionally follow Manim VMobject transformed control-point hulls. The existing `analytic_geometry_authoring::tests::rotated_nonuniform_ellipse_uses_manim_control_point_hull` regression failed because the analytic ellipse AABB is tighter than Manim's public `width`/`height` semantics.

Changing `GeometryRef::world_bounds()` therefore conflates two contracts:
- semantic/layout bounds, which need Manim parity;
- tighter renderer/spatial bounds, which may use mathematical geometry when safe.

An `Ellipse`-specific exception would make the shared geometry contract representation-dependent and is not an acceptable workaround.

## Resolution

Do not change `Circle` semantic world bounds this way. If tighter culling/locality bounds are needed, introduce a separately named renderer/spatial bound contract with explicit conservative-safety requirements and keep public semantic/layout bounds unchanged.

Closed as not planned after #637 CI demonstrated the incompatibility. Related: #631, #569.

yongkyuns:
## Summary
- compute exact transformed `Circle` AABBs analytically from the scale+rotation linear transform
- keep `Rectangle`, `Line`, `VectorPath`, and `External` behavior unchanged
- add focused regressions for rotated nonuniform scale and reflected/negative scale

## Architecture
`noon-core` remains the sole owner of semantic bounds. The calculation is O(1), allocation-free, and feeds existing layout/spatial consumers without renderer or frontend reconstruction.

## Validation
- added focused `noon-core` tests for the cases that previously overbounded transformed circles
- branch is one commit / one core file with no unrelated changes
- fresh GitHub Actions status will be used as the integration gate

Closes #636.

### DIFF
--- a/crates/noon-core/src/lib.rs
+++ b/crates/noon-core/src/lib.rs
@@ -646,7 +646,19 @@ impl GeometryRef {
                 transform.transform_point(*end),
             ]),
             Self::VectorPath(path) => path.transformed_conservative_bounds(transform),
-            Self::Circle { .. } | Self::Rectangle { .. } => {
+            Self::Circle { radius } => {
+                let (sin, cos) = transform.rotation.sin_cos();
+                let radius = radius.abs();
+                let half_extents = Vec2::new(
+                    radius * (transform.scale.x * cos).hypot(transform.scale.y * sin),
+                    radius * (transform.scale.x * sin).hypot(transform.scale.y * cos),
+                );
+                Some(Rect::new(
+                    transform.translation - half_extents,
+                    transform.translation + half_extents,
+                ))
+            }
+            Self::Rectangle { .. } => {
                 let bounds = self.local_bounds()?;
                 let corners = [
                     Vec2::new(bounds.min.x, bounds.min.y),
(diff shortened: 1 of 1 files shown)


yongkyuns:
CI exposed a semantic conflict that makes this change unsafe to merge as written. `cargo test --workspace --all-features` fails `analytic_geometry_authoring::tests::rotated_nonuniform_ellipse_uses_manim_control_point_hull`: Noon intentionally matches Manim VMobject bounds from transformed control points, while this PR switches affinely stretched Circle/Ellipse snapshots to the tighter mathematical ellipse AABB. That changes public `width`/`height` and layout semantics even though renderer/culling bounds would be tighter.

I am closing this PR rather than adding an Ellipse-specific exception. If tighter spatial bounds are still desirable, they should be introduced as a separate renderer/spatial bound contract rather than changing Manim-compatible semantic bounds.

Issue #642, opened 2026-08-30
[Roadmap C6] Measure and reduce post-consolidation playground startup


yongkyuns:
## Roadmap
Phase C6 child of #955. Architecture authority: `docs/architecture.md`.

## Goal
Measure and reduce playground startup and edit-to-frame latency around the consolidated architecture. The interactive playground should preload the target authoring/execution session after the initial UI/source paint so the first edit does not pay the full Pyodide/runtime cold-start cost.

Do not optimize, preserve, or split around the pre-reset legacy-vs-retained worker topology.

## Rule
Preloading is an integration/UX choice, not a new architecture boundary. Reuse the one Python facade, Semantic Scene, execution session and renderer topology; do not create a second semantic implementation, mirror, scheduler or disposable execution mode merely to preload.

Browser UX work that only drives existing target handoffs may proceed during Phase A. Remeasure the complete startup topology after the relevant Phase A frontend/IR/browser ownership changes land before making package/worker specialization decisions.

## Measurement cases
Record cold and warm phase timings for:
- page ready -> preload start;
- authoring/Pyodide ready;
- Semantic Scene ready;
- Execution Plan/runtime ready;
- renderer/device ready;
- first presented frame;
- edit -> Python execution -> reconciliation -> presented frame;
- warm full-source rerun;
- worker count, WASM instance count, transferred/instantiated bytes and peak memory.

Test representative geometry and text/resource scenes plus a constrained/mobile-class configuration.

## Optimization cases
Only optimize topology/package placement from measurements.

### C6.1 Prewarm the interactive session
After the initial source/gallery UI has painted, start the target Python authoring and execution/render session without waiting for an explicit Run. Keep the resulting Pyodide worker and execution owner resident across edits.

The preload must not block initial source visibility or introduce a parallel scene/runtime authority. Edits arriving while preload or another Run is active should coalesce to the latest source and execute through the same normal authoring/reconciliation path.

### C6.2 Overlap independent work
Authoring/source execution may overlap runtime/renderer initialization where dependencies permit. Do not serialize unrelated startup phases unnecessarily.

### C6.3 Remove duplicate initialization
If the consolidated topology still instantiates the same heavy WASM/dependencies multiple times without architectural need, remove/split that duplication based on measured bytes/time/memory.

### C6.4 Package/worker specialization
Split WASM surfaces or worker responsibilities only if measurements show meaningful startup/memory benefit. Shared semantic behavior remains Rust-owned and frontends do not fork implementation.

[...]

PR #644, opened 2026-08-30
Overlap Python authoring with cold runtime startup

yongkyuns:
## Superseded by #646 + #645

This PR proved that Python authoring can safely overlap execution startup, but its concrete cold path overlaps authoring with the existing speculative empty **legacy** engine bootstrap. #646 removes that wrong-engine bootstrap entirely, which is the stronger architectural step and avoids retained first-run legacy -> retained churn.

The useful overlap objective remains part of #642 and will be recovered after #646 using #645's mode-free persistent render-owner preparation, so we can overlap Python/Pyodide work with only mode-independent GPU/WASM preparation rather than booting an engine before the authored mode is known.

The deterministic playback-control smoke fix discovered here remains available on the branch if the direct-mode path exposes the same timing edge.

Original implementation details:
- start selected Python authoring while deferred execution startup proceeds
- join both siblings with `Promise.allSettled`
- preserve stale-generation checks and reconciliation ordering
- deterministic playback busy assertion via the existing `beforeReconcile` test hook

Closed without merge in favor of the cleaner #646 -> #645 -> mode-free overlap sequence. Tracks #642.

### DIFF
--- a/scripts/playground-playback-controls-smoke.mjs
+++ b/scripts/playground-playback-controls-smoke.mjs
@@ -279,13 +279,34 @@ try {
       .find((id) => id && id !== selected);
   });
   assert.ok(targetExampleId, "gallery must expose another example");
+  await page.evaluate(() => {
+    let releaseSwitch;
+    const switchGate = new Promise((resolve) => {
+      releaseSwitch = resolve;
+    });
+    globalThis.__NOON_PLAYGROUND_TEST_HOOKS__ ??= {};
+    globalThis.__NOON_PLAYGROUND_TEST_HOOKS__.beforeReconcile = async () => {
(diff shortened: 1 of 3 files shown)


yongkyuns:
Sequencing note after reviewing #646: this PR overlaps authoring with the current empty legacy-engine bootstrap, while #646 removes that bootstrap and waits for the authored result to select the correct engine mode. Those two `runScene()` strategies should not merge independently as-is. Prefer #646 as the architectural base; after it lands, rework this lane around mode-independent overlap (for example Python authoring in parallel with persistent render-owner / other mode-agnostic preparation) rather than speculatively starting a legacy engine that retained first-runs must discard. Holding this PR unmerged avoids optimizing a startup path #646 intentionally removes.

Issue #116, opened 2026-08-24
Testing hardening: comprehensive gap analysis and parallel workstream tracker


yongkyuns:
## Objective
Raise Noon from a collection of good focused tests and browser smokes to a layered verification system suitable for a realtime, interactive, Manim-compatible browser engine.

This tracker is based on an audit of the current Rust/Python/JS tests, browser smoke harnesses, Manim differential suite, CI workflows, renderer tests, and performance harnesses on `master`.

## What is already strong
Do **not** replace these; build around them.

- Rust workspace tests cover substantial transform/lifecycle/patch/geometry/cache behavior.
- `noon-geometry` has focused tessellation/morph/stroke correctness tests.
- `noon-runtime` has broad transform/lifecycle/live-patch integration fixtures.
- the web build runs Node unit tests and the Python unittest suite.
- browser CI already splits authoring/reactive/rendering and exercises both WebGPU and WebGL fallback paths.
- browser smoke renders every playground picker scene at multiple deterministic times and has targeted Create/path continuity assertions.
- #57 provides a pinned ManimCE v0.21.0 differential semantic harness.
- #73 tracks Manim public API/example coverage; #91 tracks executable tutorial/example coverage.
- `docs/performance.md` has useful named-machine native/render/transport/GPU baselines.
- #68 defines explicit architecture complexity invariants and required large-scene workloads.

### 1. Coverage is broad but not measurable as one system
There is no cross-language test inventory/coverage report or ratchet, and `docs/ci.md` no longer describes the current CI topology exactly.

→ #104

### 2. Stateful invariants are mostly example-driven
Semantic handles, patches, tracks, family graphs, composition, and reactive graphs have combinatorial mutation state spaces that hand-authored examples cannot exhaust.

→ #105

### 3. Interactive determinism lacks a broad oracle
Direct seek, incremental playback, backward scrub, patch/reconcile, fresh reload, native runtime, and WASM runtime need explicit equivalence contracts.

→ #106

### 4. No repository-level fuzzing/adversarial corpus
Transport, geometry/tessellation/morphing, and mutation sequences need malformed/extreme input coverage and minimized regression seeds.

→ #107

### 5. Cross-language schemas are not independently compatibility-gated
Scene/patch/callback/worker JSON is central to authoring but lacks one versioned old/new fixture corpus and round-trip contract.

→ #108

### 6. Browser screenshots are not general visual regressions
[...]

Issue #104, opened 2026-08-24
[Test T1] Add unified test inventory, coverage reporting, and coverage ratchets


yongkyuns:
## Goal
Make Noon's test surface measurable across Rust, Python, JavaScript, WASM/browser, and compatibility layers without turning a single line-coverage percentage into the quality target.

## Current gap
The repo has substantial tests, but coverage is distributed across inline Rust tests, integration tests, Python unittest files, Node tests, Playwright smoke scripts, Manim differential probes, and performance harnesses. There is no single machine-readable inventory showing which production modules/behaviors are exercised, which tests run in which CI job, or where coverage is absent.

`docs/ci.md` also describes an older fast/full-gate split that no longer matches the current workflow, so test-policy documentation can drift from what is actually required.

## Scope
- Add a generated testing inventory grouped by layer/crate and test type:
  - `noon-core`, `noon-ir`, `noon-compile`, `noon-runtime`;
  - `noon-geometry`, `noon-render-wgpu`, `noon-web`, public `noon` facade;
  - Python compatibility/authoring modules;
  - browser JavaScript/worker/editor modules;
  - Manim differential and tutorial/example corpora.
- Produce coverage artifacts in CI:
  - Rust via `cargo llvm-cov` or equivalent;
  - Python via `coverage.py`;
  - Node/JS via V8/c8 coverage where practical.
- Exclude generated wasm-bindgen output and other mechanically generated files from coverage goals.
- Publish per-layer summaries as artifacts and a concise PR summary.
- Establish a baseline and ratchet policy. Prefer changed-code coverage and/or per-layer non-regression over an arbitrary global threshold.
- Reconcile `docs/ci.md` with the actual workflow and document which tests are blocking vs diagnostic/manual.

## Acceptance
- A CI artifact shows coverage and test ownership for every production layer in the workspace.
- New production modules cannot silently appear outside the inventory.
- Coverage regressions are visible; after baseline stabilization, changed-code/per-layer ratchets can fail CI.
- The report distinguishes unit/integration/property/fuzz/browser/differential/performance coverage rather than presenting only a percentage.
- CI documentation matches the checked-in workflows.

## Non-goals
- Do not require 100% line coverage.
- Do not duplicate Manim API-surface accounting in #73 or semantic comparison in #57.

## Parallelism / dependencies
**Wave T0; independent and can start immediately.** This is shared test infrastructure and should not block the other test lanes from beginning.

PR #647, opened 2026-08-30
Classify CI workflow topology

yongkyuns:
## Summary
- add a self-registering Node regression that inventories every `.github/workflows/*.yml` file
- classify workflows into explicit architecture/testing families rather than letting new standalone lanes appear silently
- require the core main, coverage, platform/release, Manim, and playground families to remain represented
- keep `docs/ci.md` anchored to the core workflow documents it claims to describe

## Why
#116/#104 explicitly call out CI-topology/documentation drift. The repository has grown several dedicated playground/recovery/browser lanes beyond the older prose description. This slice adds a cheap mechanical ownership gate: adding a workflow with a new naming family now requires an explicit classification decision instead of silently expanding the CI surface.

The test is automatically syntax-checked and executed by `scripts/build-web-demo.sh` through the existing `web/*.test.mjs` discovery, so there is no duplicate test inventory to maintain.

## Scope / risk
Test/governance only; no production/runtime behavior changes. Prefix classification is deliberately coarse and architectural (Manim, playground, renderer, authoring, performance) rather than enumerating every workflow, so adding another workflow inside an existing family remains cheap.

Tracks #104 and #116.

### DIFF
--- a/web/ci-workflow-topology.test.mjs
+++ b/web/ci-workflow-topology.test.mjs
@@ -1,0 +1,69 @@
+import assert from "node:assert/strict";
+import { readdir, readFile } from "node:fs/promises";
+
+const workflowDir = new URL("../.github/workflows/", import.meta.url);
+const workflowFiles = (await readdir(workflowDir))
+  .filter((name) => name.endsWith(".yml") || name.endsWith(".yaml"))
+  .sort();
+
+assert.ok(workflowFiles.length > 0, "CI workflow inventory must not be empty");
+
+const exactFamilies = new Map([
+  ["ci.yml", "main"],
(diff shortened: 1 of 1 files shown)


yongkyuns:
Fresh exact-head validation on the first commit exposed one existing workflow family the classifier had missed: `retained-typst-authoring.yml`. That caused every lane executing the web package test to fail at the same new governance assertion, while Rust and the Manim semantic/API lanes remained green. Commit `f9b55f7d` classifies the existing `retained-` family explicitly rather than weakening the gate. A new full matrix has started on that corrected head; the web-package job is currently executing the build/test step.

PR #650, opened 2026-08-30
Attach authored execution after render-owner preparation

yongkyuns:
## Summary

Build the production-client handoff on top of #645's now-merged mode-free render-owner preparation protocol.

This tranche teaches `ExecutionWorkerClient` to prepare the permanent render/WASM owner before an authored engine mode is known, then attach exactly one legacy or retained engine once authoring has selected the mode. It deliberately does **not** change playground orchestration yet; that remains the next #642 slice after this client boundary is qualified.

## What changed

- add `ExecutionWorkerClient.prepare()` to transfer the canvas and initialize only the persistent render owner;
- keep preparation mode-free: no legacy or retained engine is created and no renderer/backend is selected;
- allow existing `start()` / `startRetained()` to finalize an already-prepared owner;
- validate that prepared transport/shared-slot configuration matches final engine startup;
- if authoring finishes before render preparation, engine creation waits for the prepare barrier rather than speculating a mode;
- use #645 `start_engine` for the initial authored engine and preserve `ready()` semantics through the render worker's `engine_started` acknowledgement;
- preserve the existing one-shot startup path unchanged for current callers;
- restore a fresh DOM canvas if a prepared render owner is abandoned before any engine starts, so failed/stale authoring cannot poison the next Run;
- preserve canvas recovery when the render-worker constructor fails after OffscreenCanvas transfer, while avoiding double replacement if preparation is concurrently terminated;
- preserve transferred-canvas rollback/retry semantics when prepared engine attachment fails.

## Tests

- race regression: call retained startup before the render `prepared` acknowledgement and prove no engine exists until preparation completes;
- assert preparation creates neither legacy nor retained engine and carries no mode;
- assert the first attached engine is retained and no legacy engine is ever created;
- assert `ready()` resolves from `engine_started` only after the render handoff;
- failure regression: crash the first prepared retained engine, require both workers to be torn down, replace the transferred canvas, and successfully retry with the next session generation;
- constructor regression: fail render-worker construction after canvas transfer, require canvas restoration, then successfully prepare again;
- abandonment regression: prepare successfully, terminate before engine selection, require the transferred canvas to be replaced, then successfully start retained execution on the same client without consuming a phantom session.

## Validation

#645 merged fully green as `304e67b25af934dfd799f63c1f94a3f070b9f527`. Master later advanced only through disjoint #649 manifest/reference-lock files. The exact two-file #650 tree was replayed byte-for-byte onto current master `c8e69581df882a74c18469983bf83b562f2f0412` as one commit `011220a2edf842e06356f745cde7cfe36d278d9e`. Only fresh exact-head validation for `011220a2…` is valid for merge.

Tracks #642.

### DIFF
--- a/web/execution-worker-client.js
+++ b/web/execution-worker-client.js
@@ -23,6 +23,7 @@ export class ExecutionWorkerClient {
   #candidateEngineWorker = null;
   #candidateEngineReject = null;
   #renderWorker = null;
+  #renderPrepared = null;
   #nextRequestIds = { engine: 0, render: 0 };
   #pending = new Map();
   #session = 0;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Blocking correctness issue on exact head `73b5c48bc622c35cdadd66a9b6f3d24758ff428c`: `#startMode()` checks `#engineWorker` before `await this.#renderPrepared`. Two `start()`/`startRetained()` calls issued while `prepare()` is pending can both pass the check, await the same preparation promise, then each enter `#startPreparedMode()` and create an engine. The second render `start_engine` may fail, but a second engine has already been created/assigned by then.

Fix by reserving prepared startup synchronously before the prepare await (e.g. private `#preparedStartPending`, checked alongside `#engineWorker`, set before `await #renderPrepared`, cleared in `finally`). Add a regression that starts retained execution while preparation is pending, immediately attempts a competing legacy `start()`, requires rejection, and verifies worker count stays render-only until `prepared` is acknowledged.

I reproduced the corrected behavior locally against this exact PR source: both focused prepared-start tests pass with that reservation. Do not merge this head; fresh exact-head CI is required after the fix.

yongkyuns:
Rechecked current head `758a83752a81784c174b83a6e60acccda350e378`: the previously reported prepared-start concurrency race is still present. The new canvas abandonment/constructor recovery does not reserve startup ownership. Two `start()` / `startRetained()` calls can still both observe `#engineWorker === null`, both enter the `preparedRender` branch, await the same `#renderPrepared`, then both call `#startPreparedMode(...)` and create competing engines.

Please keep this head unmerged even if its CI is green. The fix should reserve the prepared start *before* awaiting `#renderPrepared`, reject a second start while that reservation is live, and clear the reservation with token/identity semantics so an old cancelled `finally` cannot clear a newer post-terminate start. Add a regression that starts retained twice before acknowledging `prepared`, requires the second call to reject immediately, and proves no engine has been created until the first preparation completes.

PR #651, opened 2026-08-30
Gate retained spatial refits during native animation

yongkyuns:
## Summary
- add a focused `noon-runtime` integration regression for a native Position track crossing the pointer query
- prove the retained spatial index performs zero full rebuilds as the animated object moves into and back out of the query point
- require exactly one leaf upsert per animated evaluation step while 255 unrelated sparse objects remain untouched
- keep hit-test candidate work bounded and require zero full-scan fallbacks

## Why
#641 covered reconciliation-time transform refits and #648 covers structural slot reuse. #569/#66 also require animated objects to enter and leave retained query results without rebuilding the spatial index. This closes that distinct runtime acceptance gap using the public `SlottedSceneInstance::last_spatial_update_stats()` counters.

The test uses a native `Position` track from x=+8 to x=-8 over 2 seconds. At t=0 the object misses the origin, at t=1 it hits, and at t=2 it misses again. Each evaluation must update only that object's retained leaf.

## Scope
Test-only; no renderer, browser transport, or frontend spatial state is introduced. The actual WASM/browser selection/hover consumer under #569 remains separate.

Tracks #569 and #66.

### DIFF
--- a/crates/noon-runtime/tests/spatial_animation_refit.rs
+++ b/crates/noon-runtime/tests/spatial_animation_refit.rs
@@ -1,0 +1,72 @@
+use noon_compile::CompiledScene;
+use noon_core::{
+    Easing, GeometryRef, Property, SceneDefinition, TrackTiming, TrackValues, Transform2D, Vec2,
+};
+use noon_runtime::SlottedSceneInstance;
+
+const STATIC_OBJECTS: usize = 255;
+const MAX_LOCAL_CANDIDATES: usize = 64;
+
+#[test]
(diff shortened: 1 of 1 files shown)


yongkyuns (commented):
Closing this slice rather than weakening it: current master already contains `execution_slots::tests::forward_timeline_motion_refits_only_active_object_leaf`, which covers the same native timeline → local spatial refit invariant. The new integration test also incorrectly expected the constructor's intentional initial index build to report zero full rebuilds. Keeping both would duplicate coverage without adding a new contract.

Issue #78, opened 2026-08-24
[Roadmap B2] VMobject paths, partial paths, smoothing, and boolean geometry


yongkyuns:
## Roadmap

Phase B2 path/VMobject parent under #954. Architecture authority: `docs/architecture.md`.

## Goal

Provide the path-level API needed for custom VMobjects, plotting, Create/Write, morphing and boolean geometry without exposing Manim's mutable NumPy point-array storage or conflating local content-space data with current world/effective state.

## State/query contract

Path resources store immutable/local content-space geometry. Persistent path edits author a new semantic/resource version through shared operations.

Queries must distinguish coordinate space explicitly:
- local/content-space path queries operate on the retained path resource itself;
- normal current/world-space compatibility queries that include object transform observe the latest coherent **effective Runtime transform** when a live session exists;
- one query call must use one coherent frame/resource context rather than combining stale authored transforms with newer path/resource versions.

### Path construction/query
- set points as corners; start/append/close subpaths;
- line/quadratic/cubic Bezier helpers;
- anchors/handles/subpath/start/end queries where source compatibility requires them;
- `point_from_proportion`, partial-path and arc-length queries;
- reverse/subdivide/align helpers needed by transforms;
- explicit local-vs-world/current query semantics where the public API exposes both concepts.

### Smoothing
One deterministic shared `make_smooth()` implementation over canonical retained path/subpath semantics, preserving explicit subpath breaks. B5/#696 consumes the same implementation.

### Boolean geometry
Union/Intersection/Difference/Exclusion with deterministic winding/fill/cleanup rules. Boolean results are new authored immutable path resources unless explicitly used as transient effective content through C4.

## Architecture constraints

- frontend persistent path mutators use typed semantic/resource operations;
- heavy path data uses immutable versioned resources #53;
- local persistent edits produce bounded semantic/resource changes;
- current world-space queries use effective Runtime transforms rather than frontend copies;
- morph/path alignment uses the shared B3 transform pipeline and activation snapshot rules;
- static path/plot content has no per-frame Python work;
- transient dynamically regenerated paths, where needed, use C4 effective resource versions rather than semantic replacement every frame;
- no plotting-specific/boolean-specific renderer geometry model.

## Acceptance

- custom VMobject cases build deterministic multi-subpath shapes through public path calls;
- local point/proportion/partial-path semantics agree on lines/arcs/Beziers;
- world/current path queries remain correct while an active transform driver changes the object;
[...]

PR #652, opened 2026-08-30
Share analytic circle path-proportion queries

yongkyuns:
## Summary
- add a renderer-independent `point_from_geometry_proportion` boundary in `noon-geometry`
- preserve the existing ManimCE v0.21 sampled measure for `VectorPath` and exact interpolation for retained `Line`
- support analytic retained `Circle` without converting it into persistent path geometry
- reproduce ManimCE v0.21 `Circle` query semantics from its nine-component quadratic `Arc(TAU)` representation rather than using an ideal trig circle
- route the existing WASM/browser `manimPointFromProportion` bridge through the shared geometry authority and use `Transform2D::transform_point` for world-space adaptation

## Current integration
The feature is based directly on current master `9496b36b92c53eedec81f3c6cef164251215b688` after the old branch had fallen far behind. On replay head `c4a4aec626e067811e8ec863a176ad177fa9b061`, Rust unit tests, web static/unit checks, the one-time WASM build, deterministic playback, Manim authoring, and primary WebGPU rendering all passed; the sole fast-gate failure was `cargo fmt --check` in the new nested error-pattern assertion.

That formatting defect is corrected on current head `e02de0336488c2226430333cbff6f7d028bf1478` without changing behavior or tolerances. Fresh exact-head qualification is running.

## Architecture
Path semantics remain in `noon-geometry`; the browser bridge only decodes a retained snapshot and applies its retained transform. Circle support remains analytic in scene storage and only reproduces Manim's quadratic Arc measure for the query itself. No frontend geometry reconstruction, renderer primitive, or per-frame work is introduced.

## Focused coverage
- pinned Manim-compatible Circle samples at alpha 0, 0.25, 0.4, and 1
- existing Line and quadratic VectorPath measures remain unchanged
- browser bridge applies retained nonuniform scale/translation after the shared Circle query
- invalid proportions and unsupported geometry fail explicitly

Tracks #78 and unblocks #77/#256.

### DIFF
--- a/crates/noon-geometry/src/geometry_proportion.rs
+++ b/crates/noon-geometry/src/geometry_proportion.rs
@@ -143,9 +143,7 @@ mod tests {
         let circle = GeometryRef::circle(1.0);
         assert!(matches!(
             point_from_geometry_proportion(&circle, f32::NAN),
-            Err(GeometryProportionError::Path(
-                PathProportionError::InvalidProportion(_)
-            ))
+            Err(GeometryProportionError::Path(PathProportionError::InvalidProportion(_)))
         ));
         assert_eq!(
             point_from_geometry_proportion(&GeometryRef::rectangle(1.0, 1.0), 0.5),
--- a/crates/noon-geometry/src/geometry_proportion.rs
+++ b/crates/noon-geometry/src/geometry_proportion.rs
@@ -92,8 +92,7 @@ mod tests {
 
     fn assert_point(actual: Vec2, expected: Vec2) {
         assert!(
-            (actual.x - expected.x).abs() <= 1.0e-5
-                && (actual.y - expected.y).abs() <= 1.0e-5,
+            (actual.x - expected.x).abs() <= 1.0e-5 && (actual.y - expected.y).abs() <= 1.0e-5,
             "{actual:?} != {expected:?}"
         );
     }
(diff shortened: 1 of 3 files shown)


yongkyuns:
Blocking semantic issue on current head `a580c365095bfcab6dc337c20f71a159de7a52ee`: the new Circle query reproduces the OpenGL `Arc.init_points` representation (9 quadratic components), not the pinned/default Cairo semantics that our Manim differential uses. In ManimCE v0.21 `Arc.generate_points` calls `_set_pre_positioned_points`; with `num_components=9`, `np.linspace(..., num_components)` creates 9 anchors and therefore **8 cubic Bezier curves** using the `4/3 * tan(d_theta/4)` handles. `Circle` inherits that path with `angle=TAU`. This is observable: for the symmetric 8-curve Cairo circle, `point_from_proportion(0.25)` is the quarter-circle anchor `(0, radius)` (within float tolerance), whereas this PR currently locks `[-0.005740..., 2.002169...]`, which comes from the 9-quadratic path. Please make the shared query match the pinned Cairo VMobject representation / sampled-length contract, ideally by reusing the same shared Arc path construction rather than carrying a second Circle curve formula. Then qualify the samples against an actual pinned Manim oracle, not hard-coded values derived from the implementation. Do not merge this head into #77/TangentLine.

yongkyuns:
Validation note: the current head moved from `1999af300cc1b25f7bf7f2076e404b82f3ec5aab` to `a580c365095bfcab6dc337c20f71a159de7a52ee` at 22:17 UTC via `Format geometry proportion tests`. That new commit changes the `assert_point` condition back to the multiline form that rustfmt had previously rejected on the earlier CI run. I did not overwrite the concurrently updated branch. Please require fresh exact-head `cargo fmt --all -- --check` before merge; the green CI/coverage/differential results cited for `1999af3…` do not validate `a580c36…`.

PR #654, opened 2026-08-30
Keep prepared authoring execution unpublished until engine selection

yongkyuns:
Superseded by #941, merged as `a5824e4e8e9550b955ec3a6d1113d3bc5db08dc5`.

The original branch targeted the older split-retained startup API. #941 ports the same facade-level prepared-owner design onto current canonical `startRetainedCanonical(SceneSpec)` semantics, preserves later callback/mode-switch/lifecycle work, and depends on the generation-safe low-level rollback fixed in #931.

The merged replacement keeps render/WASM preparation unpublished until authored engine selection, reuses the prepared owner for legacy or canonical retained startup, inherits prepared shared-slot capacity when startup does not override it, and adopts the replacement canvas safely on cancellation.

Its exact head passed PR Fast Gate plus the dedicated playground authoring-recovery/stress workflow before merge. Closing this stale branch in favor of the current implementation.

### DIFF
--- a/web/authoring-execution-client.js
+++ b/web/authoring-execution-client.js
@@ -25,6 +25,7 @@ const EMPTY_HOST_METRICS = Object.freeze({
 export class AuthoringExecutionClient {
   #canvas;
   #player = null;
+  #preparedPlayer = null;
   #mode = null;
   #rendererBackend = "";
   #loopDurationSeconds = DEFAULT_LOOP_DURATION_SECONDS;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Sequencing blocker after #650 merge: please hold this facade tranche until #656 (`Serialize prepared execution startup`) lands and requalifies the low-level prepared-start boundary. Current master inherited #650's double-start race: two authored starts can await the same render preparation and attach competing engines. #656 adds the single-owner reservation/regression without changing the facade contract. After #656 merges, replay this two-file facade diff onto that master and require fresh exact-head CI; do not qualify/publish the facade on the pre-hotfix boundary.

yongkyuns:
Additional blocker found during facade review (independent of the #656 restack): `start()` / `startRetained()` currently guard only `#player` and `#transition`, so while a `#preparedPlayer` exists two authored starts can both enter. Post-#656 the low-level `ExecutionWorkerClient` will correctly reject the second engine attachment, but the losing facade call has already mutated `#loopDurationSeconds` and `#sharedSlotCapacity` before calling `#startMode`. With different options, the winner can therefore run with its own captured options while the facade/recovery metadata is left holding the rejected caller's values. Please add a facade-level synchronous initial-start reservation (identity/generation-safe across terminate, analogous in purpose to #656) and stage startup configuration in locals; publish `#loopDurationSeconds`, `#sharedSlotCapacity`, `#transportMode`, mode/backend only after the winning player becomes ready. Add a race regression with two starts using different loop durations/capacities and assert one engine plus winner-consistent facade/restart state. Then replay onto post-#656 master and run fresh exact-head CI.

PR #679, opened 2026-08-31
Gate retained ShrinkToCenter visual playback

yongkyuns:
## Summary

Add the missing end-to-end regression for the shipped retained `ShrinkToCenter(Text(...))` gallery example.

The existing retained animation coverage proves Python emits a `scale 1→0` track and Rust evaluates it, but it does not prove the actual browser execution/render worker presents visibly changing Text frames. A static public example can therefore escape those component-level gates.

## Regression

The new Chromium smoke runs the exact gallery source through `PythonAuthoringClient` + `AuthoringExecutionClient`, starts retained execution, pauses playback, then performs deterministic seeks to 0.25s and 0.65s. It screenshots the real transferred canvas and asserts:

- authored duration is 1 second;
- there is exactly one retained Text object and no legacy placeholder geometry;
- exactly one retained scale track lowers `1→0`;
- execution is actually in retained mode;
- both seeks are presented by the render worker at the requested logical times;
- foreground width, height, and pixel mass materially shrink between checkpoints;
- the visual center remains stable.

The retained-text-animation workflow now runs this playback test after the existing authoring smoke.

## Why

This is intentionally a test-first regression PR. If current master reproduces the reported static ShrinkToCenter behavior, the failing checkpoint will distinguish execution/transport timing from stale presentation pixels before any production fix is applied.

Related: #609, #659, #661, #665, #666, #668.

### DIFF
--- a/.github/workflows/manim-raster-differential.yml
+++ b/.github/workflows/manim-raster-differential.yml
@@ -52,6 +52,22 @@ jobs:
       - name: Install pinned ManimCE reference
         run: python -m pip install --disable-pip-version-check "manim[typst]==0.21.0" "typst==0.15.0"
 
+      - name: Report Manim default text oracle
+        run: |
+          fc-match sans-serif
+          python - <<'PY'
+          from manim import Text
+
+          text = Text("Hello World!")
+          center = text.get_center()
+          print(
+              "Manim Text('Hello World!'): "
+              f"font={text.font!r} font_size={text.font_size} "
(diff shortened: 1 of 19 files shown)


yongkyuns:
Superseded by #691, which isolates the actual retained-renderer backpressure fix and its end-to-end ShrinkToCenter(Text(...)) playback regression on current master. #691 merged as 3790457661b28f746022a321156cf86b0c7f3166. The unrelated native-text/raster-parity work in this PR should be handled separately.

Issue #705, opened 2026-08-31
[Roadmap B3] Retained text member-animation semantics


yongkyuns:
## Roadmap

Phase B3 text-member animation behavior under #954. Architecture authority: `docs/architecture.md`.

## Goal

Match ManimCE v0.21 Text family-member timing/creation behavior while preserving one authored semantic Text object, using #737/#741 for generic family/member scheduling and keeping intermediate member progress/render geometry in effective Runtime/renderer state rather than semantic mutation.

## Compatibility contract

Manim Text visibly animates rendered character/submobject members. Supported Create/Uncreate/Write/Unwrite behavior therefore needs ordered member-local progress rather than one object-wide reveal scalar.

Preserve pinned v0.21 semantics for:
- member sub-alpha calculation from global progress, member index/count and lag ratio;
- reverse rate/member behavior for Uncreate/Unwrite;
- distinct Write/DrawBorderThenFill behavior rather than aliasing Write to Create;
- discrete letter-by-letter add/remove where supported;
- exact lifecycle boundaries and intermediate member visibility/reveal state.

## State/resource model

- Text remains one authored semantic object with stable semantic identity;
- authored text source/options/style and persistent source replacement remain Semantic Scene state;
- shaped source/cluster/member descriptors are immutable/versioned resource-local metadata;
- #741 lowers those descriptors plus semantic family ordering into a derived execution member plan;
- Runtime owns current member progress/effective visibility/reveal/presentation at `FrameEpoch`;
- renderer realization may produce transient outline/path/reveal geometry from that effective state, but this does not replace the authored Text content every frame;
- persistent text/source changes use normal semantic content mutation and resource-version publication.

## Architecture constraints

- no persistent glyph semantic objects merely to animate Text;
- no Python character scheduler or per-frame Python callback for deterministic built-ins;
- no global combined-path approximation that destroys observable member ordering;
- no semantic `ReplaceContent` per frame for reveal/write progress;
- stable source/member resources remain resident across pure progress frames;
- renderer primitive substitution or temporary path realization remains derived renderer/effective output, never authored text authority;
- stale member plans/resource references cannot survive content replacement.

### B3 text member descriptors
Map shaped source/cluster information to stable ordered rendered-member descriptors, including ligatures, whitespace, multiline and non-rendered source spans where applicable.

### B3 family progress
Consume #737's generic timing and #741's global member indices/ranges. Text-specific code supplies realization capabilities only, not its own scheduler.

### B3 Write realization
[...]

Issue #89, opened 2026-08-24
[Roadmap B6] Scene ordering/lifecycle and 2D camera support


yongkyuns:
## Roadmap

Phase B6 child under #954. Architecture authority: `docs/architecture.md`.

## Goal

Complete common 2D Scene ordering/lifecycle/camera behavior on the same Semantic Scene, Runtime and Renderer used by every other feature, while excluding Manim file-writer/CLI internals.

## Camera contract

The Semantic Scene owns the authored/base camera declaration and identity. The Runtime owns the current effective camera after timeline/reactive/host/editor drivers.

Phase A/#1081 establishes the host boundary: an authored semantic camera role lowers through the canonical execution path, and `ExecutionSession::camera()` derives the current effective `Camera2DState` from the runtime frame. Native/web hosts consume that state; they do not own another mutable camera.

Camera animation/reactivity therefore changes **effective runtime camera state** frame to frame. It must not rewrite the authored camera declaration or mark every scene object semantically changed. Persistent camera edits reconcile through normal semantic mutations when required.

### B6.1 — scene ordering/lifecycle
- foreground add/remove;
- bring-to-front/back and stable family-aware painter order;
- family-aware replace/remove/membership queries;
- scene-level waits/hooks through the shared logical segment/session contract;
- lightweight sections/timeline metadata useful to browser/export workflows.

### B6.2 — 2D camera semantics
- one authored semantic camera/frame identity/declaration;
- current live camera queries observe coherent effective Runtime state;
- MovingCameraScene-compatible pan/zoom;
- deterministic/reactive camera properties use ordinary shared driver/timeline machinery;
- direct seek and forward playback agree where deterministic.

### B6.3 — viewport projection
Viewport/aspect/surface sizing remains host/renderer-boundary data. Hosts combine viewport facts with the effective semantic camera; they must not create target-specific camera semantics.

### B6.4 — ZoomedScene/compositing
Support only through the existing scene/runtime/renderer/surface model. Do not create a second scene authority/runtime merely for an inset or moving camera.

## Explicit divergences

File writer/FFmpeg/CLI filesystem semantics are not part of the runtime architecture. Browser-equivalent metadata/subcaption/audio behavior is added only when product-relevant and explicitly defined.

## Acceptance

- common foreground/front/back cases match painter semantics;
- camera pan/zoom changes effective camera state without mutating every object;
- normal live camera getters return coherent effective values;
- authored/base camera inspection, if exposed, is separate;
[...]

PR #710, opened 2026-08-31
Derive retained Text animation members from rendered glyphs

yongkyuns:
Second #705 slice, stacked on #708's shared family timing primitive.

## What this adds
- renderer-independent `TextAnimationMember` derivation from immutable `TextResource`
- preserves one public retained `Text` semantic object while exposing internal rendered-glyph membership for animation/render preparation
- matches ManimCE v0.21 Cairo's `Animation.get_all_families_zipped()` / `family_members_with_points()` model: one rendered SVG glyph submobject is one default-Text animation member
- preserves first painter-stream appearance as family order across runs
- explicitly excludes whitespace-only source spans even when the shaping backend emits an advance glyph
- ligatures naturally remain one member when shaping emits one glyph spanning multiple source characters; multi-glyph combining clusters remain multiple rendered members
- fails closed for non-plain Text, backend vector content, malformed run references/source spans, and u32 glyph-index overflow
- keeps the result derived/non-serialized: no authoring-wire expansion, no Python-owned character IDs, and no fake scene objects

## Pinned Manim contract
ManimCE v0.21 `Animation.interpolate_mobject()` enumerates Cairo `family_members_with_points()`. Default `Text` sets `chars` from its rendered SVG `submobjects`; its own docs explicitly state whitespace/newlines never become submobjects. `disable_ligatures=True` is a separate mode that reconstructs source-character slots, so default Text animation must follow rendered glyphs rather than raw Unicode characters or shaped-cluster count.

#715 remains an independent metadata invariant cleanup because native shaping currently increments `TextClusterIdentity.cluster_ordinal` per glyph despite documenting it as cluster identity. Animation membership no longer relies on that ordinal.

Tracks #705. Depends on #708. Related: #700, #61, #89, #185, #715.

### DIFF
--- a/crates/noon-core/src/reactive.rs
+++ b/crates/noon-core/src/reactive.rs
@@ -44,6 +44,10 @@ pub use font_resources::*;
 mod text_resources;
 pub use text_resources::*;
 
+#[path = "text_animation_members.rs"]
+mod text_animation_members;
+pub use text_animation_members::*;
+
 #[path = "resource_mutation.rs"]
 mod resource_mutation;
 pub use resource_mutation::*;
--- a/crates/noon-core/src/text_animation_members.rs
+++ b/crates/noon-core/src/text_animation_members.rs
@@ -1,0 +1,298 @@
+use std::collections::BTreeMap;
+
+use crate::{TextRenderItem, TextResource, TextSourceKind, TextSourceSpan};
+
+/// Stable reference to one shaped glyph inside an immutable [`TextResource`].
+///
+/// This is internal retained-content identity, not a semantic scene object. A public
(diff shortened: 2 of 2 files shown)


yongkyuns (commented):
Architecture correction after pinning the behavior directly to ManimCE v0.21 source: default Cairo `Text` animation iterates `family_members_with_points()`, and default `Text.chars` is built from rendered SVG submobjects. Therefore this PR now models one non-whitespace rendered glyph as one animation member, not one shaped source cluster. Ligatures collapse naturally when shaping emits one glyph; a multi-glyph combining cluster remains multiple rendered members. #715 still fixes `cluster_ordinal` as a separate core metadata invariant, but family animation no longer depends on it.

PR #732, opened 2026-08-31
Carry retained Text family state in execution protocol v2

yongkyuns:
Next #705 transport slice, stacked on #729.

## Architecture
- keeps the existing `noon.execution.retained` channel
- introduces explicit protocol v2 rather than silently extending v1
- delegates sequence numbers, snapshot/session rules, sparse object selection, slot/content identity, structural-change handling, and stale-drop behavior to the proven v1 encoder/mirror
- flattens the #729 `text_family_animation` field into each retained object delta
- carries no glyph IDs; renderer membership remains derived from immutable `TextResource`
- only indexes family state for objects already selected by the v1 sparse delta

## Transactional behavior
- encoder uses clone → encode → validate family state → commit, so a rejected family payload cannot consume a sequence number or initialize the stream
- mirror uses clone → apply v1 invariants → validate/update family state → commit, so malformed v2 state cannot partially mutate retained frame state
- stale deltas are classified by the proven v1 mirror before family state is touched and cannot roll back renderer state

## Safety / compatibility
- v2 mirror rejects protocol-v1 payloads instead of silently ignoring animation state
- family state is validated against object content before encode and after decode
- snapshots replace the full family-state mirror; incrementals update only transported objects, including explicit transition back to `None`

## Tests
- snapshot JSON round-trip on the same retained object stream
- sparse incremental family-state removal
- stale-delta rollback prevention
- rejected family state does not advance encoder sequence
- v1-at-v2-boundary rejection
- geometry target rejection

This PR intentionally stops before `RetainedAuthoringPlayer`/renderer wiring. After #716 and #729 land, the next slice will feed `RetainedTextFamilyFrame` directly into this encoder and expose the mirror family slice to `RetainedFramePreparer`.

Tracks #705. Related: #716, #729, #700.


yongkyuns:
Superseded by the unified family-member architecture in #741. Since this PR was opened, #744/#745/#746/#749 established authoritative semantic-family leaf traversal, retained Text/geometry member descriptors, one global member index space, and global lag/reversal evaluation. Merging a public protocol-v2 field named `text_family_animation` now would freeze the earlier per-object Text abstraction that #741 explicitly replaces. The useful invariants from this PR—sparse transport, transactional validation, stale-drop safety, no serialized glyph IDs—should be carried into the eventual generic family-plan transport rather than merged under a Text-specific wire contract. Exact head `cd018074...` is green; closing is architectural supersession, not a validation failure.

PR #740, opened 2026-09-01
Cache retained Text family progress evaluation

yongkyuns:
Renderer-preparation prerequisite for #705.

## What this adds
- `TextFamilyProgressEvaluator`, which validates one `TextFamilyAnimationState` and resolves its `UniformFamilyAnimationPlan` once for a rendered member count
- O(1), allocation-free per-member progress lookup after construction
- preserves Manim ordering: member lag -> optional reverse-rate -> rate function
- preserves independent `reverse_member_order` semantics for reversed Write/Unwrite

## Why
`TextFamilyAnimationState::member_progress()` is a convenient correctness API but constructs the uniform composition plan for each call. The retained renderer will visit every rendered Text glyph each animated frame, so it should construct the plan once per animated object/frame rather than once per glyph.

## Tests
- Create lag-one progression
- Uncreate reverse-rate behavior
- reverse-member-order independent from reverse-rate
- invalid state, empty family, and out-of-range member rejection

No renderer or transport behavior changes in this PR. Tracks #705; intended for the subsequent retained Text renderer integration.

### DIFF
--- a/crates/noon-core/src/reactive.rs
+++ b/crates/noon-core/src/reactive.rs
@@ -52,6 +52,10 @@ pub use text_animation_members::*;
 mod text_family_animation;
 pub use text_family_animation::*;
 
+#[path = "text_family_evaluator.rs"]
+mod text_family_evaluator;
+pub use text_family_evaluator::*;
+
 #[path = "resource_mutation.rs"]
 mod resource_mutation;
 pub use resource_mutation::*;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Superseded by #741's generic global family-member planner. `TextFamilyProgressEvaluator` caches a per-Text uniform plan, but #749 now evaluates leaf-local members through one global member sequence and #750 prepares those spans/descriptors once for the renderer. Keeping this Text-only evaluator would create a second timing path and break mixed-family lag semantics. Closing in favor of #749/#750/#758.

Issue #741, opened 2026-09-01
[Roadmap A1/B3] Unified animation-member plans


yongkyuns:
## Roadmap

A1/B3 lowering bridge under #953/#954. Architecture authority: `docs/architecture.md`.

## Goal

Create one **derived execution member plan** that flattens Manim-visible members across semantic object families and resource-local members while preserving one global ordered sequence, without promoting renderer/resource members into authored semantic identity.

Example:

```text
VGroup
├── Text("AB") -> resource-local members A, B
└── Circle      -> one semantic/content member

Global animation order: A, B, Circle
```

The Semantic Scene owns family/content/animation intent. This plan is lowering output. Runtime owns current member progress/effective presentation; renderer realization consumes the plan plus Runtime state.

## Target plan

A lowering-time plan contains compact identity/count/range/capability metadata such as:
- originating semantic animation/family/object identity;
- operation/timing reference;
- total global member count;
- ordered semantic leaf identities and execution bindings;
- each leaf's first global member index + local member count;
- resource identity/version or descriptor generation needed to validate resource-local members;
- required realization capability.

It does not copy glyph/path/mesh payloads, current frame transforms, or mutable per-frame progress.

## Revision/invalidation contract

The member plan is derived execution state:
- it may change when authored family order/topology, animation intent, content/member descriptors or relevant resource versions change;
- such a change participates in `ExecutionRevision` publication under #1087;
- pure animation progress, transform/style changes, camera motion and other effective `FrameEpoch` updates do **not** rebuild the plan when its structure/bindings/resources remain valid;
- stale plans are rejected when semantic/execution identity generations or referenced resource versions no longer match;
- `SceneRevision` is not a blanket cache key: an unrelated authored edit must not invalidate this plan.

### Semantic leaf traversal
Use authoritative A1/B1 family ordering with deterministic nested traversal and generation/cycle validation. Traversal reads authored family structure, not a renderer/frontend copy.

### Content-local descriptors
Content types report ordered member descriptors without creating persistent semantic children solely for rendering. Retained Text uses source/cluster/member metadata; path/analytic geometry supplies its own capability metadata; future SVG/mesh types follow the same interface.

### Global flattening
Lag/easing/reversal uses one global member count/index across semantic leaves and resource-local members. The plan preserves exact observable ordering without concrete-content scheduler branches.

### Runtime binding
[...]

PR #754, opened 2026-09-01
Carry retained Text family animations through authoring wire

yongkyuns:
Next #705 end-to-end prerequisite, stacked on #732.

## What this does
- adds optional `text_family_animations` to the existing retained authoring protocol-v2 payload
- keeps object-only and ordinary-track documents backwards-compatible
- preserves `TextFamilyAnimationDefinition` as raw semantic timing state instead of lowering it into generic property tracks
- carries the definitions alongside `MixedRetainedAuthoringScene` so the production player can install `RetainedTextFamilySceneInstance` in the next slice
- leaves glyph membership out of the authoring wire; render members remain derived from immutable `TextResource`

## Why separate from ordinary tracks
Manim family timing applies lag/member mapping before the rate function. Encoding this as a generic `Reveal` track would consume easing too early and recreate #700's whole-string reveal bug.

## Tests
- object-only v2 remains valid and has no family definitions
- existing retained property tracks are unchanged
- family sidecar round-trips mode/start/duration/lag/rate/reverse flags without creating ordinary tracks

This PR intentionally stops before changing the large production player file. The next slice swaps `RetainedAuthoringPlayer` to `RetainedTextFamilySceneInstance` + protocol-v2 encoder against this typed sidecar.

Tracks #705. Related: #700, #716, #729, #732.

### DIFF
--- a/crates/noon-web/src/retained_authoring_wire_scene.rs
+++ b/crates/noon-web/src/retained_authoring_wire_scene.rs
@@ -1,6 +1,6 @@
 use noon::RetainedScene;
 use noon_compile::RetainedCompiledScene;
-use noon_core::{ObjectId, SceneDefinition, TrackDefinition};
+use noon_core::{ObjectId, SceneDefinition, TextFamilyAnimationDefinition, TrackDefinition};
 use serde::Deserialize;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


yongkyuns:
Superseded by the #741/#737 unified semantic-family path. This PR carries a Text-only authoring sidecar into `RetainedTextFamilySceneInstance`; that was appropriate before semantic family targets gained target-independent `FamilyAnimationSpec` and one global member plan. The next authoring/runtime integration should lower semantic family animations into the generic plan/state path so Text, geometry, VGroup, Axes, and graph families share one scheduler. Closing this Text-specific wire slice to avoid locking in a parallel compatibility channel.

PR #756, opened 2026-09-01
Evaluate retained family progress against resolved members

yongkyuns:
Renderer-preparation slice for #705, stacked on #732.

- extends `RetainedAnimationMembers` with allocation-free per-member progress evaluation
- combines authoritative resolved member count/order with shared `FamilyAnimationState`
- keeps lag, rate reversal, and member-order reversal in core instead of duplicating them in WebGPU/WebGL/native renderers
- covers whitespace-filtered Text membership, Create-style lag, Uncreate reverse-rate behavior, reverse-member-order behavior, and invalid member indices

No transport, authoring, or GPU draw submission changes in this PR. Tracks #705; related #700, #732.

### DIFF
--- a/crates/noon-core/src/retained_animation_members.rs
+++ b/crates/noon-core/src/retained_animation_members.rs
@@ -1,6 +1,6 @@
 use crate::{
-    plain_text_animation_members, ObjectContentRef, TextAnimationMember, TextAnimationMemberError,
-    TextResourceArena, TextResourceHandle,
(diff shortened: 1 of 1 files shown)


yongkyuns:
Superseded by #749/#750. This PR evaluates progress against one retained leaf's resolved members, but #741 now requires lag/reversal over the authoritative global member sequence across all semantic leaves. #749 provides that global leaf-progress mapping and #750 binds retained Text/geometry descriptors to it. Closing to avoid a second local-only evaluator.

Issue #731, opened 2026-08-31
Make PR CI feedback consistently sub-2-minute


yongkyuns:
## Goal
Make the normal pull-request feedback loop consistently complete in about 2 minutes without reducing the repository's full correctness coverage.

## Result
Completed. Noon now has one unconditional pull-request workflow, `.github/workflows/pr-fast.yml`, with four substantive jobs running in parallel:
- Rust format/lint
- Rust unit tests
- web static/unit preflight
- one browser/WASM package build followed by deterministic replay, Manim-style authoring, primary WebGPU smoke coverage, plus change-aware high-risk integration escalation

Exhaustive compatibility, coverage, recovery, platform/release, fuzz, and broader renderer matrices remain on master/manual/scheduled paths instead of independently cold-starting for every PR.

## Acceptance criteria
- [x] Typical PR creates one primary fast-gate workflow instead of a large set of independently bootstrapped workflows.
- [x] Fast gate targets <=120 s on warm hosted runners.
- [x] Web package is built at most once in the fast PR path.
- [x] Specialized/exhaustive validation remains available off the PR critical path.
- [x] Shared-core/high-risk changes can opt into broader validation without making every PR pay the cost.
- [x] CI responsibilities and cache ownership are documented and easy to extend.
- [x] Timing regressions are visible without turning noisy hosted-runner wall time into a fragile correctness check.

## Completed slices
- [x] Consolidate normal PR validation into `PR Fast Gate`. #734.
- [x] Remove duplicate heavyweight/feature-specific `pull_request` entry points while preserving master/manual/scheduled validation. #734.
- [x] Parallelize independent Rust lint/test compilation and web/browser validation. #734.
- [x] Remove the fifth no-op aggregate runner; the four real jobs are the validation signals.
- [x] Retain the full deterministic-replay corpus, including the 1,000-object morph fixture and 32 forward samples per target, while reducing replay from ~35–40 s to ~3–5 s through one compiled scene/runtime verifier and Rust-side observable-state comparison. #743.
- [x] Remove PR debug information from Cargo dev/test profiles after warm measurement showed the Rust test compile path fall from ~91 s to 53.10 s.
- [x] Separate native and WASM compiler-cache transports based on measured behavior; keep native GHA sccache read-only on PRs and use a trusted master-owned local WASM archive for the browser lane.
- [x] Align PR WASM archive lookup to the PR base/master seed and bundle local WASM sccache objects with the wasm-pack helper cache. #793, #797.
[...]

PR #793, opened 2026-09-01
Align PR WASM cache keys with master seed

yongkyuns:
## Summary
- keep `CARGO_PROFILE_TEST_DEBUG=0` scoped to the Rust lint/test jobs that actually need the test-profile override
- remove that irrelevant `CARGO_*` variable from the browser job so its sccache key environment exactly matches the master WASM seed
- retain the source-neutral CI-policy note documenting how compiler-cache changes should be measured

## Root cause
#787 fixed archive transport: the PR restored the exact ~297 MiB master archive in ~2 s. But the first source-neutral probe still saw **0/315** compiler hits.

A control rerun of the exact same master seed on the exact same SHA restored the same archive and got **315/315 hits (100%)**, reducing Rust WASM compilation to ~26 s. Comparing job environments exposed the mismatch: PR Fast Gate globally set `CARGO_PROFILE_TEST_DEBUG=0`, while the WASM seed did not. sccache hashes `CARGO_*` environment variables into Rust cache keys, so a test-only setting was invalidating every browser compiler object.

The clean fix is to scope the test-profile override to the Rust jobs rather than teaching the browser/seed lanes about an unrelated test setting.

## Validation
The updated PR remains Rust-source-neutral and is based on the seeded master archive. The decisive signal is the browser lane: restore the exact default-branch archive, report high/complete sccache hits, and measure total wall time against the <=120 s target.

Tracks #731.

### DIFF
--- a/.github/workflows/pr-fast.yml
+++ b/.github/workflows/pr-fast.yml
@@ -13,7 +13,6 @@ concurrency:
 env:
   CARGO_INCREMENTAL: "0"
   CARGO_PROFILE_DEV_DEBUG: "0"
-  CARGO_PROFILE_TEST_DEBUG: "0"
   RUSTC_WRAPPER: sccache
   SCCACHE_GHA_ENABLED: "true"
   SCCACHE_GHA_RW_MODE: "READ_ONLY"
@@ -22,6 +21,8 @@ jobs:
   rust-lint:
     name: Rust format and lint
     runs-on: ubuntu-latest
+    env:
+      CARGO_PROFILE_TEST_DEBUG: "0"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


yongkyuns:
Measurement result: the exact default-branch archive restored successfully (`Linux-sccache-wasm-v1-be8d27c...`, ~297 MiB, ~2 s restore), but the source-neutral PR still reported **0/315 Rust cache hits**. The browser package build took ~156 s and the browser lane remained ~3m15s, while preflight (~26 s), lint (~45 s), and Rust tests (~56 s) were all comfortably under budget. This isolates the remaining problem to compiler-cache key identity rather than archive transport. I am rerunning the exact same master seed job on the same SHA as a control to determine whether the mismatch is PR-merge-specific or local-sccache-wide.

PR #825, opened 2026-09-01
Scope retained native Text identity by authoring store

yongkyuns:
Advances #741 through the live WASM authoring identity boundary after #816.

## What changes

- `AuthoringSemanticIdentity` gains a crate-private WASM bridge that wraps the existing shared semantic-store owner plus node ID without exposing either as a new frontend-controlled token.
- `WasmAuthoringFamilyMemberHandle` derives that opaque store-scoped identity from the shared Rust store it already owns.
- `RetainedNativeTextAuthoringHandle` now retains the full store-scoped identity rather than only a numeric `SemanticNodeId`.
- retained Text family/layout/arrange validation compares both semantic node and originating authoring store.
- rebinding the exact same identity remains idempotent; an equal numeric node ID allocated by another store is rejected.

## Architecture

No JS/Python or renderer-wire shape changes. Store ownership stays represented by Rust `Rc` identity and never crosses the frontend boundary. Existing family/layout sessions continue to validate their member handles with Rust store pointer identity; retained native Text now carries the same ownership invariant instead of trusting slot/generation alone.

This deliberately stops before introducing the final retained `ObjectId` materialization binding session. With Text identity now safe, the next #741 slice can bind live Mobject/Text handles to individual retained `ObjectId`s and feed `RetainedFamilyAuthoringLoweringSession` without ever sending ordered family member IDs through JavaScript.

## Tests

Adds a regression that constructs two independent authoring stores which allocate identical numeric node IDs, binds Text identity to the first, verifies same-identity rebinding is idempotent, and rejects the colliding identity from the second store.

### DIFF
--- a/crates/noon-web/src/authoring_mobject.rs
+++ b/crates/noon-web/src/authoring_mobject.rs
@@ -1484,7 +1484,7 @@ mod wasm {
 
     use wasm_bindgen::prelude::*;
 
-    use crate::WasmRetainedNativeTextAuthoringHandle;
+    use crate::{AuthoringSemanticIdentity, WasmRetainedNativeTextAuthoringHandle};
 
     use super::{
         manim_family_align_to_delta, manim_family_next_to_delta, render_f64,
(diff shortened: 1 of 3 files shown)


yongkyuns:
Exact head `00e95bbdf168b63653a1cfcb2671a476fcc8d791` passed the full PR Fast Gate: Rust tests, rustfmt/Clippy, web preflight, WASM build, deterministic playback, Manim-style authoring, canonical retained execution, and primary WebGPU rendering. Rechecked `master` immediately after the gate; it remains `e6dcc2c2d8755290079b74b60cb1b4fdc61f67d4`, the PR base, and the PR is mergeable.

Issue #835, opened 2026-09-01
[Roadmap C3] Resident incremental family-animation rendering


yongkyuns:
## Roadmap
Phase C3 child of #955. Architecture authority: `docs/architecture.md`.

## Goal
Make family-animation rendering obey the same locality invariant as ordinary retained rendering: after immutable member content is resident, animation progress changes only small mutable member state.

## Architecture rule
> Family timing changes mutable member presentation state; immutable glyph/path/content stays resident.

Use the shared semantic family/member plan from #737/#741 and the incremental renderer state from #362. Do not create a Text-specific scheduler, persistent per-glyph semantic objects, or a parallel scratch-frame renderer.

### C3.1 Resident member bindings
Bind each lowered semantic leaf/member range to already-resident atlas/path/geometry resources plus compact mutable state such as reveal/border-fill/discrete visibility.

### C3.2 Progress-local updates
When family topology/content/resource identity is unchanged:
- update only affected member state/ranges;
- do not rematerialize equivalent `VectorPath`/glyph geometry;
- do not reprepare unrelated objects;
- do not rebuild global mixed painter order.

### C3.3 Lane transitions
Localize transitions between steady atlas-backed text and outline/path-level animation behavior. Do not keep every glyph permanently vectorized to simplify animation.

### C3.4 Generic realization boundary
The renderer seam must support future VMobject/SVG/graph/mesh member resources without adding concrete-type branching to the generic family scheduler.

## Acceptance
For a mixed `VGroup(Text("AB"), Circle())` embedded in a large static scene, after warmup:
- family plan/member order remains stable;
- outline/resource cache misses are zero for already-resident content;
- progress-only frames do not rematerialize equivalent geometry;
- unrelated objects are not re-prepared;
- mixed painter order is not globally rebuilt;
- GPU writes/ranges scale with affected members;
- direct seek == forward playback;
- WebGPU/WebGL output remains equivalent where supported.

## Dependencies
#737/#741 family semantics/member plans and #362 incremental renderer state.

PR #836, opened 2026-09-01
Document retained renderer locality and caching

yongkyuns:
## Summary
- document the renderer work-avoidance principles behind Noon's retained architecture
- distinguish all-live resource residency from visible draw projection
- pin the target `O(visible + dirty)` preparation/submission model
- define invalidation rules for transform/style/reveal/content/structure changes
- record why generic per-object bitmap caches are not the default fit for Noon
- document the current viewport-culling and dirty-text trackers
- add the new family-animation locality gap tracked by #835
- keep render-bundle/static-subtree raster caching profile-gated rather than creating speculative implementation work

## Issue alignment
- #569 updated to current viewport-culling state and the remaining #605 integration path
- #362 updated so retained family playback must inherit dirty-range renderer locality
- #835 opened for resident/incremental family-animation render realization
- #68 annotated with the refined render-preparation/submission invariant

Documentation only; no runtime behavior changes.

### DIFF
--- a/docs/renderer-locality-and-caching.md
+++ b/docs/renderer-locality-and-caching.md
@@ -1,0 +1,307 @@
+# Renderer locality and caching
+
+## Purpose
+
+Noon should borrow the **work-avoidance principles** used by high-performance retained/cached drawing systems without copying Canvas-specific implementation choices.
+
+The architectural goal is simple:
+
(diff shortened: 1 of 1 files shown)


yongkyuns:
Superseded by #849. The renderer-locality document was replayed onto current master there and expanded with the ManimGL review, interactive/live-authoring architecture, and #847 measurement gate. Closing this stale docs-only branch to keep one current documentation path.

PR #838, opened 2026-09-01
Gate retained Text playback before merge

yongkyuns:
## Summary
- run the dedicated retained Text animation workflow on pull requests touching the same owned paths already checked after master pushes
- preserve the post-merge master gate for direct/merge validation
- add a focused policy test requiring PR and master path filters to remain identical, including the retained playback regression and core Python animation owner
- run that policy test in the normal web preflight

## Why
Current master `54809da42fdec8ee3a1ff31c0b3d293b673f5bbb` fails the dedicated `Retained text animation` workflow at `Test retained Text animation playback` (run `33564244424`), while the preceding authoring/state/family checks pass. The merged PR #834 changed `crates/noon-web/src/retained_authoring_wire_scene.rs` and `web/python-compat-modules.js`, both already inside this workflow's path ownership, but the workflow previously ran only on pushes to `master`. The regression therefore could not be caught by this dedicated suite before merge.

This change does not alter retained Text implementation or relax the failing visual assertion. The active ShrinkToCenter/text fixes remain separate; this PR only moves the existing regression gate to the correct pre-merge boundary.

## Validation
Opening this PR intentionally exercises the new pull-request trigger because the workflow file itself is in the owned path list. Web preflight also runs `scripts/retained-text-workflow-policy.test.mjs` to prevent future trigger drift.

## Remaining risk / merge condition
The new retained Text PR check is expected to reproduce the currently known master playback failure until the active implementation fix lands. Keep this PR draft/blocked until that underlying regression is green; once fixed, the same suite becomes the pre-merge guard that would have caught #834.

### DIFF
--- a/.github/workflows/retained-text-animation.yml
+++ b/.github/workflows/retained-text-animation.yml
@@ -28,6 +28,31 @@ on:
       - "scripts/retained-text-playback-smoke.mjs"
       - "scripts/retained-text-scene-lifecycle-smoke.mjs"
       - ".github/workflows/retained-text-animation.yml"
+  pull_request:
+    paths:
+      - "crates/noon-web/src/authoring_mobject.rs"
+      - "crates/noon-web/src/retained_authoring_player.rs"
+      - "crates/noon-web/src/retained_authoring_scene.rs"
+      - "crates/noon-web/src/retained_authoring_tracks.rs"
+      - "crates/noon-web/src/retained_authoring_wire_scene.rs"
+      - "web/authoring-render-worker.js"
+      - "web/python-worker.source.js"
+      - "web/python-compat-modules.js"
+      - "web/python/_manim_animate.py"
+      - "web/python/_manim_animation_options.py"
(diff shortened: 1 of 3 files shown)


yongkyuns:
Exact-head validation update for `7c2986ad042c82e8a7b7f6f148b215f0c6f7c06b`: PR Fast Gate web preflight is green, so the new workflow-policy regression test passes; Rust formatting/lint is also green. The newly-added pull-request `Retained text animation` run `33565017516` triggered successfully and is currently building the browser package. This PR remains intentionally draft while the known master ShrinkToCenter(Text) playback regression is active.

yongkyuns:
Roadmap triage after #952: closing this old retained-Text CI draft. The principle is still correct—owned exact-output gates should run pre-merge—but the branch is tied to an old master regression/path set. Reintroduce the retained-text-specific gate as an A6/B7 ratchet only if it is still needed after current #899/general parity gating and the Phase A ownership cleanup.

Issue #367, opened 2026-08-27
[Arch C3.3] Unify geometry and text source authoring into one canonical mixed scene


yongkyuns:
Parent: #361
Depends on: completion/stabilization of #352-#356 stack
Related: #61, #259, #51, #53, #63

## Goal
Retire the transitional split between legacy geometry authoring and retained text sidecar authoring by introducing one canonical mixed semantic authoring scene/document that can represent geometry and text source specs directly.

The current sidecar path is correct migration scaffolding. This issue owns its retirement path.

## Current transitional topology
Python/browser authoring currently needs separate channels because the legacy scene document is geometry-only:

```text
legacy SceneDocument (geometry)
        +
retained_document sidecar (Typst source)
        |
        v
MixedRetainedAuthoringScene
        |
        v
RetainedScene
```

This also forces special retained object-ID allocation/painter-order bookkeeping in the Python compatibility wrapper.

## Target
One language-neutral canonical authoring model:

```text
SceneSpec
  objects: [
    ObjectSpec { id, order, content: Geometry(...) },
    ObjectSpec { id, order, content: Text(TextSpec { source_kind, source, options... }) }
  ]
  tracks / families / camera / metadata
```

Text source specs remain source-level until shared Rust compilation/lowering. They must not contain glyph arrays, font bytes, SVG, vectorized glyph outlines or atlas state.

## Scope
- define canonical mixed authoring content/spec type capable of geometry and backend-neutral text source definitions;
- preserve stable `ObjectId`, family identity and global painter order in one object list;
- make public Rust authoring produce this canonical model directly or through one authoritative semantic store;
- expose the same shared semantic handles to Python and JS per #61/#259;
- adapt worker/runtime compilation so text source lowering and geometry compilation happen from one scene representation;
- preserve existing legacy scene decode as a compatibility/import adapter during migration;
- remove special retained text ID range once canonical ID allocation owns all semantic objects;
- remove `retained_document` sidecar from normal authoring result once all supported text paths are migrated;
- keep JSON as transport/debug/export rather than normal mutation/query API.

## Compatibility / migration
Do not break geometry-only consumers abruptly. Provide a bounded migration sequence:
1. canonical mixed model + adapters from legacy geometry document and current retained sidecar;
2. Python/Rust/JS authoring switched to canonical model;
3. browser engine path consumes canonical model directly;
4. sidecar/special-ID compatibility path deprecated/removed after tests prove equivalence.

## Acceptance
[...]

PR #848, opened 2026-09-01
Make canonical SceneSpec mandatory for authored scenes

yongkyuns:
## Summary
- canonicalize every Python-authored `Scene` through the existing Rust `SceneSpec` adapter, including geometry-only scenes
- make canonical `SceneSpec` a required result at the current authoring protocol boundary instead of an optional retained-text-only attachment
- keep legacy `SceneDocument` and retained sidecar payloads as compatibility projections during #367 migration
- preserve the existing Rust-owned geometry/text merge, text-source normalization, retained-track materialization, and family-animation lowering; no new frontend IR is introduced

## Why this slice
After #841, geometry and retained Text already share one object-ID, painter-order, and lifecycle domain, but the worker still emitted canonical `SceneSpec` only when retained text objects were present. That left canonical transport conditional on content kind. This PR removes that content-kind branch before producer ownership moves fully to `SceneSpec`.

## Validation
- focused authoring client + SceneSpec result tests
- generated worker rebuild
- `git diff --check`

Tracks #367.

## Next boundary
Move Python producer ownership from the split legacy document + retained compatibility sidecar into direct canonical `SceneSpec` object construction, while deriving the old payloads only as compatibility projections until their consumers are retired.

### DIFF
--- a/compat/wire/v1/authoring-result-empty-scene.json
+++ b/compat/wire/v1/authoring-result-empty-scene.json
@@ -1,1 +1,1 @@
-{"kind":"scene_document","document":{"version":1,"objects":[],"tracks":[]},"duration":0.0,"identities":{"objects":[],"tracks":[]},"callbacks":null}
+{"kind":"scene_document","document":{"version":1,"objects":[],"tracks":[]},"scene_spec":{"version":1,"objects":[],"tracks":[]},"duration":0.0,"identities":{"objects":[],"tracks":[]},"callbacks":null}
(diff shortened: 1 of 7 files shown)


yongkyuns:
Exact-head qualification complete at `704acdb6e6ac785d59eafb12ad1cfb556852d210` against unchanged master `0c79da85500374639ae756756834b587f09c71cb`. PR Fast Gate run 33568325947 is fully green: Rust unit tests, rustfmt/Clippy, web preflight, deterministic playback, Manim-style authoring, and primary WebGPU rendering. The broader preflight also updated stale race and wire fixtures to the mandatory canonical `SceneSpec` protocol generation.

PR #851, opened 2026-09-01
Gate retained Typst authoring before merge

yongkyuns:
## Summary
- run the dedicated retained Typst authoring workflow on pull requests touching the same owned paths already checked after pushes to `master`
- preserve the post-merge master validation
- add a focused policy regression that requires the pull-request and push path filters to stay identical
- run that policy regression in the normal web preflight

## Why
Current master `f817a16f71d848b5008d26f387206963f70f6bae` fails the dedicated `Retained Typst authoring` workflow at `Test retained Typst Python authoring`; the browser package itself builds successfully. The same workflow also failed on the preceding master head after #841.

The workflow is currently push-only even though `web/python-worker.source.js` is one of its owned paths. #848 changed that file, so the dedicated retained Typst suite could not catch the regression before merge.

This PR does not weaken the smoke assertions or guess at the underlying Typst failure. It moves the existing regression boundary to PR time and ratchets that ownership policy so future trigger drift fails normal preflight.

## Validation
Opening this PR should itself exercise the new pull-request trigger because the workflow file is in its owned path set. The new policy test is also part of `scripts/build-web-demo.sh` preflight.

## Remaining risk / merge condition
The newly-triggered retained Typst check may reproduce the existing master smoke failure. Keep this PR draft until the underlying regression is understood or fixed; the CI-boundary change itself is independent and intentionally does not mask the failure.

### DIFF
--- a/.github/workflows/retained-typst-authoring.yml
+++ b/.github/workflows/retained-typst-authoring.yml
@@ -12,6 +12,15 @@ on:
       - "scripts/build-python-worker.mjs"
       - "scripts/manim-typst-authoring-smoke.mjs"
       - ".github/workflows/retained-typst-authoring.yml"
+  pull_request:
+    paths:
+      - "crates/noon-web/src/retained_authoring.rs"
+      - "web/python-worker.source.js"
+      - "web/python-compat-modules.js"
+      - "web/python/_manim_typst.py"
+      - "scripts/build-python-worker.mjs"
+      - "scripts/manim-typst-authoring-smoke.mjs"
+      - ".github/workflows/retained-typst-authoring.yml"
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


yongkyuns:
Exact-head validation update for `c8a6fcf768f3406fe751b1da45f6a7954eaa4aa7`: the newly-added pull-request `Retained Typst authoring` workflow triggered successfully as run `33570297107`, proving the missing pre-merge boundary is active. PR Fast Gate also started on the same exact head; formatting has passed and web preflight is currently running the branch including the new workflow-policy regression. The PR remains intentionally draft because current master still fails the retained Typst smoke itself, and this change does not weaken or bypass that regression.

Issue #856, opened 2026-09-02
[Roadmap B5] Deterministic retained implicit-curve contouring


yongkyuns:
## Roadmap
Phase B5 child of #85/#954. Architecture authority: `docs/architecture.md`.

## Goal
Implement ManimCE v0.21 `ImplicitFunction` / `CoordinateSystem.plot_implicit_curve` using a deterministic shared isoline planner that emits ordinary retained path content.

## Pinned algorithm oracle
Use the behavior of ManimCE v0.21's `isosurfaces` dependency for semantic fixtures, including:
- adaptive quadtree/min-depth/cell-budget behavior;
- tolerance and stopping rules;
- defined/NaN boundary subdivision;
- sign-change descent;
- multiresolution stitching/topology;
- binary zero-crossing refinement/asymptote rejection;
- deterministic connected-curve traversal/closed-loop closure;
- optional shared smoothing after path creation.

## Architecture
- isoline traversal/topology is shared deterministic Rust geometry logic;
- scalar function evaluation is separated from contour planning so arbitrary Python functions are sampled only at author/compile time for static implicit plots;
- coordinate-space contours map through #695 Axes semantics;
- output is ordinary immutable retained VectorPath/subpath content;
- smoothing reuses B2/#78;
- no renderer ImplicitFunction primitive and no frame-time Python for static contours.

## Acceptance
- focused tests cover quadtree budget/min-depth, NaN rules, zero refinement, topology/order and closed-loop behavior against the pinned oracle;
- unsmoothed and default-smoothed behavior are explicit—never silently downgrade smoothing;
- transformed Axes map contours correctly;
- representative circle/hyperbola/asymptote fixtures pass semantic/raster qualification;
- static contours have zero timeline/host work after authoring;
- deterministic sample ordering is identical across supported native/browser authoring paths.

## Non-goals
- no coarse fixed-grid approximation substituted for the pinned supported behavior;
- no renderer special case;
- reactive/per-frame implicit-field regeneration belongs to Phase C and must use shared dependency/content-replacement machinery.

## Dependencies
#695 coordinate semantics and B2/#78 retained path/smoothing. B7/#185 owns qualification.

yongkyuns:
Claiming a bounded non-conflicting first slice on `codex/b5-implicit-contour-planner`, based on current master `c4e78ef4d33c6aa447d45ea62d1235df38e1b10d`.

[...]

PR #828, opened 2026-09-01
Add retained PolarPlane geometry and mapping

yongkyuns:
## Summary
- add a shared Rust `PolarPlaneGridPlan` for pinned ManimCE v0.21 concentric-circle and radial-line subdivision
- reuse existing retained `Circle` / `Line` geometry and the qualified `Axes2DState` / NumberLine unit-size substrate; no renderer geometry kind is added
- expose a thin WASM serialization bridge and stable authoring-store factory
- add the Python `PolarPlane` facade with exact public defaults/validation for radius, size, azimuth units/steps/offset/direction, background/faded styles, and family composition
- add shared `polar_to_point`, `point_to_polar`, `pr2pt`, and `pt2pr` helpers over authoritative transform-safe `c2p` / `p2c`
- keep coordinate labels and nonlinear deformation explicitly deferred to their retained substrates
- add a dedicated real Pyodide/WASM browser smoke and Fast Gate step

## Architecture
The Rust plan owns subdivision, style validation, retained primitive construction, background/faded classification, and upstream family ordering. Python owns only public option normalization and facade composition. Static output remains ordinary retained Line/Circle leaves with zero tracks.

## Browser acceptance
The dedicated fixture checks:
- exact background/faded radial and circle counts for `radius_max=2`, `size=8`, `radius_step=1`, `azimuth_step=4`, `faded_line_ratio=2`
- upstream family order in the serialized scene
- custom-size NumberLine unit scaling
- azimuth offset geometry
- shared polar helpers on both Axes and PolarPlane
- transform-safe polar round trips
- subclass-preserving independent copy
- ratio-0 behavior and non-integer azimuth divisions
- invalid azimuth units/direction
- explicit label deferral
- Line/Circle-only retained scene output with zero tracks

## Exact-head qualification
- exact head: `85ebaa57849a39e3bbd42409a13aa405947fac49`
- PR Fast Gate run: `33556842406` (#340)
- result: **green** across Rust formatting + Clippy, Rust unit tests, web static/unit checks, one-time WASM build, deterministic playback, Manim authoring, retained Axes plotting, retained NumberLine authoring, retained ComplexPlane authoring, dedicated retained PolarPlane acceptance, lazy namespace dependencies, literal PlotExample, and primary WebGPU rendering

Stacked on exact-green #821.

Tracks #824 / #85. Related label work: #83.

Issue #85, opened 2026-08-24
[Roadmap B5] Coordinate systems and plotting


yongkyuns:
## Roadmap

Phase B5 plotting parent under #954. Architecture authority: `docs/architecture.md`.

## Goal

Support common ManimCE v0.21 coordinate/plotting behavior from shared semantic coordinate definitions, retained resources and the standard Runtime driver classes—without frontend coordinate math or plot-specific renderer/runtime architecture.

## State/resource contract

- authored coordinate ranges, sampling declarations and static plot source belong to the Semantic Scene;
- live coordinate queries observe coherent effective Runtime transforms per #695;
- static sampled curves compile to immutable retained path resources;
- a persistent source/function edit uses semantic content/declaration mutation;
- transient native/host dynamic output may publish a versioned **effective runtime content/resource** result through C4/#368 without rewriting authored content or incrementing semantic revision every frame.

### Coordinate systems
NumberLine/UnitInterval, Axes, NumberPlane, ComplexPlane and PolarPlane with shared coordinate conversions/tick/label semantics. ThreeDAxes remains Phase D.

### Static curves/functions
ParametricFunction, FunctionGraph and supported ImplicitFunction with deterministic sampling/discontinuity policy and retained graph metadata/queries.

### Derived helpers
Vertical/horizontal lines, labels, area/Riemann geometry, line graphs and derivative/secant helpers where they compose shared primitives.

### Dynamic functions
- native/reactive functions use shared dependency evaluation plus effective resource/content updates only when dependencies change;
- arbitrary Python functions use C1 host callbacks only where unavoidable;
- persistent user/source edits remain semantic mutations;
- dynamic behavior must not make static/unrelated plots host-driven or trigger full-scene relowering.

## Architecture constraints

- coordinate transforms/tick/query math is shared Rust behavior;
- axes/planes/plots are normal semantic families/resources, not renderer primitives;
- live queries use effective state rather than stale authored transforms;
- static plots have no frame-time Python/content regeneration;
- dynamic regeneration is dependency-local and versioned;
- transient runtime regeneration is not automatically semantic `ReplaceContent`;
- sampling/contouring policy is deterministic/backend-neutral.

## Implementation children

- #695 NumberLine/Axes coordinate/live-query foundation;
- #696 deterministic function/parametric plotting;
- #778 area/Riemann geometry;
- #804 NumberPlane;
- #810 NumberLine facade;
- #856 implicit curves.

## Acceptance

[...]

PR #860, opened 2026-09-02
Add deterministic retained implicit-curve contouring

yongkyuns:
## Summary
Add ManimCE v0.21 `ImplicitFunction` and `Axes.plot_implicit_curve` on a shared deterministic retained contour substrate.

- ports the adaptive 2D topology semantics of Manim's pinned `isosurfaces>=0.1.1` dependency using the v0.1.1 release as the compatibility oracle
- breadth-first quadtree with exact min-depth/cell-budget precedence and NaN-aware refinement
- multiresolution simplicial stitching, binary zero refinement/asymptote rejection, deterministic connected-curve tracing, and closed-loop closure
- lowers only to ordinary retained `VectorPath` geometry
- reuses Noon's existing Manim-v0.21 cubic smoothing implementation for the default `use_smoothing=True`
- direct `ImplicitFunction` evaluates arbitrary Python scalar fields synchronously only during authoring through a WASM callback; topology/path ownership stays Rust-side and no per-frame Python callback is retained
- `Axes.plot_implicit_curve` keeps contouring in coordinate space, then applies the pinned origin-space x/y unit stretch plus origin shift

## Architecture
`noon-geometry` owns the renderer-independent isoline algorithm. `noon-web` exposes a testable authoring plan plus a thin WASM callback bridge. `_manim_implicit.py` is only facade/range/style/Axes composition; it does not own contour topology.

No fixed-grid marching-squares substitute, renderer special case, reactive rebuild, or new serialized geometry kind is introduced.

## Acceptance
Focused Rust fixtures cover min-depth precedence, deterministic zero curves, closed-loop smoothing, NaN refinement, and asymptote rejection. A dedicated real Pyodide/WASM browser smoke covers direct smooth/unsmoothed circles, NaN-defined boundaries, shifted/scaled Axes mapping, callback exception/non-scalar propagation, ordinary retained VectorPath serialization, zero tracks, and no callback activity after Scene.add.

Tracks #856 / #85. Independent base: exact-green PolarPlane #828 (`85ebaa57849a39e3bbd42409a13aa405947fac49`).


yongkyuns:
Superseded by the Phase A architecture consolidation in #953 / merged #952. This implementation is built on the pre-consolidation retained authoring/scene bridge; preserve it as specification/evidence, but replay the useful contouring semantics after the authoritative Semantic Scene and lowering boundary exist.

PR #882, opened 2026-09-02
Make Rust own per-scene canonical authoring context

yongkyuns:
Closes the next #367 production boundary after #858.

This introduces one Rust/WASM `CanonicalAuthoringSceneContext` per Python Scene. Geometry and retained Text bind events append directly into its authoritative `ObjectSpec` vector; static updates preserve their slot, and the existing authoring checkpoint path now rolls back canonical binds with legacy compatibility state.

`Scene.to_scene_spec()` no longer reads `to_document()` or `retained_document()`. It finalizes the Rust-owned object stream directly with authoritative geometry tracks, retained tracks, canonical family requests, and camera identity. The legacy geometry document and retained sidecar remain transport/compatibility views only.

The old `canonicalRetainedSceneSpecJson` adapter remains exported for compatibility while callers migrate.

Validation added in Rust covers mixed geometry/Text ordering, update/rollback slot stability, direct camera ownership, and content-domain rebinding rejection.


yongkyuns:
Rebased this implementation onto current `master` (including merged #884 and #901) and marked the PR ready for review.

Current validation:
- `cargo fmt --all -- --check` and `git diff --check`
- `cargo test -p noon-web` — 233 passed
- `NOON_WEB_PREFLIGHT_ONLY=1 NOON_SKIP_PLAYGROUND_TEST=1 bash scripts/build-web-demo.sh` — passed; all 171 Python tests
- `node --test web/authoring-scene-spec-result.test.mjs web/authoring-client.test.mjs web/playground-scene-spec-routing.test.mjs web/retained-execution-topology.test.mjs` — 26 passed
- WASM package build — passed
- `node scripts/manim-typst-authoring-smoke.mjs` — passed
- `node scripts/manim-compat-smoke.mjs` — passed

The branch remains scoped to the Rust-owned per-scene canonical `SceneSpec` context; legacy geometry and retained sidecar outputs remain compatibility projections while later migration slices retire them.

web/python-worker.source.js
```diff
@@ -48,6 +49,8 @@ async function initializePyodide() {
   ]);
   const [, pyodide, compatibilityModules] = await startupResourcesReady;
   const authoringStore = new WasmAuthoringStore();
+  self.noonCreateCanonicalAuthoringSceneContext = () =>
```
yongkyuns, line 52: **Blocking:** the exact-head Fast Gate currently fails in the browser lane with `Python authoring worker crashed`; the later Manim/canonical/WebGPU checks are then skipped. This PR changes this worker bootstrap directly by adding the canonical context constructor and installing `_manim_canonical_scene`, so I would require a green exact-head browser run before merge. Please reproduce/capture the underlying worker initialization exception (rather than only the generic client error) and fix or otherwise demonstrate this path is stable.

yongkyuns (commented):
The ownership direction looks right and matches the next #367 boundary: one Rust-owned mixed object stream, stable slots for updates, and direct SceneSpec finalization. I don't see a core data-model correctness blocker.

**Merge recommendation: do not merge yet.** The exact-head PR Fast Gate is currently red in the browser lane. Rust unit tests, lint, and web static/unit checks pass, but deterministic playback ends with a Python authoring worker crash and the remaining browser integration checks are skipped. Since this PR changes worker bootstrap and installs the new canonical scene adapter, I would require a green exact-head browser run before merge.

The JSON mutation boundary and append-only checkpoint comments below are architectural concerns/follow-ups; I would not independently block the PR on them if they are intentionally tracked as migration seams.

web/python/_manim_canonical_scene.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def _append_snapshot(
+    self: _ir.Scene,
+    snapshot: dict[str, Any],
+    key: str | None,
+) -> _ir.Object:
+    checkpoint = _ORIGINAL_AUTHORING_CHECKPOINT(self)
+    obj = _ORIGINAL_APPEND_SNAPSHOT(self, snapshot, key)
+    try:
```
yongkyuns, line 65: **Architecture / non-blocking if explicitly tracked:** this makes Rust authoritative for object ownership, but JSON is still the live mutation API (`dict -> json.dumps -> JS/WASM -> serde`) for every geometry bind/update, and finalization similarly serializes all track/family state back into Rust. #367's target says JSON should be transport/debug/export rather than the normal mutation/query API. Can we pass existing typed WASM/semantic handles here instead? If that is intentionally deferred, please document this as a remaining migration seam rather than treating this as the final producer boundary.

crates/noon-web/src/canonical_authoring_scene.rs
```diff
@@ -0,0 +1,398 @@
...
+        let position = self.position(id)?;
+        if !matches!(&self.objects[position].content, ObjectSpecContent::Text(_)) {
+            return Err(format!("canonical object {} is not text-backed", id.get()));
+        }
+        let scale_factor = retained_scale_factor(&text);
+        self.objects[position] = canonical_text_object(id, text)?;
+        self.retained_scale_factors.insert(id, scale_factor);
+        Ok(())
+    }
+
```
yongkyuns, line 78: **Robustness / non-blocking:** this checkpoint only captures `objects.len()`, and `restore()` only pops objects appended after it. Any `update_geometry` / `update_text` to an already-bound slot after the checkpoint survives rollback. Today I don't see a clear failing call path that mutates an existing canonical slot inside a transaction, so I wouldn't block on it, but the API name suggests stronger transactional semantics. Please either make the checkpoint restore updates too, or narrow/document the append-only contract and add a regression test for the intended behavior.

yongkyuns:
Addressed the review comments in `ab8bdd3b`:

- Browser worker failures now retain the worker error stack and source location; initialization errors sent by the worker also carry diagnostic stack data. Added a client regression test.
- Documented the JSON bind/update/finalize calls as an intentional, temporary #367 migration seam. Typed semantic/WASM handles remain the follow-up transport boundary; JSON is not used in the frame loop.
- Documented checkpoint/restore as append-only and added coverage proving existing-slot updates survive rollback while failed appended binds are reclaimed.

Validation on the exact PR browser profile (`wasm-pack@0.15.0`, dev, `--no-opt`):
- `cargo fmt --all -- --check` and `git diff --check`
- `cargo test -p noon-web` — 233 passed plus integration/doc tests
- `node --test web/authoring-client.test.mjs web/python-worker-source.test.mjs` — 22 passed
- deterministic replay — 11 scenes passed
- Manim compatibility smoke — passed
- WebGPU browser smoke — passed

Issue #697, opened 2026-08-31
[Roadmap B5] Retained Graph topology with local incident-edge updates


yongkyuns:
## Roadmap

Phase B5 child of #87/#954. Architecture authority: `docs/architecture.md`.

## Goal

Land explicit-position retained Graph/DiGraph topology with stable identity and local **effective** incident-edge updates.

## Scope

- stable vertex identity adapted from host/user keys into semantic handles;
- stable edge identity from endpoints + directedness;
- Graph/DiGraph construction with explicit authored positions;
- ordinary semantic vertex mobjects and Line/Arrow-derived edge mobjects;
- persistent add/remove vertex/edge mutations preserving unrelated identity;
- active vertex motion updates only the vertex's effective transform plus incident-edge effective dependency state;
- common vertex/edge configuration required by initial fixtures.

Automatic layout algorithms, broad labels/partition/root helpers and callable-layout compatibility remain later #87 work.

## Architecture

- Graph topology/family/dependency declarations are authored Semantic Scene state;
- lowering derives incident-edge runtime dependencies;
- timeline/reactive/host/editor vertex motion does not commit semantic edge geometry every frame;
- persistent topology/position edits use shared semantic mutation/local lowering;
- ordinary immutable geometry/resources remain retained;
- directed edges reuse B2/#77;
- frontend key mapping is an adapter, not topology authority.

## Acceptance

- topology construction/identity has focused tests;
- one effective vertex-motion update dirties only that vertex + incident dependencies, independent of unrelated graph/scene size;
- no semantic revision is required for pure active-driver vertex motion;
- persistent add/remove preserves unrelated semantic/execution/resource identity;
- directed/undirected ownership is deterministic;
- explicit-layout source-equivalent fixture passes B7/#185 qualification;
- no full-graph frame scan, Python edge-following updater or graph-specific renderer path.

## Dependencies

B1 family/identity/live-query state, B2 Line/Arrow geometry and Phase A/C mutation/effective-state/locality machinery. Later #87 work adds author/edit-time layouts and labels.

yongkyuns:
I opened #866 as the first architecture-complete #697 slice: shared Rust retained Graph topology with monotonic vertex/edge identity, deterministic insertion order, undirected edge-key normalization, and per-vertex incident-edge indexing so move-one-vertex dependency resolution is O(degree) without a graph scan. Focused tests cover locality, identity preservation under removal, duplicate semantics, self-edges, and invalid endpoints.

[...]

PR #884, opened 2026-09-02
Implement validated roadmap P0 and graph topology slices

yongkyuns:
## Summary
- Add WebGL2/WebGPU host-updater diagnostics for the RotationUpdater path, including actual `Queue::write_buffer` tracing and distinct committed/prepared/upload/draw-plan/present-call boundaries.
- Reconcile Manim v0.21 compatibility bookkeeping against current-master exports and executable evidence.
- Add a shared renderer-independent `Graph`/`DiGraph` topology identity core with transactional mutations, monotonic identities, incident-edge indexing, and tombstoned stable slots.

## Review follow-ups addressed
- Run the host-updater diagnostic harness in PR browser CI when relevant renderer/transport/host paths change.
- Clear stale deterministic-render diagnostics before the host phase.
- Use semantic browser assertions instead of fixed packing/byte-count assumptions.
- Trace actual CPU→GPU writes with buffer/range/offset/length/payload fingerprints.
- Use primitive instance indices, not global frame-object indices, for upload/draw diagnostic matching.
- Add a mixed-primitive regression fixture where the target is frame object 2 but line instance 1.
- Tighten static public-export detection to require actual `__all__` registration.
- Make graph removals local by tombstoning slots rather than globally reindexing vectors.

## Validation
Latest PR Fast Gate on head `ed76b370` is green, including:
- Rust formatting and workspace Clippy with `-D warnings`
- Rust workspace library tests
- Web static/unit preflight
- Deterministic browser playback
- Manim-style authoring smoke
- Canonical retained execution smoke
- Primary WebGPU rendering smoke
- WebGL2/WebGPU host-updater diagnostic harness using the mixed-primitive regression fixture

Additional validation performed while addressing review feedback included `cargo test -p noon-core --lib`, `cargo test -p noon-render-wgpu --lib`, `cargo test -p noon-web`, the Manim API coverage policy tests, and the Node reference/coverage tests.


yongkyuns (commented):
Architecture / scalability / correctness / validation review

Overall direction is good, especially the renderer-independent graph topology, stable monotonic identity, and the effort to make the WebGL/WebGPU updater path diagnosable. I think there are three issues that should be addressed before merge.

## Merge-blocking

### 1. The claimed GPU/presentation diagnostic boundaries are not independently observed

`DiagnosticGpuState.state` is reconstructed from the same CPU-side `PreparedFrame` values used for the prepared state (`prepared_transform`, `prepared_endpoints`, `line_dirty_ranges`). `bytes_uploaded` for the target is also inferred from dirty-range length × `size_of::<LineInstance>()`; only aggregate `UploadStats` comes from `renderer.upload()`.

Similarly, draw membership is inferred from `prepared.render_batches`, and presentation is recorded as `submitted: true` / `presented: true` after the corresponding calls.

That means a bug such as "correct PreparedFrame, wrong GPU buffer offset / wrong buffer / stale uploaded payload" can still produce a diagnostic saying the GPU state is correct, because the diagnostic rereads the prepared CPU instance rather than observing what was actually handed to or consumed by the GPU.

For this diagnostic to localize the WebGL/WebGPU divergence reliably, please either:

- instrument `GpuRenderer::upload()` at the actual CPU→GPU write boundary and record the target buffer/range/offset/payload (and use framebuffer/readback/screenshot evidence for presentation), or
- rename these states to something like `upload_plan`, `draw_plan`, and `present_call` so the diagnostic does not claim stronger observation than it has.

A small fault-injection test would be valuable here: deliberately corrupt an upload offset/payload in a test seam and prove this diagnostic catches it.

### 2. The new end-to-end diagnostic harness is not run by CI

The PR Fast Gate is green and already builds WASM, installs Playwright Chromium, and runs browser checks, but `scripts/manim-host-updater-diagnostics.mjs` is only syntax-checked via `node --check`; the actual WebGL2/WebGPU frame-90 test is not executed.

Please add `node scripts/manim-host-updater-diagnostics.mjs` to the browser CI path (potentially risk-gated on renderer/transport/host-diagnostic changes). Since the expensive WASM + Playwright setup is already paid for, this should be the authoritative validation for the feature introduced here.

### 3. `GraphTopology` lookup is local, but deletion is not scalable/local

The incident index correctly gives O(degree) lookup, but `remove_edge()` uses `Vec::remove(index)` and then reindexes all following edges. `remove_vertex()` calls that once for every incident edge and then also removes/reindexes the vertex vector.

So roughly:

- incident lookup: O(degree)
- remove one edge: O(E)
- remove a degree-d vertex: approximately O(d × E)

That conflicts with the architectural goal in #697 that graph mutations remain local and preserve unrelated semantic/execution identity/work.

Before this becomes the shared Graph/DiGraph substrate, I would strongly consider stable slots/tombstones (or another arena-like representation) so removal does not physically relocate/reindex unrelated topology. Deterministic insertion order can still be preserved by ordered live-slot iteration, with explicit compaction as a separate controlled operation.

## Other issues / follow-ups

### Potential stale host diagnostic

`advanceOneFrame()` renders the deterministic delta first, then the host delta, and only afterward calls `takeHostUpdaterDiagnosticJson()`. If the deterministic delta renders but the host patch produces no renderable delta, the previously stored diagnostic can be consumed as if it belonged to the host updater.

Please clear/take the old diagnostic before applying the host patch, then only consume a new diagnostic if the host delta actually rendered. Tagging diagnostics with transport/session/sequence/source would make this contract much safer.

### Browser assertions are over-specified in places

The test hardcodes internal representation details such as exactly `88` bytes, packed instance index `{1,2}`, and exact batch shape. Those are useful if they are intentional packing/performance contracts, but otherwise they make a semantic regression test fragile.

Also, each backend individually permits `surface_status` to be either `success` or `suboptimal`, but the final deep equality of the whole `presentation` object requires both backends to choose the same status. Two individually valid runs can therefore fail the cross-backend comparison.

I suggest separating semantic invariants (identity, transform/endpoints, dirty membership, target uploaded/drawn, final raster) from renderer packing/performance assertions.

### PR scope

This PR contains three mostly independent changes: compatibility bookkeeping, GraphTopology, and host-renderer diagnostics. Splitting these into separate PRs would make review, rollback, CI risk classification, and later bisection much easier. I would at least consider that before landing, though I do not consider it as important as the three issues above.

### Static export scanner can false-positive

The new AST rule treats any two-variable `for ... in {literal}.items()` as public exports. It works for the current `Write`/`Unwrite` registration, but a future internal mapping loop with the same syntax would be incorrectly reported as public API. Please consider checking that the loop body actually appends/registers the loop variable into `__all__`, and add a negative test.

## Graph tests I would add

- removing one vertex preserves every unrelated vertex/edge ID;
- remove + re-add of the same endpoints gets a new monotonic edge ID;
- failed removals are state-equivalent no-ops;
- self-loop removal keeps incident indexes consistent;
- dense/star mutation work-count or benchmark;
- once lowering exists, moving a degree-d vertex proves exactly d incident edge updates rather than O(E) work.

Overall: good architectural direction, but I recommend addressing the diagnostic-observation gap, executing the new harness in CI, and settling the retained topology deletion model before merging this as the shared foundation.

yongkyuns:
Addressed the review feedback in commit `53f70e36`:

- Added an opt-in `Queue::write_buffer` trace with buffer, instance range, byte offset/length, and payload fingerprint; diagnostics now distinguish upload writes, draw plans, and presentation calls.
- Added the WebGL/WebGPU host-updater diagnostics harness to the PR browser CI path behind the risk classifier.
- Cleared stale diagnostics before the host phase and only consumed diagnostics when the host delta actually rendered.
- Relaxed browser checks to assert semantic invariants instead of fixed byte counts or exact batch packing.
- Changed graph topology removal to tombstone slots, preserving unrelated IDs/indices and insertion order without global reindexing; added removal/self-loop/failed-operation tests.
- Tightened static export-loop detection so internal mapping loops are not counted as public exports.

Validation: `cargo test -p noon-core --lib` (210 passed), `cargo test -p noon-render-wgpu --lib` (83 passed), `cargo test -p noon-web` (passed), WASM check (passed), Python policy tests (4 passed), Node tests (23 passed), and the actual WebGL2/WebGPU host-updater harness (passed).

Local `cargo clippy -p noon-web --all-targets -- -D warnings` still reports the pre-existing `clippy::collapsible_match` at `crates/noon-web/src/execution_transport.rs:762`, which is outside this PR.

yongkyuns (commented):
Follow-up review after commit `53f70e36`

This update resolves nearly all of my previous findings. In particular:

- the upload diagnostic is now instrumented at the actual `Queue::write_buffer` boundary;
- the diagnostic terminology correctly distinguishes upload writes, draw plans, and presentation calls;
- the stale host-diagnostic issue is fixed;
- the WebGL/WebGPU host-updater harness is now executed in CI and passes;
- graph deletion now uses tombstoned stable slots rather than shifting/reindexing unrelated entries;
- the static export scanner is constrained by actual `__all__` registration and has a negative test.

I have one remaining correctness issue that I recommend fixing before merge.

## Remaining merge-blocking issue: diagnostic mixes frame indices with renderer instance indices

`build_host_updater_diagnostic()` computes `frame_index` as the target object's position in `FrameState.objects`, then uses that value to identify the target GPU write:

```rust
write.instance_range.contains(&frame_index)
```

Those are different index spaces. `UploadWrite.instance_range` is an index into the primitive-specific GPU instance buffer (`prepared.lines` here), not the global frame object array.

The browser test currently has the same accidental assumption when it checks whether the line instance range contains `committed.frame_index`, and later when it uses `targetObject` to test line draw-batch membership.

The current RotationUpdater fixture happens to contain two Line objects in matching order/IDs, so all three values collapse to the same number and CI passes. A mixed-primitive scene can break this immediately, e.g.:

```text
Circle(id=0)
Line(id=1)   <- diagnostic target
```

Here `frame_index == 1`, while the target's line-instance index is `0`. The diagnostic could then claim the target was not uploaded/drawn even though it was.

`prepared_line_state()` already discovers the correct line-instance index. I recommend returning that index explicitly and using it for upload-range and draw-range matching. Please also add a mixed-primitive regression fixture so frame/object/primitive-instance index domains cannot accidentally collapse again.

## Non-blocking follow-up: tombstone churn

The graph topology fix removes the important global O(E) reindexing problem, so I no longer consider this a merge blocker. There are two scalability details worth tracking separately:

- `vertices()` / `edges()` scan historical tombstone slots, so iteration cost and memory grow with total historical additions rather than live graph size;
- `remove_incident_reference()` still performs a linear search plus `Vec::remove`, so repeated removal around very high-degree vertices can become quadratic in degree.

A future compaction/free-slot policy and/or deletion-friendly incident structure would address those if large, highly mutable graphs become important.

## PR description

The PR description is now stale: it still says the frame-90 browser harness was not run and only lists `node --check`, while the current PR Fast Gate actually runs the WebGL/WebGPU diagnostic harness successfully. Please update that before merge.

Overall, this is now very close. Once the renderer-instance-index vs frame-index bug is fixed and covered by a mixed-primitive regression test, I would be comfortable merging; the remaining graph churn concern can be handled as follow-up work.

yongkyuns:
Addressed the follow-up review in commit `ed76b370`:

- Added an explicit primitive `instance_index` to the prepared diagnostic state. Upload-range and draw-plan matching now use the resolved line-buffer index, never the global `FrameState.objects` index.
- Added `web/python/examples/manim_host_updater_mixed_primitive.py`, where the target is global object 2 but line instance 1, and updated the end-to-end harness to run this fixture.
- Relaxed the browser assertions to validate the primitive instance domain explicitly.
- Updated the PR description to remove the stale limitation and record the actual WebGL2/WebGPU harness run.

Validation: `cargo test -p noon-web` (225 unit tests plus integration/doc tests), WASM build/package, and `node scripts/manim-host-updater-diagnostics.mjs` all pass. The mixed-primitive harness passes on both WebGL2 and WebGPU.

Issue #109, opened 2026-08-24
[Roadmap A6/B7] Visual golden and backend parity regression suite


yongkyuns:
## Roadmap

A6/B7 verification infrastructure under #953/#954. Architecture authority: `docs/architecture.md`.

## Goal

Provide a compact deterministic visual-golden/backend-parity suite for raster-visible behavior that semantic/effective-state assertions cannot fully prove, while ensuring screenshots are tied to the **same coherent engine publication context** rather than to loosely corresponding wall-clock frames.

## Evidence model

Raster evidence supplements, but does not replace, authored/effective semantic evidence.

For each canonical fixture record enough context to identify what was rendered:
- source fixture/version and backend;
- viewport/DPR/background/resource configuration;
- exact authored/simulation time or logical segment boundary;
- relevant `SceneRevision` / `ExecutionRevision` where exposed;
- sampled `FrameEpoch` or equivalent coherent effective publication context;
- relevant immutable `ResourceVersion`s when a resource transition is under test.

Different backends do not need numerically identical internal IDs/counters, but they must render equivalent authored/effective state for the same fixture/time/context.

### Canonical renderer fixtures
Cover representative:
- analytic/path geometry;
- stroke/fill/opacity/antialias edges;
- authored vs active effective transforms, including non-uniform scale/rotation;
- partial-path/reveal boundaries and intermediate effective states;
- morph/cross-geometry transforms, including transient Runtime-provided render geometry;
- mixed painter order/transparency;
- camera movement and clipping/canvas edges;
- small/thin geometry;
- representative text/image/resource replacement cases;
- representative 3D/depth cases as those capabilities land.

### Comparison
- sample at exact authored times/logical boundaries rather than “frame N at approximately the same wall time”;
- use reviewed tolerant/perceptual image diff rather than exact PNG bytes where rasterization permits harmless variation;
- semantic/numeric assertions remain primary for state that can be checked directly;
- current-value assertions use coherent effective Runtime state, not stale authored/base declarations;
- supported backends compare against the same observable expectation and one another within explicit tolerance;
- failures retain expected/actual/diff images plus authored time, publication context and backend/resource metadata.

### Segment/activation boundaries
For animation fixtures:
- sample before activation, immediately after activation where relevant, representative intermediate times, endpoint and post-lifecycle state;
- activation-time current-state capture must agree with #63/#82 semantics;
[...]

PR #899, opened 2026-09-02
Gate canonical Manim raster differential before merge

yongkyuns:
Clean current-master replay of #865 after the palette export regression was fixed. Runs the unchanged canonical raster differential on PRs targeting master and ratchets push/PR trigger parity.

Supersedes #865.

### DIFF
--- a/.github/workflows/manim-raster-differential.yml
+++ b/.github/workflows/manim-raster-differential.yml
@@ -6,6 +6,11 @@ on:
       - master
     paths-ignore:
       - "crates/**/tests/**"
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - "crates/**/tests/**"
   workflow_dispatch:
 
 permissions:
@@ -29,6 +34,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Validate raster workflow trigger policy
+        run: node scripts/manim-raster-workflow-policy.test.mjs
       - name: Validate raster ratchet policy
         run: node scripts/manim-raster-policy.test.mjs
       - name: Set up Python
--- a/.github/workflows/manim-raster-differential.yml
+++ b/.github/workflows/manim-raster-differential.yml
@@ -9,8 +9,6 @@ on:
   pull_request:
     branches:
       - master
-    paths-ignore:
-      - "crates/**/tests/**"
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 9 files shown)


yongkyuns:
I would request changes before merging this.

Two blockers:

1. **`paths-ignore` is unsafe for a required pre-merge gate.**

The PR copies the existing push filter onto `pull_request`:

```yaml
pull_request:
  branches:
    - master
  paths-ignore:
    - "crates/**/tests/**"
```

For a required check, this is problematic: if the whole workflow is skipped by a path filter, the required check may never report success and the PR can remain blocked waiting on a check that never ran. The new policy test also locks this coupling in by requiring push and PR path filters to stay identical.

I think the safer structure is:
- always trigger the workflow for PRs targeting `master`;
- classify changed paths inside the workflow;
- skip only the expensive raster job when it is not needed;
- keep an always-running gate job that reports success when raster validation is unnecessary.

The post-merge push policy and pre-merge required-check policy do not need to be identical.

2. **The newly introduced PR raster gate is currently red on this PR.**

The trigger-policy test passes, the existing raster-policy test passes, the browser package builds, and the retained Typst checks pass. The failure is later in `Render and compare canonical Manim scenes`, specifically:

```text
mixed-object-parity-stress:
logical Manim duration 5.333333333333324 != expected 5.0
```

Before making this oracle a merge gate, the current baseline should be green (or this failure should be explicitly resolved), otherwise enabling it as required CI will block unrelated development immediately.

One additional issue: the trigger-policy regression test currently runs *inside the workflow whose trigger it validates*. If somebody later breaks/removes the `pull_request` trigger, the test stops running too. I would move that inexpensive policy check into the always-running PR Fast Gate/static CI path, or at least enforce it there as well.

The core direction is good: moving the canonical raster oracle from post-merge discovery to pre-merge enforcement is the right idea. I just would not merge the current CI structure as-is.

yongkyuns:
Roadmap triage: **KEEP / merge independently when green.** Pre-merge canonical Manim raster gating is aligned with Phase A6 #953 and continuous Phase B7 qualification #954. This strengthens the new architecture rather than preserving an old scene model.

yongkyuns:
Addressed the review feedback and pushed the fixes.\n\n- PRs now always trigger the raster workflow; an in-workflow scope classifier skips only the expensive raster job when appropriate, while the always-running required gate reports the result.\n- Moved workflow-policy validation into the always-on PR Fast Gate.\n- Frame-aligned the mixed-object stress fixture so ManimCE 0.21.0 reports the manifest’s 5.0s duration.\n- Added bounded floating-point duration checks to semantic capture and seek/playback validation.\n- Large-fixture raster/seek harness loads use full scene replacement to avoid incremental reconciliation failures.\n\nLocal focused checks pass, and all required PR checks are green. The PR is still awaiting reviewer approval.

yongkyuns:
Closing this replay under the clarified architecture. The raster/semantic qualification goal remains valid under #109/#185, but this PR makes the current `scene_document` + `JSON.stringify(...)` + `loadScene(...)` bridge part of the required canonical PR gate. Phase A3/A4 explicitly removes that authoring/transport path. Rebuild the gate after the typed semantic-handle path is authoritative; test/debug snapshots may still serialize explicitly, but CI must not require the migration SceneDocument authoring bridge to exist.

Issue #69, opened 2026-08-24
[Roadmap C2] Native browser input/event signals


yongkyuns:
## Roadmap

Phase C2 child of #955. Architecture authority: `docs/architecture.md`.

## Goal

Make pointer, keyboard, viewport, controls, wheel/gesture and other ordinary inputs first-class native runtime sources without requiring Python on the frame path.

## Ownership

This issue owns normalized input/event ingress and delivery into runtime/native-reactive state. #846 owns editor/session concepts such as selection, capture and manipulation. #70 owns arbitrary host callbacks and persistent semantic mutation initiated from them.

## Input model

Separate two classes:

- **sampled latest-value state** — pointer position/buttons, key-down state, viewport/control values; may coalesce to the newest value for the same source;
- **ordered discrete events** — press/release/click/key events and other occurrences whose order is observable; must not silently coalesce or disappear.

Each ingress source carries stable source identity plus explicit `InputSequence`/revision context.

### C2.1 — normalized ingress
- normalize browser/native input into one backend-neutral model;
- reject malformed/non-finite values at the boundary;
- preserve equivalent semantics across native and web hosts.

### C2.2 — bounded queues/backpressure
- sampled state may overwrite older unread state for the same exact source;
- discrete events use bounded ordered storage;
- overflow behavior is explicit and observable: apply backpressure, surface a session/input fault, or another documented policy; never silently drop required ordered events.

### C2.3 — runtime/reactive binding
Inputs feed the same native reactive/runtime driver machinery as other signals. Only the affected dependency closure becomes dirty. Thin Rust/Python handles expose authoring declarations without owning execution semantics.

### C2.4 — paused interaction and wake
Input may wake a clean paused scene and update effective reactive/session state while authored timeline `dt == 0`. A clean scene should otherwise remain settled rather than busy-ticking for possible input.

### C2.5 — hit-test integration
Consume the retained execution-owned spatial query path from #569. Do not scan frontend objects or add another spatial index.

### C2.6 — replay classification
Ordered input/events carry enough identity/sequence/timing information for deterministic replay where supported. Wall-timed/external input is explicitly classified when replay is not deterministic.

## Acceptance

- pointer/keyboard/control examples run with zero Python callback phases after authoring when behavior is native;
- sampled state coalesces deterministically by source;
- discrete event order is preserved with explicit bounded-storage policy;
- paused input wakes the scene without advancing authored time;
- native input dirties only affected dependencies/candidates;
[...]

PR #925, opened 2026-09-02
Batch native input bursts without losing discrete events

yongkyuns:
## Summary

Advance #69 with a renderer/browser-agnostic pending-input batch in `noon-core`.

- coalesce sampled state by exact `NativeStateSource`, retaining only the newest pending sample
- preserve every discrete `NativeEventOccurrence` in ingress order
- expose batch-local received/coalesced counters for later runtime instrumentation
- keep timeline/reactive/commit phase ordering outside this container so the runtime dispatcher remains the single owner of execution semantics
- bound sampled-state memory by active source count instead of browser sample count, without imposing an unsafe drop policy on discrete events

## Tests

Focused Rust coverage exercises a 10,000-sample pointer burst, exact-source separation, 4,096 alternating pointer down/up events with no loss or reordering, and batch reset/accounting.

## Scope / remaining risk

This deliberately does not choose cross-phase dispatch ordering or a discrete-event overflow/drop policy: #69 explicitly requires discrete occurrences not to disappear. Browser normalization, worker transport, application into the reactive graph, and paused-scene wakeup remain follow-up slices.

Closes no issue; contributes to #69.

### DIFF
--- a/crates/noon-core/src/reactive/native_input_runtime.rs
+++ b/crates/noon-core/src/reactive/native_input_runtime.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeMap;
+
 use serde::{Deserialize, Serialize};
 
 use crate::{NativeEventSource, NativeStateSource, ValueKind, Vec2};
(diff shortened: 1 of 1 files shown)


yongkyuns:
Roadmap triage after #952: closing this draft and replaying the useful input-batching idea in Phase C2 (#955 / #69) after Phase A #953. The sampled-vs-discrete contract remains valid, but it should land against the consolidated runtime/module ownership rather than extend the pre-reset core layout.

yongkyuns:
Superseded for now by Phase A (#953). The sampled-vs-discrete batching semantics and tests remain useful design evidence for Phase C/#69, but this implementation targets the pre-consolidation core/runtime shape. Replay it after the authoritative Semantic Scene/mutation/lowering boundaries are in place.

PR #936, opened 2026-09-02
Preserve retained text locality in mixed renderer

yongkyuns:
## Summary

- preserve parent retained scratch and mixed painter order for compatible text-only object changes
- retain per-object text item ranges and copy only changed glyph quad ranges into the parent snapshot
- route those ranges through the existing generation-safe GPU range uploader
- include the retained family reveal scene index prerequisite for O(1) active-family object lookup
- keep structural, metric, resource, lane-transition, and family-local realization changes on the conservative full-rebuild paths

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p noon-render-wgpu --all-targets -- -D warnings`
- `cargo test --workspace`

The focused 10,000-object regression covers 128 one-object translation/opacity updates and asserts smaller-than-initial GPU text uploads, one parent scratch rebuild, one text snapshot copy, and one mixed-order rebuild.

Related: #362 #835

### DIFF
--- a/crates/noon-render-wgpu/src/gpu/retained_text.rs
+++ b/crates/noon-render-wgpu/src/gpu/retained_text.rs
@@ -69,12 +69,15 @@ pub struct RetainedPrepareStats {
 /// Cumulative counters for retained mixed-frame preparation locality.
 ///
 /// A scratch reuse means the semantic object/text/vector/outline walk was skipped.
-/// Mixed-order rebuilding and text snapshot copies remain separate follow-up costs
-/// and are intentionally not hidden by this counter.
+/// Text snapshot copies and mixed-order rebuilds are counted separately so a
+/// transform/opacity-only text update can prove that both parent-level scans stay
+/// out of the frame loop.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub struct RetainedFrameIncrementalStats {
     pub scratch_rebuilds: u64,
     pub scratch_reuses: u64,
+    pub text_snapshot_copies: u64,
+    pub mixed_order_rebuilds: u64,
 }
 
 pub struct PreparedRetainedTextSnapshot<'a> {
@@ -84,6 +87,9 @@ pub struct PreparedRetainedTextSnapshot<'a> {
     pub items: &'a [PreparedTextItem],
     pub stats: RetainedTextPrepareStats,
     pub atlas: &'a GpuGlyphAtlas,
+    pub partial_upload: bool,
+    pub dirty_mask_ranges: &'a [std::ops::Range<u32>],
+    pub dirty_color_ranges: &'a [std::ops::Range<u32>],
 }
 
 impl PreparedRetainedTextSnapshot<'_> {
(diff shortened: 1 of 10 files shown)


crates/noon-render-wgpu/src/gpu/retained_text.rs
```diff
@@ -1378,10 +1561,20 @@ impl GpuRenderer {
             prepared.text_generation,
         ) {
             let text_frame = prepared.text.as_prepared_frame();
-            let uploaded =
+            let uploaded = if prepared.text.partial_upload {
```
yongkyuns: **P1 / correctness:** `upload_ranges()` is only safe when this GPU state already contains the *immediately preceding* prepared generation. Here we only check that `last_uploaded_generation != prepared.text_generation`, so a skipped upload can leave stale ranges on the GPU. Example: GPU has gen 1; prepare gen 2 changing A but don't upload it; prepare gen 3 changing B; uploading gen 3 writes only B's dirty ranges, leaving A at gen-1 data. Please only use the partial path when `last_uploaded_generation` matches the partial update's base generation (e.g. the previous generation); otherwise fall back to a full `upload()`. An explicit `partial_upload_base_generation` would make this contract robust. Please also add a regression test that prepares two local generations while skipping the first upload.

crates/noon-render-wgpu/src/gpu/retained_text.rs
```diff
@@ -793,6 +880,28 @@ impl RetainedFramePreparer {
...
+            return false;
+        }
+        changes.object_indices().iter().all(|&index| {
+            let Some(object) = frame.objects.get(index) else {
+                return false;
+            };
+            if !frame.is_present(index) || !matches!(&object.content, ObjectContentRef::Text(_)) {
+                return false;
+            }
+            self.sources.iter().filter(|source| match source {
```
yongkyuns: **P2 / scalability:** this still scans all of `self.sources` for every changed object, so the supposedly local eligibility check is O(K·N) for K changed objects in an N-source scene. In the 10k-object / one-object animation case, we still walk roughly the whole retained source list each frame. Could we cache per-object eligibility/source metadata while building scratch (e.g. `fast_text_only: Vec<bool>` or per-object source ranges), making this check O(K)? That would better match the PR's goal of removing scene-wide work from the frame loop.

yongkyuns (commented):
Two issues from the retained-text locality review. The first is a correctness issue that should be fixed before merge; the second leaves an O(scene-size) scan in the hot local-update path.

crates/noon-render-wgpu/src/gpu/retained_text.rs
yongkyuns: Fixed in 95421711. Partial snapshots now carry `partial_upload_base_generation`; the range uploader is used only when the GPU state has that exact base generation. If an intermediate local generation was prepared but skipped, the next upload falls back to a full upload. The retained 10k-object regression now explicitly prepares one local generation without uploading it, then verifies the following update performs a full upload.

crates/noon-render-wgpu/src/gpu/retained_text.rs
yongkyuns: Fixed in 95421711. Scratch construction now records `fast_text_only` per retained object slot. The local eligibility check uses that cached metadata in O(K) over changed objects and no longer scans `self.sources` for each update.

PR #948, opened 2026-09-02
Ratchet Examples browser WebKit paint workaround

yongkyuns:
Adds a focused regression test for the scoped Examples-browser WebKit paint workaround that already exists on master.

The test requires `.example-browser-layer .example-card` to keep `content-visibility: visible` and `contain-intrinsic-size: none`, and requires that eager-paint override to occur exactly once so gallery virtualization is not disabled globally.

This is test-only: no production CSS or browser behavior changes. Replayed onto current master after #947.

### DIFF
--- a/web/example-browser-webkit-paint.test.mjs
+++ b/web/example-browser-webkit-paint.test.mjs
@@ -1,0 +1,37 @@
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+import test from "node:test";
+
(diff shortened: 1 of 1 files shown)


yongkyuns:
Roadmap triage: **KEEP / merge independently when green.** This is a narrowly scoped Examples-browser regression ratchet with no scene/runtime/authoring architecture coupling. It can land during Phase A.

Issue #957, opened 2026-09-02
[Roadmap A1] Build the authoritative mutable Semantic Scene


yongkyuns:
## Roadmap

Phase A1 child of #953. Architecture authority: `docs/architecture.md`.

## Goal

Own the one authoritative authored Semantic Scene and the one typed Semantic Scene -> execution lowering boundary. Do not preserve flat scene documents, frontend snapshots or retained sidecars as peer authorities.

## Core contract

The Semantic Scene owns **authored/base state**:
- generational semantic identity;
- scene/family membership and ordering;
- content/resources and high-precision transform/style/presentation;
- lifecycle/source identity;
- animation, signal/binding and host/updater declarations;
- persistent authored mutations.

The Runtime owns **effective state at a published frame epoch**. Timeline/reactive/host/editor driver output normally does not mutate the Semantic Scene or require relowering. Persistent changes reconcile through the semantic mutation contract.

## Landed A1 foundation

A1.1-A1.5 have established:
- one scene-global generational semantic identity allocator;
- one store-owned semantic object/content/value model;
- local ordered root/family/lifecycle operations;
- semantic animation/reactive declarations;
- one validated atomic `SemanticMutationTransaction` vocabulary with local impact classification.

Historical implementation details and exact merge evidence remain in this issue's progress comments rather than the roadmap body.

## Transaction hardening — implemented

Structural callback/live-edit batches must be able to create a semantic identity and refer to it later in the **same** transaction.

Use transaction-local provisional references/tokens that resolve only on successful commit. Do not:
- reserve permanent semantic IDs before commit;
- expose fake/stale `SemanticNodeId`s;
- introduce a draft/temporary peer scene.

This is part of the one semantic transaction contract and is a prerequisite for the full C1 structural-callback path in #70.

## A1.6 — one explicit lowering/publication boundary

Target:

```text
SemanticStore
   -> analyze/lower
   -> existing compiled execution representation
   -> existing Runtime
```

Rules:
- one compiler-owned typed entry point; no JSON/transport for in-process Rust/WASM;
- preserve existing stable/tombstoned execution identity and local runtime machinery;
- mixed content/resource handles remain typed;
- native-reactive and animation lowering use canonical compiler-owned paths rather than parallel lowerers;
- lowering/preparation failure cannot publish a partial semantic/execution mapping;
- activation-time animation semantics may read current effective runtime state where required;
- local authored mutations update only affected execution/runtime/resource state;
[...]

Issue #953, opened 2026-09-02
[Roadmap A] Architecture consolidation — one Semantic Scene, one lowering boundary


yongkyuns:
## Authority

Phase A execution tracker under the single authoritative `docs/architecture.md`.

Phase A consolidates Noon onto one authored Semantic Scene, one typed lowering path, one Runtime and one Renderer before broad feature expansion. Child issues own implementation detail; this umbrella owns only cross-track invariants and the Phase A exit gate.

## Target

```text
Python facade       idiomatic Rust API
       \                 /
        shared semantic operations
                  |
           Semantic Scene
          authored/base truth
                  |
          analyze / lower
                  |
          Execution Plan
                  |
              Runtime
       effective live state
                  |
             Renderer
            /          \
      native host   web/WASM host
```

Normal in-process Rust/WASM engine boundaries are typed Rust calls. Serialization is only for explicit codecs or genuine cross-context transport.

### One authority per layer
- Semantic Scene owns authored/base identity, structure, content, declarations and persistent mutations.
- Execution/lowering owns derived specialization, never authored truth.
- Runtime owns effective state at a coherent published `FrameEpoch`.
- Renderer owns retained GPU projection only.
- Platform hosts own lifecycle/surface/event-loop integration only.

### Host execution and live plan revisions
As clarified in #1152, Rust compiled control flow, Python interpreter continuations and editor commands share semantic operations and completion/publication behavior; they do not need identical source execution machinery. The Semantic Scene remains live across derived execution revisions.

Plan changes must use atomic, impact-local publication and preserve compatible runtime state/resources. Rust source hot reload is distinct from semantic mutation. Optional host-expression capture belongs to later, measurement-driven #955 C1 work and is not a Phase A exit prerequisite.

Qualification must cover post-barrier mutations/new membership through the same live session and actual realtime continuation startup, alongside deterministic explicit-time proofs. No Rust interpreter, frontend scene engine or whole-program compilation layer is required.

### Authored vs effective state
Timeline, native-reactive, host-updater and editor-driver output normally changes effective Runtime state without changing authored/base state or forcing semantic relowering. Persistent results reconcile through the shared semantic mutation contract only when product semantics require it.

Normal live getters observe coherent effective state. Authored/base inspection, if exposed, is explicit and separate.

### Coherent publication
[...]

PR #962, opened 2026-09-02
Establish ordered semantic scene membership and lifecycle

yongkyuns:
## Roadmap ownership

Phase A / #953, A1.1 in #957. Migration seam deletion owner: #959/A4.

## What this implements

The stale branch was reset onto current `master` and reworked against the A1.1 implementation/review/test contract in #957.

This establishes one generational authored identity space plus authoritative ordered scene/family membership:

- new semantic nodes start detached;
- attaching appends the existing `SemanticNodeId` to authored scene order;
- detaching preserves semantic identity, source identity and family/alias edges;
- re-attaching preserves identity and appends according to current authored order;
- scene membership and order are one private intrusive storage representation; public `SemanticNodeResidency` exposes only `Detached | SceneOwned`, so previous/next storage topology is not frontend/serialization API;
- ordered family membership uses private identity-indexed links, preserving deterministic order while making duplicate lookup, append and member removal local;
- a member may belong to multiple families without transform ownership;
- deletion removes attached root membership locally, cleans only direct family/source/legacy mappings, invalidates the generation and never renumbers unrelated nodes;
- stale generations are rejected for read/attach/detach/delete;
- source identity survives detach/re-add and is released on deletion;
- family membership remains independent from scene residency at the storage primitive; A1.3 owns user-visible recursive scene/group operations.

## Temporary migration seam

`SemanticNodeKind::Object(ObjectDefinition)`, legacy `ObjectId -> SemanticNodeId` lookup, and `SemanticStore::from_scene_definition()` remain compatibility-only while A2/A3 migrate callers. The adapter imports current flat objects into semantic scene order only to keep the repository coherent. These paths are explicitly owned for deletion by #959/A4 and are not target architecture or a format-compatibility commitment.

No JSON/serialization/transport boundary is introduced into the Rust engine path. Deriving `Serialize`/`Deserialize` on semantic value types does not make serialization an internal engine boundary.

## Locality

- `NodeId` lookup: O(1);
- root attach: O(1), writes node + former tail at most;
- root detach: O(1), writes node + immediate previous/next at most;
- ordered root iteration: O(number of attached roots);
- direct family duplicate lookup/append/remove: expected O(1) identity lookup plus constant link rewrites;
- deleting a node touches its immediate root links plus its direct parent/member relationships; unrelated family siblings are not scanned;
- cycle validation is proportional to the affected reachable family subgraph, not hidden mutation work.

## Validation added

Focused tests cover:
- authored scene order A/B/C -> detach B -> A/C -> reattach B -> A/C/B;
- idempotent attach/detach;
- identity/source/family preservation across detach/re-add;
- 100k-node unrelated identity stability;
- bounded attach/detach mutation counts with 10k unrelated nodes;
- 10k-member family removal and node deletion without sibling scans;
- deterministic family order after remove/re-add;
- stale generation rejection for lifecycle operations;
- deletion repair of ordered scene links;
- family alias cleanup, family deletion cleanup and cycle behavior;
- source identity release/reuse;
- compatibility import preserving authored order while the adapter exists.

## Independent review result

Architecture/correctness/locality review found and fixed two issues during the PR:
1. root previous/next links were moved out of public `SemanticNodeResidency` into private storage so topology cannot become serialized/frontend API;
2. the pre-existing family `Vec` removal path was replaced with ordered identity-indexed links so A1.1 deletion/member locality no longer scans unrelated siblings.

Merge only after the final exact head passes architecture ratchet, format/clippy, Rust unit tests and the normal browser/web PR gate.

Refs #953 #957 #959.

### DIFF
--- a/crates/noon-core/src/semantic_family.rs
+++ b/crates/noon-core/src/semantic_family.rs
@@ -30,7 +30,7 @@ impl SemanticStore {
                 }
                 SemanticNodeKind::Family => {
                     for member in node.members() {
-                        collect(store, *member, seen, leaves)?;
+                        collect(store, member, seen, leaves)?;
                     }
                 }
             }
--- a/crates/noon-core/src/semantic_store.rs
+++ b/crates/noon-core/src/semantic_store.rs
@@ -68,6 +68,97 @@ enum SemanticSceneMembership {
     },
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct FamilyMemberLink {
+    previous: Option<SemanticNodeId>,
+    next: Option<SemanticNodeId>,
+}
+
+/// Ordered family membership with local lookup/add/remove.
+///
+/// The hash map is identity lookup only; deterministic semantic order follows
+/// the private previous/next links from `head` to `tail`.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+struct OrderedFamilyMembers {
+    head: Option<SemanticNodeId>,
+    tail: Option<SemanticNodeId>,
+    links: HashMap<SemanticNodeId, FamilyMemberLink>,
+}
+
+impl OrderedFamilyMembers {
+    fn contains(&self, member: SemanticNodeId) -> bool {
+        self.links.contains_key(&member)
+    }
+
+    fn push(&mut self, member: SemanticNodeId) -> bool {
+        if self.contains(member) {
+            return false;
+        }
+        let previous = self.tail;
+        if let Some(previous_id) = previous {
+            self.links
+                .get_mut(&previous_id)
+                .expect("family tail must have a membership link")
+                .next = Some(member);
+        } else {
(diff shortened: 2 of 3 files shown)


yongkyuns:
A1.1 now has a concrete implementation/review/test plan in #957. Please use that case as the acceptance contract for this PR before marking it ready.

Key review finding from the planning pass: detached/scene-owned state cannot remain only an independent residency flag if top-level authored scene ordering is stored elsewhere. A1.1 requires one authoritative representation for scene membership + root order, with local attach/detach, explicit detach/re-add ordering semantics, and no second truth that can drift.

This PR remains the preferred starting implementation: rebase/rework it on current `master`, preserve its good identity/source/family lifecycle work, and extend/revise it to satisfy the A1.1 merge gate. The `ObjectDefinition`/`ObjectId`/`from_scene_definition()` paths remain temporary #959-owned migration seams, not target architecture.

PR #982, opened 2026-09-03
Remove retained SceneDefinition lowering detour

yongkyuns:
## Roadmap ownership

- Owning issue/case: #959 / incremental A4.2
- Architecture layer: migration deletion / retained web lowering

## What changed

Delete one production legacy-scene hop identified by the A4.1 inventory:

```text
SceneSpec
  -> geometry ObjectDefinition list
  -> SceneDefinition::from_parts
  -> RetainedScene::from_legacy
```

The retained web lowerer now consumes the already-normalized `SceneSpec` object stream once and inserts geometry/text directly into the retained migration container in structural painter order.

To preserve the validation that `SceneDefinition::from_parts` previously supplied, `RetainedScene` gets one explicitly migration-only, doc-hidden geometry insertion entry point. It validates geometry/transform/style through the existing core object validator before committing any retained state. #959 owns deletion of this bridge together with the retained migration object model.

## Why this is an A4 deletion

- removes `SceneDefinition` and `ObjectDefinition` from the production `retained_scene_spec_runtime` lowering path;
- removes the production `RetainedScene::from_legacy` call from that path;
- does not introduce serialization or another scene authority;
- does not claim the retained migration container is the target Semantic Scene/lowering API;
- leaves unrelated legacy/test-oracle callers intact until their own per-seam migration handoff is complete.

## Correctness

The old `SceneDefinition::from_parts` detour also rejected malformed geometry state. The new migration geometry insertion preserves that behavior using `validate_object_definition` before mutation, including non-finite geometry/transform/style rejection, duplicate retained identity checks, painter-order checks, and object-ID overflow handling.

A new regression test constructs a `SceneSpec` with a non-finite geometry transform and verifies canonical retained lowering still rejects it even though no legacy `SceneDefinition` is materialized.

Existing canonical-vs-split equivalence coverage continues to protect mixed geometry/text painter order, text resources, tracks, camera identity, and compiled output.

## Architecture check

- [x] Read `AGENTS.md`, `docs/architecture.md`, #959 and the Phase A parallel work overlay.
- [x] Removes a migration authority hop rather than adapting around it.
- [x] No new semantic ID space, scene model, runtime, renderer authority, or transport format.
- [x] New helper is explicitly doc-hidden, migration-only, and has #959 as deletion owner.
- [x] Does not touch worker/cross-context transport; genuine transport remains independently owned.
- [x] Does not overlap the active A6 dependency-ratchet work.

## Validation

- Diff inspected against current `master`; branch is 0 commits behind at PR creation.
- Added focused malformed-geometry regression coverage in `noon-web`.
- GitHub CI is the executable validation for this connector-authored change; no local checkout/test run is claimed here.

## Cleanup / next handoff

This is intentionally one seam, not a wholesale resurrection of the historical retained-lowering branch. The next A4 deletion can remove another production `from_legacy`/scene-document consumer only after its corresponding caller has migrated, or tighten the structural ratchet after this seam is gone.

Refs #953 #959 #977.

### DIFF
--- a/crates/noon-web/src/retained_scene_spec_runtime.rs
+++ b/crates/noon-web/src/retained_scene_spec_runtime.rs
@@ -1,8 +1,6 @@
 use noon::{MathTypst, RetainedScene, Text as NativeText, Typst};
 use noon_compile::RetainedCompiledScene;
-use noon_core::{
-    Color, ObjectDefinition, ObjectId, SceneDefinition, Style, TrackDefinition, Transform2D,
-};
+use noon_core::{Color, ObjectId, Style, TrackDefinition, Transform2D};
 use noon_ir::{ObjectSpec, ObjectSpecContent, SceneSpec, TextSpec, TextSpecKind, TextSpecOptions};
 
 use crate::retained_authoring_scene::MixedRetainedAuthoringError;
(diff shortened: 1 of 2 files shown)


yongkyuns:
Closing this approach rather than bypassing the architecture ratchet. The PR deletes one legacy lowering hop, but its replacement introduces a new migration-only `RetainedScene` geometry insertion API and therefore grows migration-era architecture references. That is the wrong Phase A tradeoff. The correct A4.2 deletion should wait until the target Semantic Scene/lowering path can consume this caller without adding another retained compatibility seam. No ratchet relaxation is warranted.

Issue #961, opened 2026-09-02
[Roadmap A6] Add structural architecture ratchets and Phase A exit gates


yongkyuns:
## Roadmap

Phase A6 child of #953. Architecture authority: `docs/architecture.md`.

## Goal

Make the consolidated architecture executable in CI. Ratchets should reject second scene/runtime authorities, stale migration paths, ambiguous revision handling, hidden O(total-scene) fallbacks and host loops that violate the settle/wake contract.

Prefer deterministic structural/reference-model checks over fragile wall-clock thresholds.

### A6.1 — authority/frontends
Reject normal Python/JS ownership of semantic identity, scene/family/lifecycle/layout/target-state/scheduling truth, effective runtime state or semantic rollback state.

### A6.2 — migration/transport
Reject normal-path `noon::legacy`, migration scene-document/sidecar paths, `from_legacy*` adapters and in-process JSON/transport between Rust engine layers. Allow only explicit debug/export/test or genuine cross-context transport.

### A6.3 — identity and revision domains
Prove distinct responsibilities for:
- semantic identity generation;
- `SceneRevision` for committed authored semantic state;
- `ExecutionRevision` for derived execution state;
- `FrameEpoch` for coherent effective runtime publication;
- resource versions;
- GPU submission/retirement serials;
- `InputSequence` for ordered ingress.

Do not allow one ambiguous global “generation” value to substitute for these domains.

### A6.4 — authored vs effective state
Prove:
- active timeline/native-reactive/host/editor driver writes can update effective Runtime/`FrameEpoch` without changing `SceneRevision` or forcing semantic relowering;
- persistent results reconcile through the semantic mutation contract;
- live getters observe coherent effective state;
- authored/base inspection, where present, is explicit.

### A6.5 — atomic publication
For representative semantic mutation, animation activation, callback and resource-preparation paths:
- validation/preparation failure leaves the prior coherent state visible;
- no half-published semantic/execution/runtime/spatial/renderer state escapes;
- stale callback/resource results cannot overwrite newer revision/generation context.

### A6.6 — semantic transaction contract
Prove one persistent semantic mutation vocabulary and transaction-local provisional references for allocation-and-reference within one transaction. Reject reserved fake semantic IDs or a second draft scene.

### A6.7 — ordering/time
Prove:
- observable updater order is preserved across native/host execution regions;
- required ordered callbacks apply backpressure rather than being silently coalesced/dropped;
- authored scene time is distinct from wall/presentation time;
[...]

Issue #958, opened 2026-09-02
[Roadmap A2] Move Rust authoring onto the Semantic Scene and delete noon::legacy


yongkyuns:
## Roadmap

Phase A2 child of #953. Architecture authority: `docs/architecture.md`. Direct platform-host work is #969.

## Goal

Move idiomatic Rust authoring onto the authoritative Semantic Scene, delete `noon::legacy`, and expose one direct typed Rust path from authoring through execution on native and Rust/WASM targets.

## Contract

Rust is the first-class native authoring API. It calls shared semantic operations directly and does not route through Python, JSON, scene documents or a transport mirror when engine layers share one process/WASM context.

Persistent authoring operations change Semantic Scene authored/base state. Active timeline/reactive/host/editor drivers change effective Runtime state. Rust live getters and segment-completion behavior must respect that distinction.

### A2.1 — public semantic handles
- `Scene` owns/accesses the authoritative Semantic Scene;
- `Mobject`/typed objects reference stable semantic `NodeId`s;
- detached/scene-owned lifecycle is explicit;
- stale generations fail predictably;
- no execution/wire/GPU identity leaks into public semantic identity.

### A2.2 — shared object operations
Construction, add/remove/family, transforms/style/layout/bounds and copy/state/target operations call the same shared semantic implementations consumed by Python.

### A2.3 — animation authoring/activation
Rust animation objects author semantic animation intent. Execution tracks remain lowering/runtime products.

Activation may capture **current effective runtime state** when semantics require it; do not assume animation start values are always the authored declaration-time base state.

### A2.4 — `play()` / `wait()`
Expose logical segment-completion semantics through an idiomatic Rust drive/await surface. Completion means the requested authored-time boundary and required ordered updater/callback dependencies are coherently published and queryable; it does not normally wait for GPU retirement.

### A2.5 — live queries
When an execution session exists, normal current-value queries observe the latest coherent effective state. Authored/base inspection, if exposed, is explicit and separate.

### A2.6 — direct typed execution
Use #969's common `ExecutionSession` path:

```text
Rust API -> Semantic Scene -> lowering -> Runtime -> Renderer
```

No in-process serialization or execution mirror.

### A2.7 — native/web hosts
In conjunction with #969, support native and direct Rust/WASM presentation while keeping platform lifecycle outside semantic/runtime/renderer authority.

### A2.8 — caller migration + deletion
[...]

PR #990, opened 2026-09-03
Add reviewed family-aware semantic scene restructuring

yongkyuns:
Supersedes #988, whose shared head branch was concurrently rewritten after review fixes were applied.

## Roadmap ownership

- Owning case: #957 / A1.3
- Builds on direct semantic family operations merged in #986
- H1 consumers: #958 / A2 and #61 / A3

## What changed

Add recursive scene-facing add/remove semantics on the authoritative `SemanticStore`:

- `add_semantic_scene_nodes(ids)` validates the full batch, collapses descendant roots already represented in the scene projection, promotes unaffected branches in place, then appends explicit target nodes in caller order;
- `remove_semantic_scene_nodes(ids)` removes explicit targets while dissolving only affected family roots and promoting surviving branches at the former root position;
- family membership, source identity, semantic identity, and authored state are not mutated;
- a crate-private root-splice primitive performs bounded local writes to the replaced root, replacement nodes, and immediate neighbors;
- the new module is actually compiled through ordinary Rust module ownership.

## Review fixes

The prototype sorted affected roots by `SemanticNodeId` before cross-root alias de-duplication. That is incorrect after detach/re-attach because NodeId allocation order can disagree with authored scene order. This version uses current linked scene-root order whenever multiple affected roots require first-occurrence de-duplication; the common single-root case avoids the scan.

## Tests

Focused tests cover:
- adding a family collapses existing descendant roots and appends the family;
- removing a descendant promotes a surviving sibling at the exact family-root position;
- invalid mixed batches fail before scene membership changes;
- cross-root alias de-dup follows authored scene order even when NodeId order intentionally disagrees after detach/re-attach.

## Architecture

- one `SemanticStore` / `SemanticNodeId` authority;
- no duplicate family graph, root-order cache, frontend model, runtime model, codec, or renderer dependency;
- no new `#[path]`/`include!` debt;
- reparent/reorder and transform/style mutations remain for the atomic mutation vocabulary rather than ad-hoc partial transactions here.

Refs #953 #957 #958 #61 #961.

### DIFF
--- a/crates/noon-core/src/reactive.rs
+++ b/crates/noon-core/src/reactive.rs
@@ -31,6 +31,8 @@ pub use semantic_store::*;
 mod semantic_scene_operations;
 pub use semantic_scene_operations::*;
 
+mod semantic_scene_restructure;
+
 #[path = "semantic_family.rs"]
 mod semantic_family;
 
--- a/crates/noon-core/src/reactive/semantic_scene_restructure.rs
+++ b/crates/noon-core/src/reactive/semantic_scene_restructure.rs
@@ -360,9 +360,6 @@ mod tests {
 
         store.remove_semantic_scene_nodes(&[removed]).unwrap();
 
-        assert_eq!(
-            store.scene_roots().collect::<Vec<_>>(),
-            vec![shared, tail]
-        );
+        assert_eq!(store.scene_roots().collect::<Vec<_>>(), vec![shared, tail]);
     }
 }
--- a/crates/noon-core/src/reactive/semantic_scene_restructure.rs
+++ b/crates/noon-core/src/reactive/semantic_scene_restructure.rs
@@ -1,0 +1,219 @@
+use std::collections::HashSet;
+
+use super::{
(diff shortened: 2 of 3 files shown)


yongkyuns (commented):
Architecture blocker: `affected_ancestor_closure()` falls back to iterating `store.scene_roots()` whenever more than one affected root participates in alias de-duplication. That makes a two-root/local structural edit O(total scene roots), which violates `docs/architecture.md` §7: structural edits must scale with local dependencies and no feature may introduce an O(total-scene) fallback for a local operation. A1.1/#962 also explicitly established no unrelated-root/sibling scans as an invariant.

Please keep cross-root alias de-dup deterministic without scanning unrelated scene roots. If exact current-root ordering cannot be recovered locally from the existing intrusive representation, prefer an explicit local ordering contract or a genuinely local order-maintenance primitive rather than a global scan. Add a locality regression that would fail if two affected roots among a very large unrelated root set require full-scene traversal.

Issue #63, opened 2026-08-24
[Roadmap A1/B3] Semantic AnimationGraph and local subtree lowering


yongkyuns:
## Roadmap

A1/B3 semantic-animation infrastructure under #953/#954. Architecture authority: `docs/architecture.md`.

## Goal

Represent **authored animation composition** as first-class Semantic Scene intent so persistent edits can relower only affected animation subtrees while execution remains aggressively flattened and Runtime owns the current effective playback state.

Animation declarations are not per-frame state. Active interpolation/progress and activation-time captured source values are derived/effective execution state and must not rewrite the Semantic Scene every frame.

## State model

The Semantic Scene owns stable animation intent for:
- leaf animation declarations;
- parallel/sequence/lagged composition;
- unresolved/default option inheritance and authored time mapping;
- target/lifecycle/source references;
- persistent edits to run time, lag ratio, rate function, ordering and target intent.

Lowering owns derived channels/events/member plans and origin mappings. Runtime owns current activation/progress/effective values at the published `FrameEpoch`.

Execution tracks/channels and active interpolation state are never semantic authority.

## Activation contract

Some Manim-compatible animations must begin from the value that is **currently visible/effective when the animation activates**, not merely the authored/base value that existed when the animation was declared.

Therefore:
- semantic intent may reference a target/property without baking the eventual source value;
- activation may capture the required source transform/style/content-derived value from one coherent effective Runtime context;
- that capture is derived execution/runtime state, not a mutation of the authored animation declaration;
- chained/overlapping animations must not silently fall back to stale authored/base source values;
- persistent final reconciliation, where product semantics require it, occurs through the normal authored mutation/publication path rather than by treating every intermediate frame as semantic state.

### A1/B3.1 — stable animation identity
Animation nodes have generation-safe semantic identity sufficient for persistent edits, diagnostics and origin mapping. Identity generation is distinct from `SceneRevision`, `ExecutionRevision` and `FrameEpoch`.

### A1/B3.2 — local lowering
Lower semantic animation subtrees into existing execution channels/events with origin metadata. Editing run time, lag ratio, rate function, child order or a leaf target invalidates only the affected subtree plus genuine dependents.

A successful persistent edit may advance `SceneRevision`; `ExecutionRevision` advances only when the derived executable projection changes. Pure progress frames advance effective Runtime/`FrameEpoch` without semantic relowering.

### A1/B3.3 — composition semantics
[...]

PR #1002, opened 2026-09-03
Own semantic signal definitions in the Semantic Scene

yongkyuns:
## Roadmap ownership

- Owning case: #957 / A1.4 — animation/reactive/host declarations
- Parent roadmap: #953
- Related target reactive work: #63
- Structural/locality ratchets: #961

## What changed

Move the first native-reactive signal declaration into the authoritative `SemanticStore` identity model:

- add `SemanticNodeKind::Signal(SemanticSignalState)` so signals use the same scene-global generational `SemanticNodeId` allocator as objects/families;
- add high-precision `SemanticSignalValue` (`bool`, `f64`, `SemanticVec3`) and authored expression/source types;
- derived expressions reference `SemanticNodeId`, never the migration-era `SignalId` allocator;
- validate finite constants plus stale/non-signal dependencies before allocation;
- keep signal nodes detached from object/family scene topology through the shared target validators;
- creation writes one semantic slot and validates only the expression/dependency closure.

## Architecture

This deliberately does **not** adapt the target model onto `ReactiveGraphDefinition`, `SemanticScene { definition, reactive }`, or `TimedSemanticScene`. Those are migration-era parallel authored structures and remain deletion/migration work under A1/A4/A1.6 rather than becoming the new contract.

No runtime evaluator, execution slot, timeline track, JSON/codec, frontend ID mapping, or second `SignalStore` is introduced.

## Scope limits / follow-up

This is a creation-only signal declaration slice. Bindings, signal source replacement, dependency-cycle mutation, timeline/native-input declaration ownership, and lowering remain follow-up A1.4/A1.5/A1.6 work. Creation cannot form a cycle because the new node identity does not exist until all dependencies validate.

## Tests

Focused tests cover:
- signals and objects sharing one generational `SemanticNodeId` arena;
- high-precision semantic signal values;
- derived expressions using semantic node references;
- stale and non-signal dependency rejection before insertion;
- non-finite rejection without mutation;
- 10k unrelated-node locality with one-slot signal writes;
- signal rejection from scene/family/updater target operations.

GitHub CI is the executable compile/validation gate for the new enum variant and will identify any remaining exhaustive `SemanticNodeKind` target checks.

Refs #953 #957 #63 #959 #961.

### DIFF
--- a/crates/noon-core/src/reactive.rs
+++ b/crates/noon-core/src/reactive.rs
@@ -35,6 +35,9 @@ mod semantic_scene_restructure;
 
 mod semantic_declarations;
 
+mod semantic_signals;
+pub use semantic_signals::*;
+
 #[path = "semantic_family.rs"]
 mod semantic_family;
 
--- a/crates/noon-core/src/reactive/semantic_declarations.rs
+++ b/crates/noon-core/src/reactive/semantic_declarations.rs
@@ -69,7 +69,7 @@ fn target_node_checked(
     let is_target = match node.kind() {
         SemanticNodeKind::Family => true,
         SemanticNodeKind::AuthoringObject => node.semantic_object_state().is_some(),
-        SemanticNodeKind::Object(_) => false,
+        SemanticNodeKind::Object(_) | SemanticNodeKind::Signal(_) => false,
     };
     if !is_target {
         return Err(SemanticSceneOperationError::NotSemanticAuthoringNode(id));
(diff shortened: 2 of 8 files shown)


yongkyuns:
CI blocker is localized and architectural, not browser-specific: adding `SemanticNodeKind::Signal` made two matches in `crates/noon-web/src/authoring_mobject.rs` non-exhaustive (around lines 954 and 1046). Signals are not family layout/arrange authoring objects, so both sites should explicitly reject `SemanticNodeKind::Signal(_)` alongside `Object(_)` rather than use a wildcard. The WASM build, Rust tests, and Clippy all fail from these same missing arms.

PR #1029, opened 2026-09-03
Ratchet normalized noon-core module ownership

yongkyuns:
## Summary

Add a `noon-core` module-ownership ratchet and wire it into the architecture workflow. It rejects additional organizational `#[path]` and `include!` indirections, including ones present before an unrelated change.

The current `reactive.rs -> semantic_store.rs` seam is treated only as a temporary exception. The ratchet also accepts its reviewed removal, so it does not make that seam a permanent architectural requirement.

## Validation

- `bash scripts/noon-core-module-ownership-ratchet.test.sh`
- `bash scripts/noon-core-module-ownership-ratchet.sh`
- `git diff --check`

## Architecture

The Semantic Scene/store remains the authority. The guard scans the working tree, including untracked files, so it cannot miss a new indirection merely because `git grep` would omit it.

Independent review passed. Local live guard, guard self-test (including untracked regressions and removal of the temporary seam), and `git diff --check` pass. The integrated workspace compile/tests also passed with this guard installed.

### DIFF
--- a/.github/workflows/architecture-ratchets.yml
+++ b/.github/workflows/architecture-ratchets.yml
@@ -30,12 +30,15 @@ jobs:
           scripts/check.sh
           scripts/architecture-ratchet.sh
           scripts/architecture-ratchet.test.sh
+          scripts/noon-core-module-ownership-ratchet.sh
+          scripts/noon-core-module-ownership-ratchet.test.sh
           scripts/active-perf-frontend-ratchet.sh
           scripts/active-perf-frontend-ratchet.test.sh
 
       - name: Exercise architecture guardrails
         run: |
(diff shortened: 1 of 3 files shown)


yongkyuns:
Still relevant, but restack/update the ratchet against current `master` before merge. The current tree still has exactly the temporary `reactive.rs -> semantic_store.rs` `#[path]` seam, so guarding against additional organizational `#[path]`/`include!` growth is useful. One wording change on restack: do not call `reactive.rs` the canonical semantic owner—the Semantic Scene/store is the authority, while this path is only a temporary filesystem seam. The ratchet should allow its later reviewed removal/normalization in the same PR without implying the `reactive` aggregation is target architecture.

PR #1037, opened 2026-09-03
Add atomic semantic family reordering

yongkyuns:
Superseded by merged #1036, which landed the same A1.5 `ReorderMember` contract first.

Independent comparison after #1036 merged found one remaining locality issue worth extracting separately rather than merging this duplicate branch: #1036's committed reorder is O(1), but transaction preflight materializes the entire affected family order (`Vec` + identity scans) whenever `ReorderMember` is present. The #1037 prototype demonstrated an intrusive-link projected preflight that can keep reorder validation/order simulation local.

Closing this PR unmerged. A narrow follow-up will be based on current `master` and will preserve #1036's public mutation/error contract while replacing only the preflight order materialization.

Refs #957 #961 #1036.


yongkyuns:
Superseded by merged #1036 (`73a404c58729d6c76f00d64cfe5afb931465bb8f`).

I compared this implementation before closing it. Its projected-link preflight is a useful stricter read-locality optimization, but it is not required to recover a missing correctness/atomicity invariant from #1036: the merged implementation already keeps committed reorder O(1), preserves parent/membership identity, preflights transaction-local add/remove/reorder state, and bounds preflight work to the directly affected family rather than total scene size.

I am not carrying the alternate `family_reorder.rs` / shared-operation surface forward just to preserve duplicate code. If A6.5 later adds explicit validation-read counters and requires O(1) reorder preflight within very large families, the projected-link idea here is the right reference to reuse as a focused locality optimization.

Issue #960, opened 2026-09-02
[Roadmap A5] Normalize Rust modules and consolidate unjustified crates


yongkyuns:
## Roadmap

Phase A5 child of #953. Architecture authority: `docs/architecture.md`.

## Goal

Make filesystem/module/crate structure communicate the four engine authorities and real platform/resource dependency boundaries without encoding migration history as permanent architecture.

A5 is structural cleanup. It must not create wrappers, transport models or duplicate authorities merely to make moves easier.

## Ownership rules

- ordinary Rust modules mirror actual ownership;
- crates exist only for meaningful independent reuse/test/build or dependency/toolchain/target isolation;
- Semantic Scene, compiler/lowering, Runtime and Renderer remain distinct responsibilities even when crate names/layout evolve;
- platform hosts are integration shells, never semantic/runtime authority;
- `#[path]`/organizational `include!` must not hide unrelated ownership;
- migration names (`legacy`, `mixed`, sidecar/bridge terminology) disappear when the underlying migration seam disappears;
- prefer deletion/merge over compatibility wrappers.

### `noon-native` is justified
#1077/#1079 established a real native platform dependency boundary around `winit` + native wgpu surface/event-loop lifecycle. Keeping `noon-native` as a crate is justified by those platform dependencies, not naming symmetry.

It owns window/event-loop/surface/device/queue/resize/input collection/presentation only. It must not own semantic/runtime scheduling truth; #969's settle/wake work remains a session/host contract rather than a new engine layer.

### `noon-render-wgpu` remains a renderer boundary
Reusable retained WGPU preparation/resource/draw work stays independent from native/browser application lifecycle.

### compiler/runtime boundaries are behavior boundaries
`noon-compile` owns deterministic lowering/specialization; `noon-runtime` owns mutable deterministic execution/effective state. Their current migration-era type names may change, but neither should absorb the other's authority merely to reduce crate count.

### resource boundaries and remaining migration audit

[...]

PR #1038, opened 2026-09-03
Document noon-render-wgpu ownership boundary

yongkyuns:
## Summary

Document `noon-render-wgpu` as the reusable retained WGPU renderer and add its package metadata.

The README now distinguishes current behavior from the target contract. It records that renderer input must become coherent effective runtime state at a published `FrameEpoch`, with versioned resources retained until in-flight submissions are safe, without claiming those publication and retirement mechanics are already implemented. It also records that the renderer does not read mutable Semantic Scene truth or own session wake/scheduling.

## Validation

- `git diff --check`

The repository's integrated Rust gate is pending in the coordinating worktree.

## Architecture

`noon-native` owns native lifecycle and `noon-web` owns browser/WASM integration. This crate remains reusable retained GPU rendering under #960.

Independent review passed. Local `cargo check --workspace --all-targets --all-features`, `cargo metadata --no-deps`, and `git diff --check` passed in the cleanup worktree. Documentation/package metadata only; remaining publication/resource-retirement work is explicitly identified as target behavior.

### DIFF
--- a/crates/noon-render-wgpu/Cargo.toml
+++ b/crates/noon-render-wgpu/Cargo.toml
@@ -3,6 +3,8 @@ name = "noon-render-wgpu"
 version = "0.1.0"
 edition = "2021"
 license = "MIT"
+description = "Reusable retained WGPU renderer for Noon runtime state"
+readme = "README.md"
 
 [features]
 default = ["native", "ci-noop"]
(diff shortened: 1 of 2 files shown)


yongkyuns:
The crate-boundary decision is still correct and is now reinforced by merged #1077/#1079: `noon-render-wgpu` is the reusable retained renderer while `noon-native` owns native lifecycle. Restack before review, and update the README wording to the current architecture: renderer input is coherent effective Runtime/`FrameEpoch` state plus versioned resources/change sets; renderer does not read mutable Semantic Scene truth, own session wake/scheduling, or retire an old resource version before in-flight submissions are safe.

PR #1045, opened 2026-09-03
Document noon-compile ownership boundary

yongkyuns:
## Summary

Refresh `noon-compile` package documentation for the current typed lowering path and add its README metadata.

The crate now documents `SemanticStore -> SemanticExecutionLoweringOutput`, installed by `ExecutionSession`, plus canonical reactive/animation lowering and fail-before-publication validation. It assigns semantic-derived mapping and dense indices to lowering while leaving durable `ExecutionSlotTable` identity to runtime. `SceneDefinition` is described only as a temporary compatibility seam, not as an authoring IR or peer scene model.

## Validation

- `git diff --check`

The repository's integrated Rust gate is pending in the coordinating worktree.

## Architecture

`noon-compile` owns typed semantic analysis/lowering/specialization into derived execution data; runtime, rendering, and platform lifecycle remain outside its boundary.

Independent review: passed after correcting compiler/runtime identity ownership. Local validation: `cargo check --workspace --all-targets --all-features`, `cargo metadata --no-deps`, and `git diff --check` passed in the cleanup worktree. This PR changes documentation/package metadata only.

### DIFF
--- a/crates/noon-compile/Cargo.toml
+++ b/crates/noon-compile/Cargo.toml
@@ -3,7 +3,8 @@ name = "noon-compile"
 version = "0.1.0"
 edition = "2021"
 license = "MIT"
-description = "Compiler from Noon authoring IR to dense runtime data"
+description = "Typed semantic lowering into renderer-independent execution data"
+readme = "README.md"
 
 [dependencies]
 noon-core = { path = "../noon-core" }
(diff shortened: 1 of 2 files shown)


yongkyuns:
The `noon-compile` boundary is still justified, but this documentation is stale enough that it should not merge unchanged. Current master now has the typed `SemanticStore -> SemanticExecutionLoweringOutput` path used by `ExecutionSession`; describing `SceneDefinition` as the current Phase A input no longer captures the primary architecture. On restack, describe this crate as owning typed semantic analysis/lowering/specialization into derived execution data, including canonical reactive/animation lowering and fail-before-publication validation. Keep `SceneDefinition` only as an explicitly temporary compatibility seam where it still exists; do not call the input an `authoring IR` or imply a peer scene model.

PR #1052, opened 2026-09-04
Superseded: A1.6 semantic projection to compiled slots

yongkyuns:
## Scope

Continue A1.6 from #1051 by materializing its already value-lowered `SemanticExecutionProjection` into the existing `CompiledScene` / `SlottedSceneInstance` execution path.

## Changes

- add `CompiledScene::from_semantic_projection` as an object-payload adapter;
- consume #1051's compact transform/style/content values directly instead of lowering semantic values a second time;
- order compact frame rows by the authoritative semantic presentation key `(z_index, insertion_order)`;
- preserve the existing semantic→execution `ObjectId` mapping while row order changes independently;
- lower analytic circle/rectangle/line content into the current geometry payload;
- reject versioned geometry resources rather than erase their version into an unversioned execution key;
- reject text until the stable compiled slot payload can carry text directly;
- keep timeline/reactive declarations outside this object-value adapter for dedicated A1.6 lowering channels;
- add a runtime integration fixture proving the compiled result initializes durable `ExecutionSlotId`s in `SlottedSceneInstance`.

## Architecture

This slice introduces no peer execution model, retained semantic mirror, serialized bridge, or alternate slot allocator. `CompiledScene` and `ExecutionSlotTable` remain the execution destination. Semantic presentation metadata is consumed as execution ordering rather than discarded.

## Validation

Focused compiler tests cover value transfer, semantic painter ordering, versioned-resource rejection, and text rejection. The runtime integration test exercises semantic store → typed projection → compiled object rows → stable execution slots.

Refs #957 #959.

### DIFF
--- a/crates/noon-compile/src/semantic_lowering.rs
+++ b/crates/noon-compile/src/semantic_lowering.rs
@@ -1,3 +1,6 @@
+mod compiled_scene;
+pub use compiled_scene::*;
+
 use std::collections::{HashMap, HashSet};
 
 use noon_core::{
--- a/crates/noon-compile/src/semantic_lowering/compiled_scene.rs
+++ b/crates/noon-compile/src/semantic_lowering/compiled_scene.rs
@@ -1,0 +1,232 @@
+use std::collections::BTreeMap;
+
+use noon_core::{
+    GeometryRef, GeometryResourceHandle, SemanticNodeId, SemanticObjectContent, StoredGeometry,
+    TextResourceHandle,
+};
+
+use crate::{CompiledObject, CompiledScene, DynamicProperties};
+
+use super::SemanticExecutionProjection;
+
+/// Failure while materializing the object-value projection into the current
+/// geometry-backed compiled execution representation.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SemanticCompiledSceneError {
+    TooManyObjects(usize),
+    UnsupportedGeometryResource {
+        node: SemanticNodeId,
+        resource: GeometryResourceHandle,
+    },
+    UnsupportedText {
+        node: SemanticNodeId,
+        text: TextResourceHandle,
+    },
+}
+
+impl std::fmt::Display for SemanticCompiledSceneError {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::TooManyObjects(count) => {
+                write!(formatter, "semantic projection contains too many objects: {count}")
+            }
+            Self::UnsupportedGeometryResource { node, resource } => write!(
+                formatter,
+                "semantic object {}:{} uses versioned geometry resource {}@{} before the compiled resource key carries version identity",
+                node.slot(),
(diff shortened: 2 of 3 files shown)


yongkyuns:
Coordination note: #1051 is the preceding A1.6 value-lowering slice and should land first. It changes `SemanticExecutionProjection` from borrowed `SemanticObjectState` to owned execution-facing values (`SemanticObjectContent`, compact `Transform2D`/`Style`, presentation) with fail-atomic validation. Please rebase this draft after #1051 rather than keeping a second transform/style lowerer here.

After rebase, the useful scope in this PR is the actual stable-slot consumer. In particular, avoid making the semantic boundary geometry-only merely because current `CompiledScene` stores `GeometryRef`: text/versioned content should remain typed at the A1.6 boundary, with unsupported backend consumption explicit and as late as possible. This should stay a consumer of the one lowerer, not become an alternate lowerer.

yongkyuns (commented):
A1.6 direction is right (`CompiledScene` should be the real destination), but this draft currently creates two lowering policies and violates the module/migration ratchets.

Blocking changes before this should leave draft:

1. **Stack on #1051 and consume its lowered execution values.** #1051 is exact-head green and now owns semantic -> compact `content` / `base_transform` / `base_style` / `presentation` lowering. This PR currently re-implements `lower_transform`, `lower_style`, `lower_paint`, scalar/vector range policy, and related errors from `object.state`. That gives A1.6 two independently evolving semantic-value lowerers. After #1051, `from_semantic_projection` should consume `SemanticExecutionObject`'s already-lowered fields and own only the conversion into the existing stable `CompiledScene` slot/index structure plus capability rejection that is genuinely CompiledScene-specific.

2. **Do not add `#[path = "semantic_compilation.rs"]` under `retained.rs`.** Architecture CI is already rejecting this as new hidden module ownership. The target semantic->CompiledScene consumer also should not be owned by the migration-era retained module. Give it ordinary module ownership in the compiler tree (or a real child module location) with no `#[path]`/`include!` seam.

3. **Do not grow migration-era references in target source/docs.** Architecture CI also rejectss the new `RetainedCompiledScene` token. State the target boundary without naming/propagating that migration model.

4. **Keep reactive binding loss explicit after stacking.** #1051's owned projection currently carries content/transform/style/presentation but not `signal_bindings`. This draft presently rejects bindings by reading `object.state`; that path disappears once it correctly consumes #1051. Either preserve the needed binding metadata in the semantic execution handoff (preferred if it is part of the eventual plan), or make the compiled consumer receive the authoritative store solely for binding lowering/validation. Do not silently accept a projection whose reactive bindings were dropped.

5. Keep the current good fail-closed behavior for unsupported versioned geometry/text/paint/presentation capabilities; also validate analytic geometry before constructing a `CompiledScene`, since this constructor bypasses `SceneDefinition` validation.

Current CI evidence matches (2)/(3): Architecture Ratchets fail on the new path seam + migration reference; Fast Gate only fails formatting, while Rust tests/browser checks pass. The design issue in (1)/(4) is not something CI can catch automatically.

Issue #969, opened 2026-09-03
[Roadmap A2/A6] Add direct Rust execution hosts for native and web


yongkyuns:
## Authority

Phase A2/A6 execution-session and platform-host owner under #953 and `docs/architecture.md`.

## Goal

Provide one direct typed Rust execution/session path shared by native and single-context Rust/WASM hosts, without introducing another scene/runtime authority or in-process transport model.

## Core path

```text
Semantic Scene
   -> canonical typed lowering
   -> existing Runtime
   -> FrameState / FrameChanges
   -> existing Renderer
   -> native host OR browser/WASM host
```

`ExecutionSession` is the live orchestration boundary over existing compiler/runtime authorities. It may coordinate activation, time advancement, effective-state queries, publication barriers, wake/sleep, input delivery and host-facing frame changes. It must not own/mirror the Semantic Scene, expose raw mutable runtime patch authority, or become a fifth engine layer.

## State/time contract

- Semantic Scene owns authored/base state;
- Runtime owns effective state at the current published `FrameEpoch`;
- animation/native-reactive/host/editor driver output normally changes effective state without semantic relowering;
- activation-time animation lowering may capture current effective state;
- authored scene time is distinct from wall/presentation time;
- `play()`/`wait()` are logical segment-completion barriers, not necessarily blocking calls;
- a clean session exposes enough wake/deadline information that platform hosts can sleep instead of busy-polling.

### A2/A6.1 — common typed execution session — landed
#1071 established `noon::ExecutionSession` over the canonical semantic lowering/runtime path. Subsequent animation activation uses the same session and existing runtime publication machinery.

### A2/A6.3 — direct Rust/WASM host + canonical 2D camera — landed foundation
- #1072/#1075 proved one-context Rust/WASM construction through `ExecutionSession` into the existing canvas renderer without scene/execution JSON;
- #1081 moved canonical 2D camera identity onto the semantic/execution path and derives `ExecutionSession::camera()` from the current effective runtime frame;
- native and direct-WASM hosts consume the same effective camera state; no duplicate host camera authority is required.

Broader B6 camera semantics remain #89, but the Phase A host boundary is established.

### A2/A6.2 — native Rust host — presentation and input foundations landed
- #1077 established the native platform-host crate and real platform lifecycle;
- #1079 proved the production native surface/render/present path under CI;
[...]

Issue #955, opened 2026-09-02
[Roadmap C] Native interaction, locality, and live authoring


yongkyuns:
## Authority

Phase C execution tracker under `docs/architecture.md`.

Phase C adds interaction, callbacks, live editing and locality on the existing four-layer engine. It must not create a second interactive scene model, runtime, scheduler or patch vocabulary.

## Phase invariant

```text
input / host callback / editor / hot reload
                    |
                    v
          live execution/session control
                    |
                    v
              staged update batch
        +-----------+------------+
        |                        |
        v                        v
 effective driver writes   authored semantic mutations
        |                        |
        +-----------+------------+
                    |
      validate + impact + prepare
                    |
                    v
           coherent publication
                    |
             Runtime -> Renderer
```

The key distinction is architectural:

- timeline, native-reactive, host-updater and editor-driver output normally changes **effective runtime state** and may advance `FrameEpoch` without changing authored/base state or forcing semantic relowering;
- persistent property/content/structural edits use the shared `SemanticMutationTransaction`, advance semantic revision when committed, and trigger only the affected lowering/runtime/resource work;
- when a driver result must persist after release/completion, reconcile that result into authored/base state through the normal semantic mutation path;
- mixed updates publish coherently: no half-published semantic/runtime/renderer state.

Observable updater order is semantic. Native and host work may be fused or reordered only when the difference is provably unobservable.

---

## C1 — host callbacks and staged publication

Primary: #70.

- coherent versioned callback read view plus transaction overlay;
- ordered native/host updater regions rather than a global native-before-host phase split;
- effective property/style writes stay on the runtime driver lane;
- persistent content/structure changes use the semantic mutation vocabulary;
- structural callback batches may use transaction-local provisional identities, never reserved semantic IDs or a second draft scene;
- required ordered callbacks apply backpressure to authored-time progression; sampled input may still coalesce independently;
- stale callback results cannot publish against newer revisions/generations.

**Gate:** arbitrary host behavior is supported without making unrelated content host-dynamic or publishing partial state.

### Optional host-expression capture (#1152)
Evaluate capture only when measurements and compatibility requirements justify it. It is a frontend translation into existing shared semantic expressions and immutable/timeline/native-reactive execution, not a new execution class or interpreter. This does not add a Phase A prerequisite or change the required host-callback gate above.

[...]

PR #1054, opened 2026-09-04
Clarify live-session and scalability architecture

yongkyuns:
## Summary

Strengthen `docs/architecture.md` around the corner cases raised by live scripting, high-object-count interaction, host-language callbacks, and CPU/GPU boundaries without changing Noon's four-layer authority model.

The update deliberately borrows **mechanisms rather than product architectures** from data-oriented/game/render/interactive systems. Noon remains a scriptable vector animation system, not a general-purpose ECS/game engine/UI framework.

## Main architecture additions

- define a **live-session control plane** for script continuations, logical `play()`/`wait()` completion barriers, input, callbacks, editor actions, hot reload and frame epochs; explicitly state that this is not a fifth scene/runtime authority;
- distinguish **authored/base semantic state** from **effective runtime state**, so ordinary timeline/reactive/host updater writes do not imply semantic mutation or relowering;
- make `FrameEpoch` the coherent effective publication unit, referencing `SceneRevision` and `ExecutionRevision`, which may remain unchanged across ordinary dynamic frames;
- preserve semantic updater order across mixed native/host execution regions rather than globally ordering by implementation class;
- define transactional host callback overlays, transaction-local/provisional identity for structural callbacks, and revision-pinned cross-context callback read mirrors;
- define safe read-miss suspension semantics, required-callback stalls, and bounded discrete-event backpressure/overflow behavior;
- distinguish **semantic structure from execution/render structure**, allowing compiler-derived transform/style/instance domains for compatible high-fanout operations;
- permit ECS/data-oriented techniques only below lowering: dense slots/SoA/change revisions/specialized queues, with no public ECS authoring model;
- extend locality requirements to include transaction validation/preflight, temporary memory, bridge calls/bytes, lowering, spatial work and GPU publication;
- strengthen immutable/versioned resource publication, in-flight GPU retirement and long-running reclamation rules;
- define compact runtime-to-renderer change publication, derived renderer acceleration structures, and a hard no-synchronous-GPU-readback normal-path rule;
- add measurable scalability workloads and explicit non-goals preventing architecture drift into ECS/world/plugin/state-machine/display-list-centric designs.

## Product boundary

The intended shape remains:

```text
Semantic Scene -> Execution Plan -> Runtime -> retained Renderer
```

with thin Rust/Python/optional JS authoring and platform hosts at the edges.

The author-facing goal is a ManimGL-like **one live scene** experience, while the execution model remains specialized so Python/JS object traversal is not required on the frame path for timeline/native-reactive behavior.

## Roadmap impact

No new phase is introduced. Existing Phase A execution-session work (#969) and Phase C interaction/locality/live-authoring work (#955) remain the implementation owners; A1.6 continues only the typed/incremental `SemanticStore ->` existing execution handoff rather than absorbing another runtime abstraction.

Documentation only; no runtime/code behavior changes.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -19,10 +19,14 @@ Noon is a high-performance animation and interactive graphics system with:
 - a built-in renderer/runtime shared by native and web targets;
 - Manim-compatible Python authoring for supported common 2D behavior as a wrapper over shared Rust semantics;
 - optional future JavaScript/TypeScript authoring as another wrapper over shared Rust semantics;
-- deterministic offline and realtime execution;
+- a live-scene programming model in which script commands, deterministic playback, native input/reactive behavior, editor actions and explicit host callbacks operate on the same semantic scene rather than separate interactive/script scene engines;
+- deterministic offline and realtime execution wherever semantics are deterministic;
 - native reactive interaction without requiring Python on the frame path;
 - explicit support for arbitrary host-language callbacks when user code genuinely requires them;
-- retained, incremental GPU rendering that scales with changed and visible work rather than total scene size.
+- retained, incremental GPU rendering that scales with changed and visible work rather than total scene size;
+- aggressive data-oriented execution/render specialization below the semantic layer without exposing a general-purpose ECS, game-engine scheduler or renderer scene graph as authored truth.
+
+The authoring experience should feel like one continuously live scene. The implementation is deliberately not one continuously traversed host-language object graph: static, timeline, native-reactive and host-dynamic behavior are separated and specialized below the semantic boundary so large mostly-static scenes do not inherit the cost of their most dynamic feature.
 
 ### Rust-native product invariant
 
@@ -157,6 +161,41 @@ Serialization is not a fifth scene model. It is an optional codec around one of 
 
 **Normal in-process engine boundaries are typed Rust boundaries.** `Rust API -> Semantic Scene -> Execution Plan -> Runtime -> Renderer` must not serialize to JSON or any other wire representation as part of ordinary native authoring, lowering, execution, mutation, or rendering. The same rule applies when the Rust engine is compiled to WASM and all layers execute in one browser context.
 
+### Live-session control plane
+
+A continuously live authoring experience requires coordination between ways of changing or advancing the four engine layers, but that coordination is **not a fifth state authority**.
+
+Conceptually:
+
+```text
+script continuation / segment completion
+native input + events
+host callbacks
+editor/session manipulation
+hot-reload reconciliation
+          |
+          v
++----------------------------------+
+| live execution/session control   |
+| ordering + barriers + revisions  |
+| no independent scene authority   |
++----------------+-----------------+
+                 |
+        +--------+--------+
+        |                 |
+        v                 v
+Semantic operations    Runtime tick/wake
+        |                 |
+        +--------+--------+
+                 |
+                 v
+              Renderer
+```
+
+The control-plane responsibility belongs with runtime/execution integration. It may coordinate script continuations, published frame epochs, wake/sleep, input delivery, ordered updater/callback evaluation and publication barriers, but it must not own a second scene, timeline, scheduler truth or renderer model.
+
+`play()`/`wait()`-class operations are **logical segment-completion barriers**, not necessarily blocking function calls and not exclusive interaction modes. Python may expose a barrier synchronously by suspending/blocking an authoring worker; native Rust may drive or await a session; browser/WASM code must be able to yield/await rather than blocking the event loop. Native input/reactive behavior may continue during playback when semantics allow it. An `interact()`-style API, if exposed, means the authoring continuation waits while the same live runtime continues processing input/reactive/callback/render work; it does not switch to another scene engine.
+
 ---
 
 ## 3. Semantic Scene
@@ -195,6 +234,28 @@ Semantic values must not be constrained by the current 2D/f32 renderer. The sema
 
 Do not maintain permanent parallel `legacy` and `semantic` transform/style/object models.
 
+### Authored/base state and effective runtime state
+
+The Semantic Scene owns **authored/base state**: the persistent authored value, declaration, relationship or resource reference from which execution is derived.
+
+The Runtime owns the **effective state at a published frame epoch**: the current value after timeline interpolation, native reactive evaluation, host/updater effects and other active drivers have been applied according to their semantics.
+
+Language compatibility getters that ask for the current live value normally observe the latest coherent effective state, not a stale authored/base value. An explicit authored/base-state inspection API may exist when useful, but it is a different query.
+
+Active timeline/reactive/host/editor drivers normally write **effective runtime state**, not authored/base state. Such writes may advance `FrameEpoch` and renderer/spatial state without creating a new `SceneRevision` or requiring semantic relowering. A driver write becomes an authored semantic mutation only when the semantic operation itself is persistent/structural or when compatibility/product semantics explicitly reconcile a resulting effective value back into authored state.
+
+When animation/updater completion semantics require an effective final value to persist after the driver is released, that persistent result is reconciled into authored/base state through the normal mutation contract and coherent publication path. A conforming frontend must therefore not disagree about a current getter merely because authored and effective representations live in different layers.
+
+### Semantic structure is not execution structure
+
+Semantic families, object hierarchy and aliases exist because they are observable authoring semantics. They must not be forced to double as the optimal execution or render hierarchy.
+
+The compiler may factor compatible high-fanout behavior into derived execution structures such as shared transform/style domains, instance groups or other compact representations when observable semantics are preserved. A semantic family and a derived execution/render group are not required to be one-to-one. Aliasing, conflicting drivers, ordering or other observable semantics may prevent factoring and require explicit member work.
+
+Compact fan-out does not imply that every downstream operation becomes O(1). If an observable consumer genuinely needs per-leaf materialization, for example world-space bounds or precise hit-test records, work may scale with the affected leaves that must be materialized. Domain split/deoptimization may scale with the size of the affected domain, but it must not silently expand to unrelated scene content.
+
+This is a one-way specialization rule: execution/render structure may be derived from semantic structure, but derived execution/render grouping never becomes authored truth.
+
 ---
 
 ## 4. Language frontends
@@ -216,7 +277,7 @@ A Rust application compiled to WASM must be able to build and render the same No
 - Manim-compatible class hierarchy and method signatures;
 - Python argument normalization and iterable/vector conversion;
 - Python callable identity and callback invocation;
-- Scene subclass discovery;
+- Scene subclass discovery and sequential authoring control flow;
 - Python-appropriate exceptions and wrapper metadata.
 
 ### Python must not own
@@ -230,6 +291,7 @@ A Rust application compiled to WASM must be able to build and render the same No
 - animation timing/scheduling/interpolation;
 - signal evaluation;
 - transaction rollback state;
+- retained execution or renderer state;
 - a second retained-text scene;
 - a second renderer/runtime model.
 
@@ -247,6 +309,40 @@ class Scene:
 ```
 
 The Rust public API calls the same semantic operations directly, without any wrapper or serialization hop.
+
+Thin does not mean one cross-language call per scalar property access. Frontends may batch construction/mutation/query operations and host-callback read/write sets where that preserves observable semantics. Large immutable content should cross a real language/context boundary as a resource payload/handle once, not be repeatedly reconstructed as scene-shaped messages.
+
+### Script continuation semantics
+
+Sequential authoring is allowed to depend on completed runtime state. `play()`/`wait()`-class operations therefore establish a logical segment-completion barrier:
+
+```text
+host authoring continuation
+        |
+        | author mutations / animation intent
+        v
+stage + coherently publish executable segment
+        |
+        | execute deterministic/realtime segment
+        v
+logical segment completion
+        |
+        v
+resume host authoring continuation
+```
+
+The host mechanism used to wait is target-specific: a Python worker may block/suspend, native Rust may drive/await, and browser/WASM must yield/await without blocking browser frame/input servicing.
+
+A segment is complete when:
+
+- the requested authored-time boundary has been reached;
+- every required ordered callback/updater dependency through that boundary has completed;
+- resulting authored semantic mutations and effective driver writes, together with any required incremental lowering/preparation, runtime/spatial changes and renderer-facing delta, have been coherently published;
+- the resulting effective state is queryable by the resumed authoring continuation.
+
+Completion normally does **not** require waiting for GPU retirement/fence completion or for an old resource version to be physically freed. After the barrier, live-value queries observe the coherent effective state defined above.
+
+This continuation protocol must not require a host interpreter while a deterministic segment is running unless host-dynamic behavior is actually scheduled.
 
 ### Paired executable examples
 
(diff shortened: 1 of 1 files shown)


yongkyuns:
Architecture review: the overall direction is strong and I think most of this should stay. The PR correctly preserves the four-layer authority model, keeps ECS/data-oriented ideas below lowering, separates semantic structure from execution/render structure, strengthens locality beyond final writes, and addresses resource/GPU lifetime and synchronous-readback hazards. It also matches the current A1.6 direction: feed the existing `CompiledScene` / `ExecutionSlotTable` incremental machinery rather than introducing another runtime model.

Because `docs/architecture.md` is the authoritative architecture, though, I think a few contracts should be made more precise before merge. These are the main changes I would make:

1. **Make transaction + lowering publication atomic.**

   The representative epoch currently has `commit/publication barrier` before `incremental lowering/runtime maintenance`. That leaves an undefined state if semantic mutation commits but affected lowering/resource preparation then fails.

   I suggest the conceptual contract be closer to:

   ```text
   ingress
     -> timeline evaluation
     -> ordered reactive/updater evaluation
     -> collect/stage mutations
     -> validate + impact analysis + incremental lowering/preparation
     -> atomic publication barrier
          semantic revision
          execution revision
          runtime state
          spatial/index changes
          renderer-facing delta
     -> visibility/render preparation
     -> GPU upload/encode/present
   ```

   For async content/resource preparation, define one of two explicit behaviors: either the operation remains staged until coherent publication is possible, or an explicit semantic `pending` state is published while the previous execution/resource generation stays active. We should not allow a half-published Semantic Scene / old Execution Plan state accidentally.

2. **Define authored/base state vs effective runtime state.**

   The document says the Semantic Scene is authored truth while Runtime owns current mutable property values, then says post-`play()` queries observe completed state. We should specify exactly what a live authoring getter observes.

   Suggested model:

   - **authored/base state** lives in Semantic Scene;
   - **effective state at published epoch E** lives in Runtime;
   - Python compatibility getters normally observe the latest coherent effective state;
   - when animation completion semantics require the final value to persist, that final state is reconciled back into authored state through the normal mutation contract.

   Otherwise two conforming implementations could disagree about `mobject.get_center()` after `play()` or while a binding/updater/editor driver is active.

3. **Do not globally order `native reactive` before `host dynamic`. Preserve semantic updater order.**

   Classification is good; global phase reordering by implementation class is not always semantics-preserving.

   In 3b1b ManimGL, animation interpolation happens first and then mobject updaters run, with registered updater ordering observable. Consider semantic updater order `native A -> Python B -> native C`: Noon cannot legally execute `A,C -> B` just because A/C are native.

   I would replace the global native-reactive/host-dynamic split with an **ordered evaluation plan** containing native regions and host-callback barriers. Native-only regions stay fully native; independent work may be fused/reordered only when the compiler proves it unobservable.

4. **Make the cross-context host callback read contract concrete.**

   The transactional read/write overlay is a major improvement, but `declared read sets`, `observed read-set caching`, or `batched miss/fetch protocols` still leaves the hard case unresolved: arbitrary Python unexpectedly reads property X across a worker/context boundary.

   I think the architecture should explicitly permit a **versioned, non-authoritative callback read mirror/cache** at genuine cross-context boundaries. It contains callback-readable state pinned to a scene/runtime revision; callback writes go to the local overlay; one transaction returns. It is explicitly not scene authority.

   If a partial mirror is allowed, define read-miss behavior explicitly (e.g. suspend/retry/fail before side effects), rather than allowing stale/undefined scalar reads.

5. **Specify slow-host-callback semantics, not only stale-result rejection.**

   `stale callback result cannot overwrite newer state` is necessary but not sufficient. If required Python updater N is still running, can authored scene time advance to N+1? Can native work consume state that should have included N?

   Suggested default: required ordered compatibility callbacks are **non-coalescible**. Authored simulation time cannot advance past that dependency barrier until the callback result is available. Presentation may keep showing the last coherent frame, and sampled input may continue coalescing. Only callbacks explicitly classified as latest-value/coalescible may be dropped/replaced.

   Slow Python should cause poor frame rate, not undefined semantics.

6. **Describe `play()` as a logical continuation barrier, not necessarily a blocking call.**

   This matters for Rust/WASM: a browser main-thread WASM function cannot synchronously block waiting for future animation frames while still servicing `requestAnimationFrame` and input.

   Architecturally, treat `play()`/`wait()` as segment-completion barriers. Python may expose this synchronously by suspending/blocking the authoring worker; native Rust can drive or await the session; Rust/WASM must yield/await. Same semantics, different host mechanics.

   Also define completion precisely: requested authored-time boundary reached; required callbacks through that boundary complete; resulting mutations/lowering published; effective state queryable. It should normally not require waiting for GPU retirement/fence completion.

7. **Separate identity generations from state revisions/epochs.**

   The current list mixes several different concepts under `generation`/`epoch`. Noon already uses generation for slot reuse validity (`SemanticNodeId`, `ExecutionSlotId`) and separately has layout generation.

   I suggest naming the conceptual domains distinctly:

   - `NodeId` / `ExecutionSlotId` generation: identity validity;
   - `SceneRevision`: committed semantic transaction revision;
   - `ExecutionRevision`: derived execution revision;
   - `FrameEpoch`: published presentation epoch;
   - `ResourceVersion`: immutable resource replacement;
   - `SubmissionSerial`: GPU retirement/fence ordering;
   - `InputSequence`: ordered external events.

   Revisions from different domains are not directly comparable. Async/callback results carry the specific revision/generation context required to prove applicability.

8. **Qualify high-fanout complexity guarantees and explicit maintenance exceptions.**

   Shared transform/style/instance domains are a good addition, but one transform affecting 100k leaves is not necessarily O(1) end-to-end if an observable consumer must materialize 100k world-space bounds, hit-test records, etc.

   Suggested wording: work scales with changed domains **plus affected leaves that must actually be materialized for an observable consumer**. Domain split/deoptimization may be O(size of affected domain), but must not become O(total unrelated scene).

   Also explicitly exempt deliberate maintenance barriers from the local-edit contract. The existing execution-slot compaction model is a good example: ordinary mutation remains local, while an explicit generation barrier may perform a controlled rebuild/compaction.

9. **Add a short time/determinism subsection.**

   We now need to distinguish at least:

   - authored scene/simulation time;
   - wall/presentation time;
   - input event sequence/timestamps;
   - GPU presentation/submission time.

   Define which clock compatibility `dt` uses. A paused scene should be able to process input while authored `dt == 0`. Behavior explicitly depending on wall time should be classified externally timed/non-deterministic.

   Replay classification should also cover pointer/keyboard input, editor manipulation, async/network results, etc., not only host callbacks.

10. **Make roadmap ownership explicit for the new requirements.**

   Suggested ownership:

   - #969: logical execution-session/segment completion, continuation semantics, serialized publication lane/session epochs;
   - #955 C1: mixed native/host ordered evaluation, callback view/mirror, latency policy, driver arbitration;
   - #955 C2: input sequencing/time semantics/replayable input;
   - #955 C3/C4: renderer/resource publication and retirement;
   - #955 C7: high-fanout execution domains as measurement-driven specialization;
   - A1.6 continues establishing the typed/incremental SemanticStore -> existing compiled/runtime handoff and should not absorb another runtime abstraction.

Overall: **the architecture shape is right**. I would keep the semantic-vs-execution separation, data-oriented specialization rule, locality expansion, resource lifetime rules, renderer publication model, no-sync-GPU-readback rule, settle/wake behavior, and explicit non-goals. The items above are mainly about making the live-script/runtime contract precise enough that Python/Rust/WASM/GPU implementations cannot choose conflicting semantics later.

yongkyuns:
Addressed the architecture review and restacked the PR onto current `master` (`168a0197...`, including merged #1071). The branch is now one documentation-only commit: `edbb1cfc...`.

Changes corresponding to the review points:

1. **Atomic transaction + lowering publication** — the runtime cycle now stages mutations, validates/does impact analysis and incremental lowering/preparation first, then publishes `SceneRevision` + `ExecutionRevision` + effective runtime/spatial state + renderer-facing delta atomically. Async resource work stays staged unless an explicit semantic pending state is part of the feature.
2. **Authored/base vs effective state** — added an explicit state model: Semantic Scene owns authored/base state; Runtime owns coherent effective state at a published frame epoch; compatibility getters normally observe effective state; persistent animation completion reconciles back through normal mutation/publication.
3. **Updater ordering** — removed the global `native reactive -> host dynamic` ordering assumption. The Execution Plan now preserves semantic updater/effect order using native regions plus required host-callback barriers; only proven-unobservable work may be reordered/fused.
4. **Cross-context callback reads** — explicitly permit a revision-pinned, non-authoritative callback read mirror/cache. Partial-mirror misses must fetch/retry or fail before externally visible callback effects; stale/undefined scalar reads are not allowed.
5. **Slow host callbacks** — required ordered compatibility callbacks are non-coalescible by default. Authored/simulation time cannot advance through an unresolved callback dependency. Presentation may retain the last coherent frame and sampled input may continue coalescing.
6. **`play()`/`wait()` semantics** — now defined as logical segment-completion barriers rather than necessarily blocking calls. Python/native/browser-WASM may block, drive/await, or yield/await respectively. Completion requires authored boundary + required callbacks + coherent publication + queryable effective state, but not GPU retirement/fence completion.
7. **Generation terminology** — separated identity generation from `SceneRevision`, `ExecutionRevision`, `FrameEpoch`, `ResourceVersion`, `InputSequence`, `CallbackEpoch`, and `SubmissionSerial`; different domains are explicitly non-comparable.
8. **High-fanout/locality qualifications** — compact domains are allowed, but observable consumers may materialize genuinely affected leaves. Domain split/deoptimization may scale with the affected domain, never unrelated scene state. Explicit/instrumented maintenance barriers are a separate exception from ordinary local-edit complexity.
9. **Time/determinism** — added authored/simulation time vs wall/presentation time vs input sequence/timestamps vs GPU submission time; compatibility `dt` comes from authored time, so paused input may run with authored `dt == 0`. Replay classification now covers input/editor/async/network/external behavior too.
10. **Roadmap ownership** — made ownership explicit: #969 for logical session/segment completion and coherent publication lane; C1 for mixed native/host ordering, callback views, latency and drivers; C2 for input/time/replay; C3/C4 for renderer/resource publication and retirement; C7 for measurement-driven high-fanout specialization; A1.6 remains the typed incremental handoff into the existing compiled/runtime machinery and does not acquire a new runtime abstraction.

I also expanded validation cases for atomic publication failure, native-host-native ordering, effective-state getters, callback mirror misses, slow required callbacks, paused `dt`, replay, maintenance compaction, and high-fanout materialization.

yongkyuns:
Follow-up architecture review on the latest head: the original blockers are substantially addressed. The publication-before-lowering issue, authored-vs-effective distinction, mixed native/host ordering, callback read mirrors, slow required callback semantics, continuation barriers, revision taxonomy, high-fanout qualification, time domains, and roadmap ownership are now described well.

I still think there is one important architectural distinction to make before merge, plus a few smaller consequences of it:

1. **Distinguish effective driver writes from authored semantic mutations.**

   The document now correctly says the Semantic Scene owns authored/base state while Runtime owns effective state, but the callback/publication sections can still be read as if every host write becomes a semantic transaction that advances `SceneRevision`/`ExecutionRevision` and relowers.

   That is too strong for ordinary per-frame updater behavior. For example:

   ```text
   timeline interpolation
          ↓
   Python updater: mob.shift(...)
          ↓
   next frame
   ```

   The updater's transform should normally be an **effective driver write**, not automatically a mutation of authored/base transform. Otherwise host updaters partially collapse the authored/effective distinction and may trigger semantic relowering every frame.

   I suggest making the conceptual output of ordered evaluation explicit:

   ```text
   StagedUpdateBatch
       effective driver writes
           -> Runtime effective slots/domains
           -> FrameEpoch advances
           -> no SceneRevision necessarily

       authored semantic mutations
           -> Semantic Scene
           -> validation + lowering
           -> SceneRevision / possibly ExecutionRevision advances
   ```

   Both can still share one ordered overlay and one coherent publication barrier.

   Concretely, `m.shift(...)` from an active updater is normally an effective driver write; `scene.add(...)`, content replacement, subscription topology changes, etc. are semantic mutations where applicable. When an updater/driver is released, its effective value may be reconciled into authored/base state if compatibility semantics require persistence.

2. **Make `FrameEpoch` the publication unit and relate it explicitly to revisions.**

   The publication barrier should say something like:

   ```text
   FrameEpoch N
       references SceneRevision S
       references ExecutionRevision E
   ```

   A new frame may therefore publish:

   ```text
   FrameEpoch N+1
   SceneRevision S       // unchanged
   ExecutionRevision E   // unchanged
   ```

   Only a semantic commit creates a new `SceneRevision`; only a changed execution projection creates a new `ExecutionRevision`. This keeps ordinary timeline/reactive frames cheap and makes cache invalidation rules much clearer.

3. **Structural callbacks need transaction-local/provisional identity.**

   The callback overlay now promises read-after-write semantics across structural changes. This case should work within one callback epoch without half-publishing the new object:

   ```python
   line = Line(...)
   line.shift(...)
   scene.add(line)
   x = line.get_center()
   ```

   So the architecture should require some equivalent of transaction-local identity:

   ```text
   reserve provisional NodeId/handle
   AddNode(provisional)
   SetProperty(provisional)
   AddMember(... provisional ...)
   read provisional state through overlay
   ...
   atomic publish
   ```

   The exact implementation/name can differ, but the capability should be architectural if arbitrary structural host callbacks are first-class.

4. **Cross-context read miss should suspend, not implicitly retry arbitrary callbacks.**

   The current wording allows `resume/retry` on a partial callback-mirror miss. Retrying arbitrary Python is unsafe because the callback may already have performed external side effects (`print`, file/network I/O, mutating unrelated host state, etc.). Noon can roll back Noon engine state; it cannot roll back arbitrary host effects.

   Suggested contract:

   - complete mirror: execute normally;
   - partial mirror miss: suspend the **same callback invocation**, fetch the missing revision-pinned value, then resume;
   - if suspension is unavailable: deterministic error;
   - automatic restart/replay only for callbacks explicitly classified replay-safe/idempotent;
   - transaction atomicity applies to Noon semantic/runtime/resource state, not arbitrary external host side effects.

5. **Required-callback stalls need bounded discrete-event backpressure.**

   The slow-callback semantics are now correct: required callbacks are non-coalescible, authored time stalls, sampled input may coalesce, discrete events retain ordering. But if a required Python callback hangs while thousands of discrete events arrive, the architecture should not imply unbounded memory growth.

   Define a bounded queue/backpressure/overflow/session-fault policy for discrete events. Never silently collapse discrete events into sampled state, but also do not leave the queue unbounded.

6. **Terminology: avoid `serialized publication lane`.**

   Elsewhere `serialization` specifically means wire/JSON representation and is prohibited on normal in-process engine paths. In the roadmap text, use `single ordered publication lane` or `coherent publication lane` instead of `serialized publication lane` to avoid ambiguity.

Overall, I would **not reopen the previous review items**; those are now in good shape. The main blocker left is #1: effective driver writes vs authored semantic mutations. The other points make that distinction operational and keep the callback/session contract implementable at scale. After that focused edit, I think this architecture is in strong shape to approve/merge.

yongkyuns:
Addressed the follow-up review and restacked the PR onto current `master` (`cd05fe64...`). The branch is again one documentation-only commit: `2ecf4bcc...`.

Changes corresponding to the follow-up points:

1. **Effective driver writes vs authored semantic mutations** — added an explicit `StagedUpdateBatch` split. Ordinary per-frame updater writes such as `m.shift(...)` normally update Runtime effective state and may advance `FrameEpoch` without advancing `SceneRevision`/`ExecutionRevision` or triggering semantic relowering. Persistent/structural operations use the semantic mutation path; driver release/completion may reconcile effective state back into authored/base state when compatibility semantics require persistence.
2. **`FrameEpoch` as publication unit** — each coherent effective publication now references one `SceneRevision` and one `ExecutionRevision`; ordinary dynamic frames may advance `FrameEpoch` while both referenced revisions stay unchanged.
3. **Transaction-local/provisional identity** — structural callback staging now explicitly requires the ability to create/mutate/attach/read a provisional object before publication, with stable resolution on commit and invalidation on abort.
4. **Cross-context read misses** — a partial-mirror miss must suspend/fetch/resume the same callback invocation when supported; otherwise it fails deterministically. Automatic restart/replay is only allowed for explicitly replay-safe/idempotent callbacks, and Noon rollback is explicitly scoped to Noon engine state rather than arbitrary host side effects.
5. **Discrete-event backpressure** — ordered discrete-event storage is bounded. Backpressurable producers throttle; non-backpressurable overflow uses an explicit surfaced overflow/session-fault policy rather than silent coalescing/reordering/drop.
6. **Terminology** — replaced `serialized publication lane` with `single ordered/coherent publication lane` to avoid confusing execution ordering with wire serialization.

I also added matching correctness invariants, Phase C ownership, A6 ratchets, validation cases, and a 100k-mostly-static + one host-updater workload proving effective writes do not imply semantic relowering.

PR #1064, opened 2026-09-04
A1.6: compile semantic reactive graph into existing VM

yongkyuns:
## Scope

Continue A1.6 after #1062 and #1063 by feeding the canonical semantic reactive projection directly into Noon's existing `ReactiveProgram` / `ComputeProgram` path.

## Changes

- extend `SemanticExecutionLoweringOutput` with the existing native `ComputeProgram` execution form;
- derive the live execution-object domain and timeline-driver keys from the same lowered `CompiledScene` snapshot;
- compile the existing `SemanticReactiveProjection` through #1063's `ReactiveProgram::compile_for_execution_domain(...)` API and lower it into the existing compute VM;
- publish the staged semantic-to-execution identity index only after reactive program/VM compilation succeeds;
- retain the previous two-part decomposition for compatibility while exposing the complete three-part execution handoff;
- add focused coverage proving a semantic input update flows through the existing compute VM to the expected compiled object/property.

## Architecture

This adds no compiler, evaluator, graph, or runtime model. `SemanticReactiveProjection` still lowers to the existing `ReactiveGraphDefinition`; #1063's `ReactiveProgram` remains the single validated dependency/dirty-closure compiler; `ComputeProgram` remains the native runtime VM. This slice only connects the canonical semantic lowering entry to those existing authorities.

## Validation

The focused canonical-lowering test instantiates the produced `ComputeProgram`, updates the execution signal, and verifies the resulting opacity invalidation targets the same stable compiled object identity. Exact-head architecture, layer, Rust, web/WASM, and browser gates remain required before merge.

Refs #957 #959 #1060 #1062 #1063.

### DIFF
--- a/crates/noon-compile/src/semantic_lowering.rs
+++ b/crates/noon-compile/src/semantic_lowering.rs
@@ -4,6 +4,7 @@ mod membership;
 mod projection;
 mod reachability;
 mod reactive;
+mod reactive_target;
 
 pub use compiled_scene::*;
 pub use entrypoint::*;
--- a/crates/noon-compile/src/semantic_lowering/entrypoint.rs
+++ b/crates/noon-compile/src/semantic_lowering/entrypoint.rs
@@ -1,4 +1,4 @@
-use noon_core::SemanticStore;
+use noon_core::{ComputeProgram, ReactiveError, ReactiveProgram, SemanticStore};
 
 use crate::CompiledScene;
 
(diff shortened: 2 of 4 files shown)


yongkyuns:
#1063 has now merged as cd710a2493076d871b642040d9b3ad3ea8b42830 and owns the reactive compiler decoupling prerequisite. This draft's `ReactiveCompileTarget` / `SceneDefinition` implementation is also the shape rejected by the architecture ratchet, so it should not be restacked as-is. I’m closing this draft and carrying forward only its useful canonical-entry `ComputeProgram` integration/test on top of #1063's execution-domain API.

PR #1065, opened 2026-09-04
A1.6: compile semantic reactivity against execution target

yongkyuns:
## Scope

Continue A1.6 after #1062 by removing legacy `SceneDefinition` from native-reactive target validation on the canonical semantic lowering path.

## Changes

- add a narrow `ReactiveCompileTarget` query contract in `noon-core` for live object membership, exact timeline-driver conflicts, any timeline activity, and deterministic object iteration;
- keep `ReactiveProgram::compile(SceneDefinition, …)` as a compatibility wrapper but move all compiler logic into one `compile_for_target` body;
- implement the target contract for existing `CompiledScene` using stable object indices/channels/tombstones;
- make `lower_semantic_execution` validate its semantic reactive projection against its actual `CompiledScene`;
- immediately lower the validated graph into the existing `ComputeProgram`, so the canonical H2 handoff is execution-ready;
- preserve fail-atomic `SemanticExecutionIndex` publication through compute lowering;
- add semantic-only compiled-target tests without adding new `GeometryRef`/`ObjectDefinition` migration dependencies.

## Architecture

No new scene/runtime model, reactive graph, scheduler, evaluator, serializer, or patch vocabulary. The flow is:

`SemanticStore -> lower_semantic_execution -> { CompiledScene, SemanticReactiveProjection, ComputeProgram }`

The compatibility `SceneDefinition` compile entry delegates to the same compiler body; it is not a parallel lowerer.

A follow-on compatibility cleanup can migrate `SceneInstance::from_semantic` to the compiled target query path too, but this PR does not require runtime host changes.

## Validation

Exact-head architecture, dependency, Rust/Clippy, web, and browser/WebGPU gates are required before merge.

Refs #957 #969 #959

### DIFF
--- a/crates/noon-compile/src/semantic_lowering.rs
+++ b/crates/noon-compile/src/semantic_lowering.rs
@@ -4,6 +4,7 @@ mod membership;
 mod projection;
 mod reachability;
 mod reactive;
+mod reactive_target;
 
 pub use compiled_scene::*;
 pub use entrypoint::*;
--- a/crates/noon-compile/src/semantic_lowering/entrypoint.rs
+++ b/crates/noon-compile/src/semantic_lowering/entrypoint.rs
@@ -1,4 +1,4 @@
-use noon_core::SemanticStore;
+use noon_core::{ComputeProgram, ReactiveProgram, SemanticStore};
 
 use crate::CompiledScene;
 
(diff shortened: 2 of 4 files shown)


yongkyuns:
Closing in favor of #1063. The architecture ratchet correctly rejected this PR's new `impl ReactiveCompileTarget for SceneDefinition` migration reference. #1063 solves the same prerequisite more cleanly by moving the existing compiler body behind execution-domain object/driver inputs while keeping the legacy wrapper as a shrinking compatibility caller. The canonical semantic lowering -> existing ComputeProgram wiring will be layered on top of #1063 rather than introducing a competing target-query API.

PR #1074, opened 2026-09-04
A1.6: generalize semantic affine animation payload lowering

yongkyuns:
## Scope

Continue A1.6 after #1073 by replacing the provisional authored-source translation payload helper with the canonical activation-time affine lowering path over existing Position/Rotation/Scale execution tracks.

## Changes

- keep `animation_payload` as the single semantic animation payload-lowering owner;
- move affine activation lowering under `animation_payload::affine` rather than creating a peer top-level lowering authority;
- retire #1073's provisional translation-only public helper now that the activation path captures the correct live effective start state;
- lower `SemanticAnimationScheduleProjection` leaves into existing Position, Rotation, and Scale `TrackValues` only;
- require execution/session-owned effective `Transform2D` capture at activation time instead of substituting authored/base state;
- sample each execution object at most once per lowering call so one activation sees coherent effective start state;
- revalidate every scheduled leaf against its current authored `TransformTo` declaration before reading payload state;
- use keyed target/property driver tracking, preserving locality for large compositions without a quadratic scan;
- fail closed on content/style/painter-order/binding-topology/depth/lifecycle changes that this affine slice cannot represent;
- fail closed when an animated affine property already has a native-reactive driver or when one activation contains multiple drivers for the same target/property;
- keep execution `TrackId` assignment outside semantic declaration identity and validate the resulting `TrackDefinition` when that ID is attached.

## Why remove the #1073 helper

#1073 established the first typed payload seam and execution-owned track identity, but its single-channel helper used authored source transform state. That is not sufficient for an animation activated while the object is already transformed by prior timeline/reactive execution. The activation-level API introduced here explicitly requires the current effective runtime transform and therefore becomes the canonical payload entry. Retaining the provisional public helper would preserve a second path with weaker live-session semantics.

## Architecture

This PR introduces no scene model, animation graph, scheduler, evaluator, runtime/session, patch vocabulary, or execution identity allocator. Semantic animation declarations/scheduling remain authoritative upstream; `TrackDefinition`/`CompiledScene`/`SceneInstance` remain the execution timeline machinery.

Activation-time effective state is supplied by the execution/session owner, preserving the architecture distinction between authored animation intent and the live state from which an activation begins. Missing or non-finite effective state fails closed rather than falling back to stale authored state.

## Locality

Work is proportional to activated schedule leaves and affected affine channels. Effective state is captured once per execution object and duplicate-driver detection is keyed by execution object/property; no unrelated scene or timeline scan is introduced. The keyed map is not iterated, so hash iteration order cannot affect emitted track order.

## Validation

Focused coverage verifies affine Position/Rotation/Scale lowering, activation-time effective capture, one capture per execution object, stale-schedule rejection, keyed duplicate-driver rejection, native-reactive driver conflicts, execution-track validation, and missing effective-state failure. Exact-head Architecture/Layer ratchets, Rust/rustfmt/Clippy, web/WASM, browser authoring, deterministic playback, and WebGPU gates are required before merge.

Refs #957 #969 #1070 #1073.


yongkyuns:
Blocking architecture/correctness review after #1073 merged:

1. **Do not leave a second public payload-lowering path.** #1073 established `semantic_lowering::animation_payload` as the payload seam. This PR currently adds a sibling `animation_tracks` module with its own full validation/lowering implementation. Refactor this slice to extend/reuse the existing payload seam (or make the lower-level primitive internal) so there is one canonical payload path, not two peer APIs.

2. **Revalidate scheduled leaves against authored animation identity before lowering.** #1073 explicitly checks `leaf.animation` is still the authored `TransformTo { target, target_state }`. The current affine lowerer only trusts the schedule projection and reads the endpoint objects. A stale schedule must fail if the animation was removed/replaced or no longer matches those endpoints; do not silently lower stale scheduling data.

3. **Fix the duplicate-driver locality claim.** `driven.iter().find(...)` makes a large activation O(number_of_emitted_tracks^2) in the worst case. Use a keyed lookup such as `HashMap<(ObjectId, SemanticObjectProperty), ...>` / equivalent bounded index so duplicate-driver detection stays proportional to activated leaves/properties.

4. **Keep the activation-time effective-state idea.** Supplying/capturing the current effective `Transform2D` from the session/runtime boundary is the correct continuation; do not regress to authored/base state as the animation start. Sampling once per execution object is also the right coherence rule.

5. **Prefer validated track publication.** `with_track_id()` currently constructs a `TrackDefinition` without running the existing timeline validator. Since TrackId assignment is the last execution-boundary step, either return a validated `Result<TrackDefinition, TimelineError>` there or otherwise prove validation happens before the definition can enter `CompiledScene`/runtime.

Once these are addressed, the affine Position/Rotation/Scale expansion is the right next A1.6 slice. Please keep lifecycle/style/content and same-property composition semantics fail-closed as the PR already does.

yongkyuns:
Addressed the blocking architecture/correctness review on exact head `a32eeef1466b4008576389605c32709706188e9e`:

1. Removed the #1073 translation-only public lowerer. `animation_payload.rs` is now only the module seam and `animation_payload/affine.rs` owns the one supported semantic animation payload path.
2. Every scheduled leaf is revalidated against the current authored `TransformTo { target, target_state }` declaration before payload lowering.
3. Activation-local effective-state capture and duplicate-driver detection use keyed `HashMap` indexes; no O(n²) emitted-track scan remains.
4. Animation starts are supplied from current effective runtime/session `Transform2D` and sampled at most once per execution object; there is no authored/base-state fallback.
5. `with_track_id()` validates the resulting existing `TrackDefinition` before it can cross into the compiled/runtime timeline.

Unsupported content/style/painter/depth/lifecycle semantics and unresolved same-property reactive/composition arbitration remain fail-closed. No `ObjectSnapshot`, full-object track payload, second scheduler, patch vocabulary, or execution identity allocator was introduced.

Exact-head Architecture Ratchets, Layer Dependency Ratchet, rustfmt, Clippy, Rust unit tests, web/WASM build, deterministic playback, Manim-style authoring, and primary WebGPU all pass.

PR #1077, opened 2026-09-04
A2/A6: add direct native Rust viewport host

yongkyuns:
## Scope

Continue #969 A2/A6.2 after #1071 established the common typed `ExecutionSession` and #1075 proved the direct Rust/WASM browser host. Add the first native platform host that owns OS/window/GPU presentation lifecycle while consuming the same typed execution/runtime/renderer path.

This is the native-host foundation slice. It deliberately does not introduce a new runtime, renderer abstraction, semantic camera model, or generic application framework.

## Changes

- add a dedicated `noon-native` platform crate because native window/event-loop dependencies must remain outside `noon` and `noon-render-wgpu`;
- expose only `run(ExecutionSession)` and `run_with_config(...)` as the initial host surface;
- own winit window/event-loop lifecycle, wgpu instance/surface/adapter/device/queue/configuration, resize, realtime clock, frame acquisition, submission, and presentation in the platform crate;
- drive `ExecutionSession::advance_to`, consume `FrameChanges`, and feed the existing `FramePreparer`/`GpuRenderer` incrementally;
- preserve invalidation across transient surface acquisition failures with one explicit full-frame retry rather than losing consumed runtime changes;
- derive the initial native viewport from the existing normalized `Camera2DState` default contract; semantic 2D camera ownership remains with B6/#89;
- add a Rust-only representative example: `SemanticStore -> ExecutionSession -> noon_native::run`;
- ratchet dependency direction so engine/authoring crates cannot depend upward on `noon-native`.

## Architecture

```text
Rust SemanticStore
    -> ExecutionSession
    -> FrameState / FrameChanges
    -> noon-native platform lifecycle
    -> existing FramePreparer / GpuRenderer
    -> native wgpu Surface
```

`noon-native` owns platform lifecycle only. Semantic state/lowering/runtime stay below it; `noon-render-wgpu` remains independent of window/event-loop ownership. There is no serialization or transport mirror in this path.

## Validation

Required before merge:

- architecture and layer-dependency ratchets;
- Rust format and Clippy across all targets/features;
- Rust unit tests, including native viewport/camera normalization helpers;
- representative native example compiles in the all-target Clippy lane;
- existing web/WASM/browser gates remain green, proving the new native host does not contaminate browser/platform boundaries.

A dedicated virtual-display native rendering smoke and native reactive input ingress remain follow-up A2/A6.2 slices before the full native-host exit ratchet.

Refs #969 #1071 #1075 #89.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "crates/noon-compile",
     "crates/noon-runtime",
     "crates/noon-render-wgpu",
+    "crates/noon-native",
     "crates/noon-web",
     "crates/noon-typst",
     "crates/noon-text-native",
--- a/crates/noon-native/Cargo.toml
+++ b/crates/noon-native/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "noon-native"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+description = "Native window host for typed Noon execution sessions"
+
+[dependencies]
+noon = { path = "../noon" }
+noon-core = { path = "../noon-core" }
+noon-render-wgpu = { path = "../noon-render-wgpu", default-features = false, features = ["native"] }
+pollster = "0.4"
+wgpu = { version = "29.0.3", default-features = false, features = ["std"] }
+winit = "0.30.13"
--- a/crates/noon-native/examples/direct_semantic.rs
+++ b/crates/noon-native/examples/direct_semantic.rs
@@ -1,0 +1,14 @@
+use noon::ExecutionSession;
+use noon_core::{SemanticObjectState, SemanticStore, StoredGeometry};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut store = SemanticStore::new();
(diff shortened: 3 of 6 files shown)


yongkyuns:
Architecture alignment note after the `docs/architecture.md`/roadmap streamline:

The crate/platform ownership in this PR is aligned, but the current host loop still needs one correction before merge: `ControlFlow::Poll` + unconditional `request_redraw()` makes a clean/static/paused session busy-tick forever.

The updated #953/#969/#961 contract is now explicit: active realtime work may schedule presentation frames, but a clean session must settle and wake from the next runtime deadline, input/event, callback/resource completion, authored mutation, resize/expose, etc.

Please make this PR either:

1. consume an existing/minimal `ExecutionSession` active-work/next-wake contract and use `Wait`/`WaitUntil`, or
2. add the narrow session wake/deadline surface needed by both native and later web host integration.

Do not put scheduling truth into `noon-native`; the session/runtime remains authoritative for whether/when execution needs another wake. A full native input/render smoke can remain a follow-up, but unconditional polling should not become the foundation host contract.

PR #1082, opened 2026-09-04
A2/A6: bridge semantic native inputs into ExecutionSession

yongkyuns:
## Scope

Continue #969 A2/A6.2 after #1079 proved real native presentation. This slice migrates the existing normalized native-input/reactive router onto the current `SemanticStore -> ExecutionSession -> SceneInstance` path without adding platform event handling yet.

## Changes

- add `SemanticNativeInputDefinition` keyed by authoritative `SemanticNodeId` rather than execution `SignalId`;
- validate native source kinds against semantic input signal kinds and reject derived/duplicate/invalid bindings;
- lower semantic native-input bindings through the existing `SemanticReactiveProjection` into the existing `NativeInputDefinition` execution vocabulary;
- refactor `NativeInputRouter` so one dispatch implementation serves both legacy `TimedSceneInstance` and current `SceneInstance`;
- construct `ExecutionSession` with semantic native inputs under the same immutable lowering snapshot, then dispatch normalized sampled state / discrete events through that shared router;
- keep normalized `NativeInputValue -> ReactiveValue` conversion inside core execution vocabulary.

## Architecture

```text
SemanticStore + SemanticNativeInputDefinition
  -> canonical semantic execution lowering
  -> SemanticReactiveProjection
  -> NativeInputDefinition (execution SignalId, internal)
  -> NativeInputRouter
  -> SceneInstance native-reactive VM
```

No second reactive graph/evaluator is introduced. Execution signal identity remains private to lowering/runtime. No winit/browser types enter core, compile, runtime, or `ExecutionSession`. Scene/reactive/input identity is resolved under one semantic snapshot; there is no post-construction configuration path that can combine a mutated store with a stale execution projection.

## Validation

Before merge:

- semantic declaration validation and lowering tests;
- current-session sampled-state coalescing and repeated discrete-event tests;
- legacy native-input router tests remain green;
- architecture/layer ratchets, rustfmt/Clippy, Rust tests, web/WASM/browser gates green on exact head.

## Boundary

This PR does **not** implement the C2/#69 bounded burst queue/backpressure policy and does not revive the closed pre-Phase-A #925 implementation. It also does not map winit events yet; that is the following A2/A6.2 host-adapter slice once this semantic/session bridge lands.

Refs #969 #69 #1079.

PR #1083, opened 2026-09-04
A2/A6: route native platform input through typed execution

yongkyuns:
## Scope

Continue #969 A2/A6.2 after #1079 proved real native surface presentation. Add the first production native input ingress onto the same typed `SemanticStore -> ExecutionSession -> SceneInstance` path.

This slice does **not** add a second input model, runtime router, transport queue, or winit-specific semantic vocabulary. It reuses Noon's existing language-neutral `NativeStateSource` / `NativeEventSource` contracts.

## Changes

- let authoritative `SemanticSignalState` own at most one optional normalized native state/event source declaration;
- validate native source kind against semantic signal kind and require externally driven signals to remain authored input signals;
- derive source -> existing execution `SignalId` routing while lowering the already-active semantic reactive dependency closure, with no separate total-scene input scan;
- add `ExecutionSession` sampled-state delivery and explicitly ordered discrete-event delivery without exposing VM signal identity;
- keep discrete events non-coalescing and reject duplicate/out-of-order ingress sequence numbers before mutation;
- collect unambiguous native platform inputs in `noon-native`: logical viewport size, physical-key state/press/release, and pointer-button state/down/up;
- preserve cross-platform key/button source identity (`KeyA`, `Space`, web-compatible button numbers);
- leave pointer-position world-coordinate normalization plus wheel/gesture/control breadth to #69 rather than guessing around camera/DPI semantics in the host boundary.

## Architecture

```text
SemanticSignalState native-input declaration
  -> semantic reactive lowering
  -> private source -> execution SignalId routes
  -> ExecutionSession
  -> existing reactive VM / FrameChanges

winit WindowEvent
  -> normalized NativeStateSource / NativeEventSource
  -> ExecutionSession
```

Semantic identity remains authoritative; `SignalId` stays below the session boundary. `noon-native` owns only platform collection/normalization and event sequence assignment.

## Coverage

- O(1), one-slot semantic declaration mutation and kind validation;
- derived-signal/native-driver conflicts fail before mutation;
- route lowering is limited to the active reactive dependency closure;
- sampled state reaches renderer-facing state through the session;
- invalid sampled values fail before runtime mutation;
- discrete events are ordered and never coalesced;
- native key/button normalization matches the existing cross-platform vocabulary;
- native app dispatch helpers exercise the real `ExecutionSession` input path.

Refs #969 #69 #1079.

### DIFF
--- a/crates/noon-compile/src/semantic_lowering/reactive.rs
+++ b/crates/noon-compile/src/semantic_lowering/reactive.rs
@@ -1,7 +1,8 @@
 use std::collections::{HashMap, HashSet};
 
 use noon_core::{
-    Property, ReactiveError, ReactiveExpr, ReactiveGraphDefinition, ReactiveValue, SemanticNodeId,
+    NativeEventSource, NativeStateSource, Property, ReactiveError, ReactiveExpr,
+    ReactiveGraphDefinition, ReactiveValue, SemanticNativeInputSource, SemanticNodeId,
     SemanticObjectProperty, SemanticSignalError, SemanticSignalExpr, SemanticSignalSource,
     SemanticSignalValue, SemanticStore, SignalDefinition, SignalId, SignalSource,
 };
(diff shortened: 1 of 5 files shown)


yongkyuns:
Architecture boundary for this slice: immediate ordered delivery is fine and this PR does not need to invent a queue just to satisfy C2. However, do not add unbounded buffering here or later. If discrete occurrences can accumulate behind callback/session barriers, that storage must use #69's bounded ordered `InputSequence` + explicit backpressure/overflow policy. Sampled state may coalesce only by exact source. The signal-owned declaration + active-reactive-closure routing in this PR is the preferred semantic path; #1082 has been closed as superseded.

PR #1084, opened 2026-09-04
Restore retained demo animation geometry parity

yongkyuns:
## Summary

- preserve `RetainedFrameState::render_geometry()` when the mixed retained adapter builds renderer scratch geometry, so transient transform geometry reaches `FramePreparer`
- stop caching pre-preparation primitive kind in painter-order sources; recover the primitive from the geometry actually prepared instead
- keep Create/Uncreate reveal-head lines immediately above their prepared path body
- add focused renderer regressions for partial analytic `Create` and semantic-circle + transient morph-path transform state

## Regression

Normal Python playground authoring now routes through canonical retained execution. Two latent adapter parity holes became visible there:

1. heterogeneous transforms (for example circle → square) carry their in-progress path in `RetainedFrameState::render_geometries`, but the retained scratch adapter rebuilt from semantic `object.content` and discarded that path;
2. analytic Create/Uncreate may enter `FramePreparer` as Circle/Rectangle/Line but be prepared as a temporary Path, while mixed painter-order reconstruction still looked up the pre-preparation primitive kind and therefore emitted no draw item.

This fixes the shared retained-frame → renderer-scratch boundary rather than special-casing the demo authoring APIs.

### DIFF
--- a/crates/noon-render-wgpu/src/gpu/retained_family_draw_border_prepare.rs
+++ b/crates/noon-render-wgpu/src/gpu/retained_family_draw_border_prepare.rs
@@ -204,7 +204,6 @@ impl RetainedFramePreparer {
                 SourceItem::Geometry {
                     object_id,
                     scratch_id,
-                    kind,
                 } => {
                     let object_index = frame
                         .retained
@@ -234,7 +233,6 @@ impl RetainedFramePreparer {
                     self.sources.push(SourceItem::Geometry {
                         object_id,
                         scratch_id,
-                        kind,
                     });
                 }
                 SourceItem::FastGlyphRun {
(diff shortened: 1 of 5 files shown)


yongkyuns:
Architecture pass: this fix is aligned with the merged authored-vs-effective contract. Using `RetainedFrameState::render_geometry()` preserves the Runtime-provided transient/effective geometry instead of falling back to authored `object.content`, and deriving painter-order primitive kind from what was actually prepared avoids treating pre-preparation semantic/analytic kind as renderer truth. I don't see a new semantic authority or per-frame semantic mutation path here. Keep the new regressions focused on transient transform geometry + analytic Create/Uncreate preparation so this boundary stays pinned.

Issue #1087, opened 2026-09-04
[Roadmap A1/A6] Add distinct scene, execution, and frame publication revisions


yongkyuns:
## Roadmap

Phase A publication/revision infrastructure under #957/#969/#961. Architecture authority: `docs/architecture.md`.

## Problem

#1100 landed distinct `SceneRevision`, `ExecutionRevision`, and `FrameEpoch` types and initial propagation. This issue remains open for complete coherent publication across semantic/execution/runtime boundaries, no-op and failure behavior, stale callback/resource/spatial result rejection, and renderer/resource context. The existence of typed counters alone does not prove those integration contracts.

This issue owns the minimum shared revision/publication metadata. It must not create another scene/runtime authority or force a semantic revision for every effective frame update.

## Required domains

Exact type names may differ, but responsibilities must remain distinct:

- **semantic identity generation** — stale `SemanticNodeId` protection; not a revision clock;
- **SceneRevision** — changes only when a new authored Semantic Scene state is successfully committed/published;
- **ExecutionRevision** — changes only when the derived executable projection/schedule/dependency/resource mapping changes and is coherently published;
- **FrameEpoch** — advances when a new coherent effective Runtime state is published, including pure timeline/reactive/host/editor updates where scene/execution revisions may remain unchanged;
- **ResourceVersion** — immutable content/resource identity/version;
- **InputSequence** — ordered ingress occurrences;
- **SubmissionSerial/fence** — GPU in-flight lifetime/retirement.

Do not compare or substitute these domains for one another.

## Publication model

A published frame carries enough context to state conceptually:

```text
FrameEpoch F
  SceneRevision S
  ExecutionRevision E
  effective runtime/camera/spatial state
  resource versions referenced by that frame
```

A pure effective update may publish:

```text
FrameEpoch F + 1
SceneRevision S       // unchanged
ExecutionRevision E   // unchanged
```

A persistent semantic edit advances `SceneRevision` only after validation and any required preparation/lowering needed for a coherent live publication. `ExecutionRevision` advances only if the derived execution projection actually changes.

### 1. Typed revision values
Introduce small strongly typed revision/epoch values with deterministic monotonic behavior and no accidental cross-domain comparison.

### 2. Semantic commit ownership
Successful persistent semantic transaction publication advances the authored scene revision exactly once. Failed/no-op transactions do not mint a misleading new revision.

### 3. Execution publication ownership
[...]

PR #1105, opened 2026-09-05
Publish coherent local transactions once and keep preflight sparse

yongkyuns:
A mixed semantic/runtime transaction could publish multiple revisions, repeated value writes could publish false progress, camera reads scanned every frame object, and transaction preflight cloned the entire object/track inventory. Make successful transaction publication coherent and keep local validation/query work local.

- Publish semantic revision once at the transaction boundary; direct construction bookkeeping no longer mints revisions. Runtime direct patches and aggregate slotted/host batches publish execution/frame context once, while failed and unchanged work preserve it. Coalesce final value writes only after validating the full ordered transaction, and preserve context across compaction.
- Replace full-inventory preflight with sparse object/track overlays and affected-channel traversal. Preserve sequential validity for moved tracks, removal/recreation, presence windows, and overwritten invalid patches.
- Resolve effective objects/camera through the compiled index and reject animation activation against a changed semantic revision before mutation.

Independent review passed for code quality, architecture, correctness and locality. Regression coverage includes mixed publication/failure/no-op batches, compaction, stale activation, 2,000 generated preflight transactions against sequential application, 100,000-object locality, and linear metadata-visit accounting for independent presence edits.

Scope: advances #1087/#961/#969; does not close them. Net-canceling structural/track edits, atomic multi-source input fanout, semantic store lineage, live semantic attach/detach/relowering, and required callback/resource publication remain with their existing owners. The revision guard compares scene revision, not store identity.

Validation: the integrated full local gate passed workspace Rust tests, web JavaScript checks/tests, 172 Python tests, and release WASM build/package validation. The final fast gate passed formatting, workspace/all-target checks, clippy and 1,160 library tests. Fresh debug WASM passed the actual WebGPU and WebGL2 renderer smokes (11 fixtures at four checkpoints, including the direct typed Rust/WASM probe); the release Manim compatibility smoke also passed.

Final fixture-only corrections passed all 7 compiler preflight tests (including 2,000 model sequences), 23 execution-slot tests, affected all-target clippy, formatting, and the architecture ratchet/self-tests. Fixtures now seed scenes through canonical semantic lowering. The existing CreateObject payload remains only in an explicitly cfg(test) fixture, with a narrow tested ratchet exception and #959 deletion ownership documented in the architecture. Independent review passed this final change too.

The earlier CI architecture failure was the missing rg dependency fixed in #1106. This branch includes that fix. Slow CI may finish after merge; local validation is the merge evidence.


yongkyuns:
Final combined local validation passed on merged master `6ab92bfd1c32b18f7eebbcf07f91d1b2f45e9fb6`: `scripts/check.sh fast` (formatting, all-target/all-feature workspace checking, clippy, 1160 Rust library tests), fresh debug WASM build/package validation, WebGPU and WebGL2 renderer smokes (11 fixtures × 4 checkpoints, including the direct Rust/WASM probe), and the complete Manim compatibility smoke. The final tree contains both #1105 and the wgpu 30 migration #1078. Architecture guard/self-tests also passed. Independent review covered every implementation change.

This supplements the earlier full workspace/web/release WASM validation. CI can continue separately; these results come from the combined local checkout.

Issue #1088, opened 2026-09-04
[Roadmap A2/A6] Add logical segment completion and authoring continuation


yongkyuns:
## Roadmap

Phase A2/A6 execution-session continuation case under #969, consumed by #958/#61. Architecture authority: `docs/architecture.md`.

## Problem

#1094/#1098 landed a target-neutral authored-time segment boundary and return the canonical segment from semantic activation. This is the deterministic time foundation. The full shared logical `play()`/`wait()` completion contract remains open: required ordered callbacks/resources, coherent publication, and Rust/Python/browser continuation consumers must be integrated before general authoring can resume safely.

A segment barrier is not just “time reached” and is not a GPU fence. It must include required ordered callback/updater results and the coherent publication they produce, while remaining target-appropriate: Python may suspend/block an authoring worker, native Rust may drive/await, and browser/WASM must yield rather than block the event loop.

## Completion contract

A segment is complete only when:
- the requested authored/simulation-time boundary has been reached;
- every required ordered updater/host-callback dependency through that boundary has completed or produced a terminal error;
- resulting effective driver writes and persistent semantic/resource changes have passed required validation/preparation/lowering;
- the resulting coherent Runtime state has been published at a `FrameEpoch` and is queryable by the resumed continuation;
- any required session-level failure/cancellation state is resolved.

Completion normally does **not** wait for GPU submission retirement/fence completion or physical freeing of old resource versions.

### 1. Target-neutral segment state
Represent the pending/completed/failed state of one authored segment in the common execution/session layer. Do not make Python, native host or browser host the scheduler authority.

### 2. Rust drive/await surface
Provide an idiomatic Rust API that can drive or await a segment while preserving native input/reactive/session work that is allowed during playback.

### 3. Python continuation
Python `Scene.play()` / `wait()` may expose synchronous source semantics by suspending/blocking the authoring worker, but the engine/runtime continues independently. Resumption occurs only after the shared segment contract says complete.

### 4. Browser/WASM continuation
Single-context Rust/WASM and future JS adapters yield/await without blocking browser frame/input servicing. JavaScript bootstrap must not become the segment scheduler.

### 5. Required callbacks/backpressure
A required ordered callback stalls authored-time progress at its dependency barrier. Presentation may retain the last coherent frame and unrelated sampled input may continue coalescing. Do not mark the segment complete merely because wall time passed.

### 6. Effective final queries
[...]

PR #1123, opened 2026-09-05
Reconcile live affine completion into authored state

yongkyuns:
Completing a live affine animation now persists its exact semantic endpoint and releases its timeline ownership in one atomic publication. A later authored setter therefore survives the next frame, and a subsequent animation captures that updated effective source. Rust and Python share this completion operation and executable examples demonstrating the full sequence.

Advances #1088, #969 and #958; stacked on #1122.

- ExecutionSession owns completion tokens and the pending affected-track set. Tokens reject foreign runtime incarnations and stale semantic-store revisions; repeated completion is an idempotent no-op, including after later progression. A zero-length wait cannot complete before its required callback publishes.
- Completion applies authored endpoint values and typed track-release metadata through the existing semantic mutation transaction. Continuing callback-owned effective domains are carried from the exact coherent endpoint receipt; unreceipted or ambiguous driver release fails before publication.
- The runtime retains original track values/timing for deterministic evaluation inside the original interval, while released channels stop overwriting subsequent authored setters. This does not promise historical replay of unrecorded external authored mutations.
- Rust exposes `LiveSession::complete_segment`; Python `LiveExecution.complete` calls the same session. Python ordinary center/width/height queries now read coherent effective layout while the authoring context owns the runtime. Queries preserve shared layout semantics, exclude renderer stroke expansion, and reject unsupported geometry drivers or a transferred runtime instead of returning stale state.
- Explicit live advancement/completion holds presentation at the coherent runtime time. A new play/wait resumes from that frame; an undersized transport duration rejects before changing the clock or runtime.
- A returned presentation runtime may become stale during a later authoring run. Ordinary getters then read authored layout, and only an explicit run/lease boundary replaces that dormant runtime; active live-session reads continue rejecting unpublished edits.
- Work is proportional to affected tracks, properties and callback domains. There is no whole-scene completion snapshot, second timeline, new semantic identity space, or native/direct-WASM transport boundary. No new migration adapter is introduced; existing legacy frontend and diagnostic seams remain deletion-owned by #959.

The first completion subset excludes mapped/instant tracks, ambiguous overlapping driver release, and arbitrary unrecorded mutation replay. Phase A and the broader roadmap remain open.

Validation:

- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev bash scripts/check.sh full` — passed at3fac88c3, including full Rust tests, Node checks, 174 Python tests, playground tests and a 165296416-byte WASM package with 93 API contracts.
- `bash scripts/architecture-ratchet.sh origin/master` — passed passed with the completion guards; the final returned-lease change introduces no new migration boundary.
- `node scripts/browser-smoke.mjs` and `NOON_BROWSER_SMOKE_BACKEND=webgl node scripts/browser-smoke.mjs` — direct Rust/WASM completion proof passed on both backends before the final presentation-clock guard; direct host code did not change afterward.
- `node scripts/shared-authoring-smoke.mjs` — passed on the final source, including completion pixels, callbacks and persisted Scene reuse (transferable and shared transport).

After the full gate, a focused returned-player lifecycle fix was qualified with `cargo test -p noon-web --all-features --lib canonical_authoring_scene::tests` (14 passed), `cargo clippy -p noon-web --all-features --all-targets -- -D warnings`, and `NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` (165297233-byte package, 93 contracts). Cargo commands used the shared target and `CARGO_INCREMENTAL=0`. The final browser proof uses that rebuilt package.


yongkyuns:

PR #1124, opened 2026-09-05
Use canonical scalar signals and timelines across Rust and Python

yongkyuns:
Canonical scalar ValueTracker authoring now uses scene-global semantic signal identity, shared Rust track declarations and the existing ExecutionSession. Rust/native, direct Rust/WASM and Python examples exercise the same bound position at midpoint and endpoint, without a timed scene document or frontend interpolation.

Advances #61 and #959; stack on the shared affine completion slice (#1088).

- SemanticStore owns finite scalar tracks, continuity and driver-conflict validation. The compiler lowers reachable signal tracks into the execution plan; the session uses its existing frame time and a derived active/event index.
- Reactive input batches stage all inputs and evaluate the dirty dependency closure before publishing once. A late failure rolls back signal values and queued work; foreign/stale prepared batches are rejected. Native multi-target occurrences reuse this atomic path.
- Scene/ValueTracker handles contain only the store and semantic identity. WASM/Python wrappers call the same declaration/query/binding operations. Explicit live evaluation holds the published time for presentation.
- Deletes the obsolete web reactive authoring facade, JavaScript value cache, and scalar JSON parity case. Canonical paired examples replace that evidence. Existing native/host duplicate runtime consumers remain explicit #959 deletion work; no new migration allowance is introduced.
- Forward work follows active or crossed tracks and the dirty reactive closure. Seek may visit tracked groups. Clean/unchanged native input avoids reactive recomputation.

Current subset: scalar deterministic positive-duration non-overlapping tracks and position binding. A timeline-owned signal rejects direct writes, including after its authored endpoint. Mixed legacy/canonical timing and scalar reads during an opaque callback fail explicitly; they do not select a fallback engine. Native source facade/callback breadth and removal of the remaining duplicate runtime are follow-up work.

Validation:

- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh full`: architecture, formatting, workspace all-target/all-feature check and Clippy, all workspace tests and WASM build passed. The final package assertion referenced the deleted facade; removed that obsolete assertion and resumed with `node scripts/check-web-package.mjs` (165,705,152-byte WASM, 97 public contracts).
- Separate Node/Python browser preflight passed (174 Python tests).
- `node scripts/shared-authoring-smoke.mjs` passed actual Python worker scalar midpoint/endpoint pixels, mixed-timing rejection, callbacks and retained scene reuse over shared and transferable transport.
- `node scripts/browser-smoke.mjs` passed direct typed Rust/WASM WebGPU proofs, including the paired tracker scene.
- `NOON_BROWSER_SMOKE_BACKEND=webgl node scripts/browser-smoke.mjs` passed the same direct Rust/WASM proofs on WebGL2.

Qualification also caught and fixed explicit seek semantics when scalar timelines coexist with completed/released affine tracks. Regression tests protect restoration of the authored base after historical playback and preserve the shared session's evaluate/seek/advance distinction.


yongkyuns:

PR #1127, opened 2026-09-05
Route browser DOM input into canonical execution sessions

yongkyuns:
Browser DOM input now drives a canonical semantic execution session. The native-input smoke authors the paired Python native-signal scene and verifies real DOM key, pointer, control and click effects; it no longer constructs a timed scene document, ReactiveCanvasPlayer or a second native router.

Advances #959 and #969. Browser glue owns DOM normalization/listener lifecycle and bounded admission; Rust owns current-camera pointer conversion on the direct path, source routing, signal identity, evaluation and publication.

- Direct single-context Rust/WASM methods accept primitive input and construct typed Rust source/value enums in ExecutionCanvas. No JSON engine boundary is added.
- The genuine worker adapter forwards envelopes immediately in order through existing semantic control APIs, with bounded in-flight admission and surfaced errors. It adds no event queue, interpolation or semantic state mirror.
- Split-worker pointer-to-scene mapping is supplied by platform viewport integration, since the DOM thread does not own the renderer camera.

Validation:

- `node --test web/native-inputs.test.mjs`: 7/7 normalization, ordering, bounded admission, error and host-surface tests passed; syntax and diff checks passed.
- Integrated with ordinary animation authoring: `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh full` completed architecture, format/check/Clippy and all Rust tests/doctests. A WASM-only export-impl placement error was fixed by moving only the six JS methods into the existing exported impl.
- Resumed `bash scripts/check.sh web` with the same environment and `NOON_SKIP_WEB_PREFLIGHT=1`; WASM built, then the unrelated new example declaration assertion was corrected and `node scripts/check-web-package.mjs` passed (166,150,546-byte module, 117 public contracts).
- `node scripts/native-input-smoke.mjs` passed actual Python authoring → canonical semantic worker → renderer with real DOM key, pointer, click and control input.
- Combined Node preflight and final Python174/174 suite passed. Exact standalone branch CI remains required before merge.

The integration tree contains additional ordinary-animation changes; this branch contains only the DOM input consumer cutover and its direct export fix.

PR #1126, opened 2026-09-05
Delete the unused duplicate Rust reactive authoring engine

yongkyuns:
Deletes the unused duplicate Rust reactive authoring engine and its public/legacy exports after canonical scalar/native authoring qualification (#1124/#1125). Canonical Scene, ValueTracker and native signal handles remain store-backed and continue through ExecutionSession.

Advances #959. Removes ReactiveScene, ReactiveTimelineScene, their old ValueTracker/vector/bool handles and their independent graph/timeline authoring modules: 663 lines removed across five files. An rg consumer inventory found no callers outside the removed implementation/tests/reexports. No compatibility aliases or adapters replace them.

The still-reachable timed/native/callback runtime and browser players remain explicit consumer-migration work; this change does not claim full duplicate-engine removal.

Validation:

- `bash scripts/renderer-host-boundary-ratchet.sh` and `git diff --check` passed on this deletion branch.
- The deletion is merged into the ordinary-animation integration tree at 2d972901. `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh full` completed architecture/formatting, workspace all-target/all-feature check/Clippy, all Rust tests and doctests with the deletion applied. The unrelated new browser export error was fixed in the integration tree and its WASM package then built successfully (117 contracts).
- Final integrated Python suite passed174/174; canonical scalar/native paired examples are already qualified in #1124/#1125.

This branch contains only the deletion and is stacked directly on #1125. Required CI validates the exact standalone branch before merge.

PR #1128, opened 2026-09-06
Run ordinary affine play and wait through the shared Rust session

yongkyuns:
Ordinary Python leaf-affine `Scene.play` and `Scene.wait` now resume one retained Rust execution session. A play declares and activates semantic intent atomically, advances to the endpoint, then reconciles completion before the next statement. A later target copy and relative edits publish through that same session, so the second animation captures the current authored state without a runtime reset.

Advances #958, #959, #969 and #1088. The semantic store owns intent and target identity; the compiler shares one affine payload/channel lowerer between predeclared and newly declared animations; the session owns activation, completion, time and atomic publication. Python handles argument adaptation and opaque-handle dispatch.

- Adds atomic post-bootstrap TransformTo declaration/activation through the existing semantic/execution transaction, without reserving a second identity or cloning the full scene/runtime.
- Creates detached continuation targets through shared semantic transactions. Detached target edits remain local and generate no renderer frame work.
- Routes ordinary one-object animate/exact Transform and waits to canonical barriers; Rust owns relative shift/scale/rotation and effective queries. No Python interpolation, track creation, clock synchronization or rollback cache is added.
- Paired `ordinary_affine_play` examples run native Rust, direct Rust/WASM and Python: move to(2,-1) at t2, wait to t3, shared shift to(3,-1), create/edit target after bootstrap, then Transform to(5,-1) at t4. The renderer receives the same completed runtime.

The canonical subset is serial leaf translation/scale/center rotation with linear timing. Unsupported payloads/options before activation retain the existing #959 legacy boundary; an active/transferred canonical session cannot fall back. Required host callbacks in this synchronous wrapper, legacy timing prefixes, and a pre-execution scalar timing prefix reject explicitly. Broader yielding continuation and migration of remaining legacy consumers stay with #969/#959; no compatibility adapter is introduced.

Read-only capability preflight lives in the shared Rust compiler and Scene API. The typed WASM result distinguishes an unsupported payload from invalid/stale/foreign handles; Python does not duplicate style policy or swallow errors. This also preserves already-materialized compatibility scenes on their existing route.

Returned execution leases remain distinct from active authoring sessions. Python dispatch queries Rust ownership, so edits after a worker returns a scene are legal authored changes and the next explicit run refreshes the dormant runtime. Active or transferred sessions cannot silently mutate a stale store.

Validation:
- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh full`: architecture, format, workspace check, Clippy, all Rust tests and doctests passed at2d972901. The web phase exposed misplaced wasm-bindgen exports, fixed in #1127; its package assertion was then corrected.
- Final focused ownership fix: `cargo test -p noon-web canonical_authoring_scene --lib` (20 passed), `cargo clippy -p noon-web --all-targets --all-features -- -D warnings` passed.
- Resumed `bash scripts/check.sh web` with the same environment: module166151973 bytes /118 contracts passed atabfbaf18.
- Node web preflight and `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s web/python -p 'test_*.py'` (174 passed).
- Actual `node scripts/shared-authoring-smoke.mjs` passed on final source, including both worker transports, the paired ordinary fixture and persisted Scene reuse.
- Actual `node scripts/native-input-smoke.mjs`, `node scripts/browser-smoke.mjs`, and `NOON_BROWSER_SMOKE_BACKEND=webgl node scripts/browser-smoke.mjs` passed. The final ownership/dispatch fixes do not change those direct renderer paths, so those unchanged suites were not repeated.

CI follow-up qualification at fd123ea4: `cargo test -p noon --lib ordinary_transform_preflight` and `cargo clippy -p noon -p noon-web --all-targets --all-features -- -D warnings` passed. Refreshed dev `bash scripts/check.sh web` validated a 166465163-byte module and 119 contracts, including the boolean capability export. Actual `node scripts/manim-compat-smoke.mjs` and `node scripts/shared-authoring-smoke.mjs` both passed after that rebuild. `node scripts/playground-stress-edit-smoke.mjs` also passed the reported recovery scenario (rows5→7→20) during dispatch repair. Debug-only direct proof exports are now checked only for debug packages, fixing release CI. Required GitHub checks still qualify the exact pushed head.

Stacked on #1127. Also contains the independently reviewed unused authoring deletion from #1126; no new migration seam. These PRs together advance consolidation; this is not the Phase A exit or completion of all animation semantics.


yongkyuns:

PR #1130, opened 2026-09-06
Animate fill and object opacity through ordinary shared play

yongkyuns:
Ordinary TransformTo now animates fill paint and object-composite opacity on the shared Rust runtime. A style transition captures one effective transform/style row, interpolates only its changed channels, reconciles exact authored endpoints at completion, and releases its tracks so the next ordinary setter publishes normally.

Advances #958/#959/#969/#1088. `Property::Fill` uses the existing optional-color interpolation; `Property::Opacity` retains its scalar channel. Fill changes do not clone object geometry. Prepared and predeclared activation share the same channel lowerer; field conflicts and unsupported stroke/content changes fail before publication. Completion reconciles fill and object-opacity channels through one atomic authored style mutation. Non-identity composition timing still rejects before activation until mapped driver release is implemented.

Shared `LiveSession` style operations reuse the same semantic edit helpers as `Mobject`. Python handles argument conversion and typed dispatch. The explicit Noon `set_object_opacity` operation is distinct from Manim `set_opacity`, which continues to control paint alpha. The new operation has no legacy raw-state fallback.

Paired `ordinary_style_play` native Rust/direct Rust-WASM/Python examples animate blue opaque fill to red fill alpha.4/object opacity.5 over2seconds, then perform a normal shared green/full-opacity edit after completion. Rust asserts midpoint and exact endpoint values; both browser paths render the final green object.

Validation: `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed at b4b3e19f. Actual direct smoke then caught unchanged reactive opacity being incorrectly claimed as a timeline channel; 0044b662 fixes that and adds prepared/predeclared regressions. `cargo test -p noon-compile --lib semantic_lowering::animation_payload` (13 tests), `cargo clippy -p noon-compile --all-targets --all-features -- -D warnings`, and a refreshed `NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` passed. The package validated 119 contracts. Actual `node scripts/shared-authoring-smoke.mjs` and `NOON_BROWSER_SMOKE_BACKEND=webgpu node scripts/browser-smoke.mjs` plus the WebGL variant passed.

The shared ordinary preflight introduced by the prerequisite PR now also reports fill support directly from this compiler; its integrated regression confirms unchanged authored revisions and continued rejection of stroke-width changes. This extends the ordinary supported subset; stroke/lifecycle/composition and full callback continuation remain consolidation work.


yongkyuns:

PR #1131, opened 2026-09-06
Advance canonical callbacks to an exactly presented frame

yongkyuns:
Paused canonical execution can now advance to an exact authored time through every required Rust callback barrier and resolve only after the matching retained frame is presented. This enables deterministic callback sampling without a host-owned playhead or callback runtime.

Advances #969/#1088 and the #959 callback-host migration prerequisite. `SemanticExecutionPlayer::advanceForwardToCallbackPhaseJson` reuses the existing session barrier. The semantic worker commits each token-pinned callback phase, bounds the control to 128 distinct phases, and publishes one final delta at the requested time. Required callback failure/cancellation remains terminal where the existing session requires it; no callback side effects are replayed automatically.

Presentation acknowledgement carries only derived transport session/sequence metadata. A failed surface presentation, stale delta, foreign acknowledgement, or unchanged advance with an outstanding earlier publication cannot resolve the control early. An already-presented unchanged frame resolves without a forced redraw. One pending presentation waiter and existing transport/control bounds keep the path local and bounded.

The shared Python callback proof pauses after forward playback, advances another half second, then captures immediately after completion and checks the two callback objects against the exact returned time. There is no post-control polling that could hide an early acknowledgement.

Validation on the combined paint integration: full Rust architecture/fmt/check/Clippy/workspace tests/doctests passed via `bash scripts/check.sh full`; Node/Python web preflight passed via `NOON_WEB_PREFLIGHT_ONLY=1 NOON_SKIP_PLAYGROUND_TEST=1 bash scripts/check.sh web` after updating source assertions for the publication argument. Runtime Node tests cover delayed/failed presentation, unchanged frames, stale/foreign acknowledgements, multiple callback barriers, cancellation and backpressure. Refreshed dev `bash scripts/check.sh web` passed (122 combined package contracts); actual `node scripts/shared-authoring-smoke.mjs` passed with the exact callback presentation assertion, and direct WebGPU/WebGL smoke plus compatibility smoke passed. Required CI qualifies this exact stacked head.

The old raster callback host and its detailed upload/slot diagnostics are intentionally still migration work. This PR supplies the control prerequisite; it does not claim those consumers or the old HostScenePlayer are deleted.


yongkyuns:

PR #1132, opened 2026-09-06
Animate stroke paint and reconcile independent style fields

yongkyuns:
Ordinary TransformTo now animates stroke paint as well as fill and object opacity through the shared Rust channels. Completion reconciles only the fields actually driven by the animation, preserving independent paint fields and reactive siblings, then releases the tracks so ordinary setters work afterward.

Advances #958/#959/#969/#1088. `Property::Stroke` uses the existing optional-color interpolation and the runtime's local style-dirty path. Prepared and predeclared transforms share the same payload/channel lowerer. An unchanged authored bound style domain remains reactive even when its current effective value differs; explicit target changes that conflict with a binding still reject. Stroke width/mode/join/cap, content and lifecycle changes remain outside this slice.

Shared `LiveSession` color/stroke operations reuse the existing `Mobject` semantic edit helpers. Python performs coercion and typed dispatch only. Completion merges Fill/Stroke/ObjectOpacity entries into one authored style mutation per affected object; it does not overwrite an unrelated sibling with a target snapshot or copy geometry for a paint track.

The paired native Rust/direct Rust-WASM/Python `ordinary_paint_play` examples now cover independent fill/stroke edits, a 0.4-second transition to different paint alpha values (.8 and .3), exact completion, a 2-second shared color/opacity transition, and a normal final yellow edit at t=2.4. Rust asserts effective midpoint/endpoint and exact authored reconciliation; Python asserts ordinary getter/time behavior. Both browser paths render yellow. The obsolete legacy export-document opacity fixture was removed only after moving its useful behavior into this paired proof.

Validation: `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed all Rust architecture/fmt/check/Clippy/workspace tests/doctests at2a291c5d; the web phase found stale renderer-source assertions, corrected in the prerequisite callback-control PR. `NOON_WEB_PREFLIGHT_ONLY=1 NOON_SKIP_PLAYGROUND_TEST=1 bash scripts/check.sh web` then passed Node checks and174Python tests. After merging the shared ordinary capability preflight, its focused Rust test passed and refreshed dev `NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` validated166380804WASMbytes/122contracts. Actual `node scripts/shared-authoring-smoke.mjs`, `node scripts/manim-compat-smoke.mjs`, `NOON_BROWSER_SMOKE_BACKEND=webgpu node scripts/browser-smoke.mjs`, and the WebGL variant all passed, including the expanded2.4-second proof.

No new migration seam or runtime is introduced. Mapped composition, ordinary lifecycle/family breadth and realtime source continuation remain separate consolidation work; this PR does not claim Phase A complete. Required CI qualifies the exact stacked head.


yongkyuns:

PR #1133, opened 2026-09-06
Activate ordinary animation compositions atomically through shared Rust sessions

yongkyuns:
Ordinary Parallel and flat Succession authoring now declares target snapshots, child animations, and the composition root in one shared semantic transaction, activates them in the existing runtime, and reconciles their completed state before subsequent ordinary edits. A two-circle example moves both objects in parallel, changes their fills in sequence, then performs an ordinary post-completion style edit through the same session.

Advances #957, #958, #61, #969, and the #1088 continuation foundation under Phase A. Semantic Scene owns provisional declarations and identity; compiler owns one schedule/payload lowering path; ExecutionSession owns atomic activation/publication and completion. Python and WASM builders carry opaque handles/options only. Rust derives duration when root/play overrides are absent.

- Reuses generic transaction-local references for all animation declarations; failed preflight allocates no semantic identity or partial runtime publication.
- Published and prepared animations share the compiler schedule resolver. Removes the obsolete affine-only prepared projection instead of retaining an adapter.
- Sparse effective-state capture visits affected objects once; unchanged channels retain their current driver. Conflicting property drivers fail before commit. Completion reconciles driven endpoints and releases mapped tracks atomically.
- Adds callback-aware segment drive using the existing callback barrier and typed segment failures. Completed receipts are idempotent; tokenless waits remain authored-time intervals.
- Adds equivalent executable native Rust, direct single-context Rust/WASM, and Python examples, with shared midpoint/endpoint assertions and actual renderer pixel checks.

No new migration seam, in-process serialization boundary, scene model, identity allocator, scheduler, or runtime. Existing unsupported Python compatibility consumers remain owned by #959. This slice supports flat linear ordinary Parallel/Succession transforms and paint, with disjoint property drivers. Nested groups, repeated drivers, lifecycle/content animation, and realtime Python source suspension remain open. Synchronous ordinary play demonstrates logical endpoint completion, not realtime source continuation.

Validation:

- `CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed at193cfff8: Rust format/check/Clippy, workspace tests/doctests, Node checks,174Python tests, playground examples, and WASM package contracts.
- After the final invalid-later-leaf preflight correction: `cargo test -p noon-web --lib ordinary_composition_candidate` passed2tests; `NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` passed (166974170byte module,127public contracts). Shared target andincremental0 used throughout.
- Exact-head browser checks passed: `node scripts/browser-smoke.mjs`; `NOON_BROWSER_SMOKE_BACKEND=webgl node scripts/browser-smoke.mjs`; `node scripts/shared-authoring-smoke.mjs`; `node scripts/manim-compat-smoke.mjs`.
- `bash scripts/architecture-ratchet.sh codex/architecture-ordinary-paint` and `git diff --check` passed.

Required GitHub checks remain additional qualification. This is stacked on#1132.


yongkyuns:

PR #1137, opened 2026-09-06
Drive live authoring continuations through shared Rust sessions

yongkyuns:
A live authoring program can now yield at an existing Rust execution segment, drive it through the normal retained renderer, and resume after its coherent completion publication is admitted. The paired example plays to `(2,-1)`, waits, applies an ordinary shift from the completed effective state, and plays to `(5,-1)`.

Advances #1088 and #969 under Phase A; shared semantic/session ownership remains in Rust. `LiveProgram` owns a consumed `Scene`, its matching `ExecutionSession`, one continuation, and one pending segment/publication receipt. Native and direct Rust/WASM hosts reuse their existing lifecycle/render paths. They reanchor wall-time conversion on actual source resume, including late deadlines and zero waits. Direct native/WASM engine boundaries stay typed and in process.

Python async constructs use a generation-pinned worker attach/cancel lane while the original interpreter task is suspended. The first attachment installs resources once; subsequent awaits retake the exact returned player without rebuilding the runtime or resetting transport sequences. Synchronous constructs can explicitly opt into Pyodide JSPI suspension via `realtime_continuation = True`; unsupported JSPI rejects before construct. The playground attaches while source is pending and adopts that attachment after authoring finishes.

Browser ticks are wake opportunities. Rust derives cadence; waits use one deadline and completed continuations settle. Publication retries remain independent of logical advancement. Unsupported playback controls and required Python callback reentry reject explicitly in this first subset.

Temporary seams remain owned by #959: ordinary synchronous endpoint-only behavior outside the explicit realtime mode and unmigrated lifecycle/family/content animation callers. This PR does not complete all #1088 cases or Phase A. Python required-callback continuation, broader ordinary semantics, and remaining migration deletion are still open. No additional scene authority, ID allocator, scheduler, or in-process codec is introduced.

Validation:
- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev bash scripts/check.sh full`: Rust format/check/Clippy, all workspace tests/doctests, JavaScript,174Python tests and playground compilation passed. WASM compiled; packaging initially hit local disk exhaustion.
- After removing superseded generated packages, `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` passed; final package167077370bytes/135contracts.
- Final wrapper fixes: focused repeated-play, pending/completed receipt, recovery and Python returned-mutation regressions passed. `bash scripts/check.sh fmt-lint` with the same Cargo environment passed; `cargo test --workspace --all-features --lib native` passed.
- Final `NOON_WEB_PREFLIGHT_ONLY=1 NOON_SKIP_PLAYGROUND_TEST=1 bash scripts/check.sh web` passed, including174Python tests; `bash scripts/architecture-ratchet.sh c7e6b87a` and `git diff --check` passed.
- `NOON_BROWSER_SMOKE_BACKEND=webgpu node scripts/browser-smoke.mjs` and the equivalent `webgl` command passed, including direct Rust continuation midpoint/endpoint pixels and wait/idle behavior.
- `node scripts/shared-authoring-smoke.mjs` passed after fixes: async and synchronous JSPI continuations, final visible state, prior shared authoring and persisted-scene cases. Deterministic client tests also cover delayed startup/result ordering, cancellation, stale generations, one resource bootstrap, same-player return/resume, cross-worker clock origins, early timer rearm, and explicit unsupported controls.
- Native Linux/Xvfb CI now runs a separate four-second continuation surface test through final publication/admission; that platform-specific test was not run on this macOS host.


yongkyuns:

PR #1134, opened 2026-09-06
Observe canonical callback publications through retained GPU presentation

yongkyuns:
Canonical Python callbacks can now request one exact renderer observation through `advanceToWithRendererObservation(time)`. The result ties the callback's committed row to the matching retained transport publication, prepared GPU instance, actual upload range, draw membership, and successful submit/present. Stale or malformed observations cannot release the control request.

Advances the #969/#961 proof foundation needed for #959 callback-host deletion. Semantic and runtime identities remain authoritative; the renderer exposes derived observations only. The native and direct Rust/WASM engine paths remain typed in-process. JSON appears only on the genuine authoring-worker/render-worker diagnostic boundary.

- Observation is opt-in and bounded to one target/publication. Normal frames allocate no upload trace and copy no diagnostic state.
- Indexed semantic/execution/slot lookups and existing packed-instance/item mappings resolve the target. Dirty-range inspection is proportional to changed ranges.
- Unsupported Text/Mixed upload, family/visibility projection, and path draw mappings report explicit ResourceUnavailable; they do not claim a successful proof.
- Common request DTOs stay native-safe. GPU-specific observation code is WASM/test-only; production native noon-web acquires no renderer dependency.
- The paired shared-Python callback proof checks committed/mirrored/prepared transform and style, local upload without rebuilding, exact presentation receipt, and matching pixels.

Old HostUpdaterDiagnostic, mixed/raster gates, and HostScenePlayer remain until their actual consumers have equivalent canonical proofs. This introduces no compatibility adapter; remaining caller deletion is owned by #959. The next bounded prerequisite is Text/Mixed target glyph upload evidence.

Validation:

- `CARGO_INCREMENTAL=0 NOON_WASM_PROFILE=dev bash scripts/check.sh full`: Rust format/check/Clippy, workspace tests/doctests, Node,174Python tests, and playground checks passed. WASM found an installed-mirror accessor mismatch, corrected in95f36552.
- `NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` then passed (167175394byte WASM,127package contracts).
- `node scripts/shared-authoring-smoke.mjs` passed on the final head after correcting the proof to assert the geometry fast path's absent mixed-item table explicitly.
- Focused endpoint Node tests cover stale/malformed observation rejection and pending-request cleanup; exact-base architecture ratchet and `git diff --check` pass.

[...]

PR #1139, opened 2026-09-06
Keep mixed Text callback updates local and observable

yongkyuns:
Simultaneous Text and geometry callback updates previously rebuilt the mixed retained frame, and native Text wrappers bypassed the callback overlay. The shared renderer now updates stable geometry slots and glyph instance ranges together; Text property callbacks use the existing effective-write path. Paired Rust/Python examples exercise both in one scene.

The existing opt-in renderer observation now identifies target glyph atlas planes/pages and exact instance ranges, recording only overlapping uploads. Text-owned vector geometry is correlated through the renderer's scratch mapping and checked against the target render-item range. Normal rendering performs no observation hashing or trace allocation. Observation work is proportional to selected prepared items and overlapping writes; supported local property updates preserve scratch/order snapshots and resident resources.

Advances #959, #969 and #955; stacked on #1134. Ownership stays in the renderer and shared callback/runtime boundaries. Observation is an explicit diagnostic across a real worker boundary. No new authority or migration seam is introduced. The old host diagnostic is deleted with its consumers in the following #959 slice. Individual-path policy is unchanged; missing exact mega-path mapping reports unavailable.

Validation:
- `NOON_SKIP_PLAYGROUND_TEST=1 NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed before the final localized fixes (workspace tests include playground integration).
- Final Rust changes: `bash scripts/check.sh fmt-lint`, `cargo test --workspace --all-features retained_`, and `cargo test --workspace --all-features --lib simultaneous_geometry_and_text_properties` passed.
- Final Python changes: web preflight passed, plus six focused callback wire tests.
- Final package: `NOON_WASM_PROFILE=dev NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web` passed (127 API contracts).
- `node scripts/shared-authoring-smoke.mjs` passed with the final package, including real Text+geometry callback publication and partial uploads.
- `bash scripts/architecture-ratchet.sh 407d6a79` and `git diff --check 407d6a79 HEAD` passed.

Rust commands used `CARGO_INCREMENTAL=0` and the shared target directory. GitHub native/backend and other required checks remain independent merge gates.

### DIFF
--- a/crates/noon-render-wgpu/src/gpu/retained_text.rs
+++ b/crates/noon-render-wgpu/src/gpu/retained_text.rs
@@ -2843,10 +2843,12 @@ mod tests {
 
     fn geometry_only_mega_path_frame() -> FrameState {
         let path = |id, y| {
-            let mut style = Style::default();
-            style.fill = None;
-            style.stroke = Some(Color::WHITE);
-            style.stroke_width = 0.01;
+            let style = Style {
+                fill: None,
+                stroke: Some(Color::WHITE),
+                stroke_width: 0.01,
+                ..Style::default()
+            };
             FrameObjectState {
                 id: ObjectId::new(id),
                 content: ObjectContentRef::Geometry(GeometryRef::path(
--- a/crates/noon-render-wgpu/src/gpu/retained_text.rs
+++ b/crates/noon-render-wgpu/src/gpu/retained_text.rs
@@ -12,7 +12,7 @@ use noon_core::{
     TextRenderItem, TextResourceLookup, TextVectorItem, Transform2D, Vec2, VectorPath,
 };
 #[cfg(test)]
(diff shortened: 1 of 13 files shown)


yongkyuns:
The browser-fast failure was a stale outer assertion: the paired Rust callback example now has retained Text plus two geometry objects, but `scripts/browser-smoke.mjs` still expected two objects. Fixed in 8447db23; the inner direct probe and semantics already expected three.

Actual `node scripts/browser-smoke.mjs` passed on both WebGPU and WebGL2 against the qualified Text package (ports 4188/4189), covering 11 fixtures at four checkpoints, direct Rust/WASM execution, and 29 public scenes. This updates the existing fixture contract without relaxing rendering assertions. Eight other CI checks passed on the previous head; new-head checks remain required.

yongkyuns:

PR #1140, opened 2026-09-06
Replace legacy callback diagnostics and updater test consumers

yongkyuns:
The old host-updater diagnostic retained separate capture state in the execution canvas, and the updater smoke/locality test constructed a second legacy host player. These consumers now use the canonical callback barrier and retained-renderer observation. The 10,000-object regression starts from typed semantic authoring and proves one staged callback row, one dirty execution target, authored callback order, effective read-after-write, stale-result rejection and atomic failure.

The backend diagnostic follows mixed Circle/Text callback changes through committed and mirrored state, exact dirty instance ranges, overlapping uploads, draw membership and actual presentation. It preserves an unchanged resident sibling and compares visible output on WebGPU/WebGL2. The updater smoke uses the same endpoint and exact forward samples.

Advances #959/#955/#969, stacked on #1139. Deletes HostUpdaterDiagnostic and its special canvas capture state/API; removes the updater smoke and locality test as HostScenePlayer consumers. The remaining raster/gallery and performance consumers still own the old HostScenePlayer deletion. Existing raster oracle fixtures remain intact. This slice does not replace their cross-target callback-read or Line-specific endpoint/rotation coverage.

No new authority or migration adapter. The semantic/runtime callback phase and retained renderer own the behavior; JSON is confined to the actual Python/execution/render worker boundaries and external test observations. Diagnostic tracing is opt-in; local changes remain target-local through publication and uploads.

Validation:
- Initial `bash scripts/check.sh full` passed format/check/Clippy, then exposed an incorrect initial staged-row assertion in the migrated 10k test. Corrected it to distinguish one read-view row from zero initial timeline/prior-driver rows; the first batch has exactly two writes to one target and later phases stage one prior-driver row.
- Final `bash scripts/check.sh test` passed all workspace tests/doctests.
- Final `NOON_SKIP_PLAYGROUND_TEST=1 NOON_WASM_PROFILE=dev bash scripts/check.sh web` passed preflight and WASM package validation (127 contracts). The workspace tests already include playground integration.
- `node scripts/manim-host-updater-diagnostics.mjs` passed actual WebGL2 and WebGPU publication/upload/pixel checks. The fixture uses normal Scene discovery and authored root order to select the observer target.
- `node scripts/updater-callback-smoke.mjs` passed actual software WebGL2 rendering and exact sample/state/locality assertions.
- `bash scripts/architecture-ratchet.sh 8447db23` and `git diff --check 8447db23 HEAD` passed.

Rust commands used `CARGO_INCREMENTAL=0` and the shared target directory. GitHub checks remain independent merge gates.

### DIFF
--- a/.github/workflows/pr-fast.yml
+++ b/.github/workflows/pr-fast.yml
@@ -266,7 +266,7 @@ jobs:
           echo "seconds=$SECONDS" >> "$GITHUB_OUTPUT"
           exit "$status"
 
-      - name: Check host-updater diagnostics on WebGL and WebGPU
+      - name: Check canonical callback renderer observations on WebGL and WebGPU
         if: steps.browser-risk.outputs.host_updater_diagnostics == 'true'
         id: host-updater-diagnostics
         run: |
--- a/crates/noon-render-wgpu/src/gpu/retained_text.rs
+++ b/crates/noon-render-wgpu/src/gpu/retained_text.rs
@@ -1126,31 +1126,22 @@ impl RetainedFramePreparer {
         } else {
             self.geometry_only_classification = Some(false);
         }
-        if self.can_update_mixed_geometry_locally(frame, changes, metrics) {
-            return self.prepare_mixed_geometry_locally(
+        if self.can_update_mixed_properties_locally(frame, changes, texts, metrics)? {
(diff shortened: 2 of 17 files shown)


yongkyuns:

PR #1138, opened 2026-09-06
Run ordinary fades through canonical lifecycle transactions

yongkyuns:
Basic ordinary FadeIn/FadeOut now uses one canonical semantic lifecycle operation and the existing runtime Appearance channel. FadeIn atomically adds a detached leaf with appearance zero; FadeOut atomically releases its driver and removes membership at completion. Re-adding the same handle restores ordinary appearance while preserving authored style and identity.

The Rust LiveProgram, native example, direct Rust/WASM example and Python example share the story: fade in, fade out, remain detached for 0.25 seconds, and re-add the same object. Python supplies typed handles/options and derives wrapper attachment from shared membership; it owns no appearance or lifecycle timeline.

Advances #959/#969/#954 and builds on #1137. Semantic Scene owns intent/membership; compiler lowers the shared track; execution session owns activation, completion and atomic publication. Local entry/exit affects one object and keeps unrelated slots. First bootstrap remains provisional on failure; foreign roots, invalid options and unsupported targets fail without changing ownership/publication.

This is the bounded single-leaf operation. Family/composed fades, shift/scale endpoints, required callbacks and first-time Text entry remain explicit unsupported cases with remaining migration owned by #959. No new compatibility seam was added.

Validation:
- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 NOON_SKIP_PLAYGROUND_TEST=1 NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed: format/check/Clippy, workspace tests/doctests (including playground examples), JS/Python checks, and WASM/package contracts. The skip avoids repeating the playground test under another feature graph after the full workspace test.
- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_INCREMENTAL=0 cargo test --workspace --all-features --lib fade` passed after correcting the shared Appearance driver key.
- `NOON_BROWSER_SMOKE_BACKEND=webgpu node scripts/browser-smoke.mjs` and the equivalent `webgl` command passed, including the direct Rust fade proof.
- `NOON_SHARED_AUTHORING_SMOKE_PORT=4197 node scripts/shared-authoring-smoke.mjs` passed with the final package and repaired observation-only Python fade test. Source-owned continuations correctly reject pause/resume; the test now observes their real midpoint frames without requesting those controls.
- `bash scripts/architecture-ratchet.sh d1311af0` and `git diff --check` passed.

[...]

PR #1141, opened 2026-09-06
Drive ordinary continuations through ordered callback barriers

yongkyuns:
Ordinary live affine playback now services required host callbacks while Python is suspended on the same source stack. Rust retains the segment, callback token, authored time and coherent frame; the browser publishes intermediate frames and completes the segment only after endpoint/reconciliation publications are admitted by the renderer. Ordered callbacks run exactly once at their scheduled boundaries, and required callback latency is excluded from the next playback clock anchor.

The same target-neutral Rust LiveProgram runs through native and direct single-context Rust/WASM hosts, with an equivalent executable Python example. Native CI explicitly exercises the callback example in an OS window. Accepted input arriving during either endpoint presentation barrier is applied through the existing bounded input lane before the source resumes. Callback-session retirement is queued after the interpreter run unwinds.

Advances #969/#1088/#955, stacked on #1138. Runtime owns advancement and completion; platform hosts service lifecycle and callback barriers. No new semantic authority, schedule, player mirror, or in-process transport. JSON remains limited to the genuine Python/execution/render worker boundary. Ordinary synchronous default promotion and callback-bearing target-copy migration remain separate work; the paired example builds its target before callback registration. Legacy callback consumers remain owned by #959.

Validation:
- `NOON_SKIP_PLAYGROUND_TEST=1 NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed all Rust tests/doctests and web preflight; the package checker initially needed correction for wasm-bindgen readonly properties.
- Final `bash scripts/check.sh fmt-lint`, `cargo test --workspace --all-features --lib callback`, `cargo test --workspace --all-features --lib clock`, and web preflight/package passed after the clock fix. Rust used `CARGO_INCREMENTAL=0` and the shared target directory.
- `node --test web/python-worker-source.test.mjs web/semantic-engine-endpoint.test.mjs` passed, including intermediate publication, queued input at both completion barriers, and callback cleanup after interpreter unwind.
- The actual shared browser run exposed and drove fixes for missing intermediate publication and callback cleanup. Its static Fade wait observation now checks committed renderer membership because settled runtime time need not tick before its deadline.
- `node scripts/shared-authoring-smoke.mjs` passed actual shared Python authoring, ordered callback midpoint/endpoint pixels, exactly-once callbacks, same-context source continuation and persisted Scene reuse.
- `NOON_BROWSER_SMOKE_BACKEND=webgpu NOON_BROWSER_SMOKE_PORT=4173 node scripts/browser-smoke.mjs` and the equivalent `webgl` run passed direct Rust/WASM proofs plus 11 fixtures at four checkpoints and 29 public source-equivalent scenes. The simulated direct callback driver now performs the same post-callback wake observation as production.
- Independent review then added terminal handling for intermediate renderer errors; focused Node regressions pass for both intermediate and endpoint failure, proving exactly one failure notification, no further drive/completion, and no returned incomplete player. Successful browser paths were unchanged.
- `bash scripts/architecture-ratchet.sh 0499b371` and `git diff --check 0499b371 HEAD` passed.
- Native display validation runs on Linux CI; this macOS session does not execute that Linux display test.


yongkyuns:
Native CI caught a shared finite-program wake bug: after the callback continuation finished, retained callback registrations still exposed Continuous cadence, and the native host attempted to drive a finished segment. b68ee8f6 restricts LiveProgram timeline wake to Awaiting while preserving runtime frame/input invalidation in the other phases. Registration/runtime authority is unchanged. The same correction covers native and direct WASM.

Added a shared Rust regression across ready/awaiting/completion/finished and a direct-WASM final-idle assertion. Independent source review found no blocker. Linux CI will rerun the native window proof; local focused qualification will run after the ongoing composition baseline build releases the shared Cargo target. This is pending validation, not a claim that native CI is already green.

yongkyuns:

PR #1142, opened 2026-09-06
Measure host callbacks through canonical execution

yongkyuns:
The host callback benchmark now measures the native timeline and Python updater through the same canonical execution endpoint and retained renderer. It deletes its HostScenePlayer, snapshot/patch replay and legacy microtiming consumer. Exact 60 Hz authored samples are an external benchmark input; the runtime still owns evaluation, callbacks and publication. Contexts are authored, measured and released sequentially so one live session is not invalidated by authoring the next case.

Advances #959/#955, stacked on #1140. The external artifact is schema 2: inclusive advance/presentation round trip, cadence, startup, and publication/upload locality. Default coverage remains 1k/10k objects × 1/100 active, 300 measured frames. Raster/gallery HostScenePlayer consumers remain pending canonical cross-target reads and Line-content coverage; this does not claim whole-player deletion.

Validation:
- JavaScript syntax checks and exact-base architecture ratchet/diff check passed.
- `NOON_HOST_CALLBACK_CASES=1000:1 NOON_HOST_CALLBACK_FRAMES=5 NOON_HOST_CALLBACK_WARMUP=2 node scripts/host-callback-perf.mjs` passed against the qualified #1140 WASM package.
- `NOON_HOST_CALLBACK_FRAMES=30 NOON_HOST_CALLBACK_WARMUP=5 node scripts/host-callback-perf.mjs` passed all four cases on explicitly asserted SwiftShader WebGL2. Artifacts were saved outside the repository.
- One active object uploaded 88 bytes and 100 active objects uploaded 8,800 bytes at both total scene sizes. Geometry cache misses and buffered deltas were zero; measured cases had 30 matching presentations and exact object/instance counts.
- Observed p95 inclusive native/host round trip: 1k/1 = 3.2/4.4 ms; 1k/100 = 12.6/38.3 ms; 10k/1 = 2.9/4.7 ms; 10k/100 = 11.9/39.7 ms. These short software-renderer samples qualify workload execution/locality; they are not hardware performance claims.
- No Rust changed or Cargo rerun. The default 300-frame benchmark remains available for longer measurement.

### DIFF
--- a/scripts/host-callback-perf.mjs
+++ b/scripts/host-callback-perf.mjs
@@ -47,17 +47,25 @@ try {
     const state = await page.locator("#status").getAttribute("data-state");
     if (state === "error") throw new Error(await page.locator("#status").textContent());
     const report = await page.evaluate(() => window.__NOON_HOST_CALLBACK_PERF__);
+    assert.equal(report.schemaVersion, 2);
+    assert.equal(report.workload.objects, testCase.objects);
+    assert.equal(report.workload.active, testCase.active);
+    assert.equal(report.native.locality.lastPublication.objectCount, testCase.objects);
+    assert.equal(report.host.locality.lastPublication.objectCount, testCase.objects);
+    assert.equal(report.native.finalState.playing, false);
+    assert.equal(report.host.finalState.playing, false);
     results.push(report);
     console.log(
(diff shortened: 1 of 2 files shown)


yongkyuns:

PR #1143, opened 2026-09-06
Run composition continuations and publish final source edits

yongkyuns:
Ordinary Parallel/Sequence playback can now activate a shared Rust segment and suspend the Python source until its coherent endpoint is presented. The consuming begin operation leaves authored time unadvanced; the existing runtime/callback/continuation machinery drives it. First activation keeps its player provisional until validation succeeds, and failed activation against a returned player preserves its exact lease, frame and publication context.

Final authored edits after the last play are flushed through the returned player’s existing encoder before the Python source result is delivered. The endpoint waits for the exact renderer publication without starting another segment, advancing time, retaking the player, or invoking callbacks again. JSON remains confined to this genuine worker boundary.

The paired target-neutral Rust LiveProgram and Python example run a parallel move followed by a sequence, then apply a persistent color change after release. The Rust example has native and direct Rust/WASM entrypoints. Browser proofs observe real parallel and sequence midpoint pixels and the final colors through the same retained context.

Advances #969/#1088, stacked on #1141. Shared semantic composition owns timing and lowering; no frontend scheduler, new runtime, or migration adapter. The existing endpoint helper and continuation begin share one activation implementation. This supports the existing flat transform-composition contract; scalar/family breadth and ordinary synchronous default promotion remain separate prerequisites, with legacy consumer deletion owned by #959.

Validation:
- `NOON_SKIP_PLAYGROUND_TEST=1 NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed all Rust tests/doctests, web preflight and package validation for the composition baseline.
- After consuming #1141’s finite-program wake correction, `cargo fmt --all -- --check` and `cargo test --workspace --all-features --lib live_program` passed all seven affected tests; the rebuilt package passed.
- The actual Python browser run caught missing final-source publication. After fixing it, `cargo test --workspace --all-features --lib returned_final_publication` and `node --test web/semantic-engine-endpoint.test.mjs web/python-worker-source.test.mjs` passed. Final `bash scripts/check.sh fmt-lint` and web preflight/package passed (144 contracts).
- `node scripts/shared-authoring-smoke.mjs` passed actual Python midpoint/sequence/final-edit pixels with one source lease. The direct probe mirrors production’s post-presentation wake observation and presents any final Rust source edit before inspecting pixels.
[...]

PR #1144, opened 2026-09-06
Drive scalar continuation through shared Track/Hold history

yongkyuns:
Ordinary ValueTracker playback can now activate and complete a shared Rust segment. A bound tracker can play 0→2, persistently set 3, wait, then play 3→5 while Python suspends on the same returned player and observes coherent values after each barrier.

The Semantic Scene owns one ordered Track/Hold history. Runtime lowering derives a per-signal schedule that appends local entries without rebuilding the reactive graph. Completion appends an endpoint Hold and releases active ownership; a persistent setter appends a Hold at the shared session time. Backward seek and fresh lowering preserve the original input, earlier interpolation and later setter gaps. Raw execution-only writes reject authored history so they cannot silently corrupt it.

Activation preflights semantic mutation, scalar lowering, schedule append, publication capacity and the existing segment receipt before committing. Setters update only the existing dirty dependency closure; no second timeline, player, clock or publication lane is introduced. The WASM begin operation and Python adapter consume the same LiveSession operation.

The paired target-neutral Rust LiveProgram and Python example replace the older endpoint-only tracker demonstration. Rust has native and direct Rust/WASM entrypoints; browser proofs observe both moving midpoints, the clean static Hold and final endpoint. Static waits remain settled at their last published time rather than inventing render ticks.

Advances #969/#1088; stacked on #1143. This slice requires an already-lowered bound scalar input. Unbound signal enrollment, general reactive topology changes, scalar composition, interruption and insertion before future entries remain explicit follow-ups. The existing declaration-only export helper remains a #959 consumer-deletion seam until those callers migrate.

Validation:
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target NOON_SKIP_PLAYGROUND_TEST=1 NOON_WASM_PROFILE=dev bash scripts/check.sh full` passed Rust format/check/clippy, all workspace tests/doctests, 174 Python tests, web checks, and the 167465600-byte WASM package with 145 contracts.
- The first full run exposed a history test using an unbound signal outside this slice's contract. Both that fixture and the adjacent rollback fixture now bind the tracker; the latter asserts the precise out-of-range lowering error, preventing a false-positive rejection test.
- Exact-base architecture ratchet and diff check passed.
- Actual `node scripts/shared-authoring-smoke.mjs` passed after deleting an obsolete outer Python rejection guard and regenerating `node scripts/build-python-worker.mjs`. It observes both scalar midpoint pixels, the persistent Hold and final endpoint. The probe now surfaces source errors directly.
- `NOON_BROWSER_SMOKE_BACKEND=webgpu NOON_BROWSER_SMOKE_PORT=4173 node scripts/browser-smoke.mjs` and the equivalent `webgl` run passed all direct Rust/WASM proofs and all 11 fixtures at four checkpoints, with 29 source-equivalent scenes validated.
- No Rust changes followed the full gate; the final fix deletes the obsolete Python guard. The paired real browser example qualifies that route.


yongkyuns:

PR #1145, opened 2026-09-06
Make supported Python play and wait use shared continuation by default

yongkyuns:
Supported ordinary Python `def construct` play/wait operations now use the shared Rust continuation by default. The first eligible operation checks Pyodide JSPI before Rust activation, then suspends the current source stack on the existing player/segment/presentation protocol. Static source does not probe JSPI; async source retains its existing awaitable API. The example-only `realtime_continuation` opt-in is removed.

The app already attaches the early semantic renderer once. The general shared browser examples now do the same, so authoring can remain suspended while the real frame is visible. Unsupported operations after canonical ownership begins continue to reject instead of selecting legacy execution.

`exportDocument` is passed into construct execution explicitly. Normal synchronous export uses the existing Rust endpoint operations without leasing a continuation renderer, then emits the existing current-state codec; async export rejects before setup. This preserves the existing external snapshot behavior after endpoint reconciliation. It does not claim to export a replayable history for completed live segments. That codec and its remaining consumers are still temporary #959 deletion work.

Advances #969/#1088/#61; stacked on #1144. No Rust, renderer authority, scheduler, clock, or new execution transport is added. The same paired native/direct Rust programs remain the engine proofs; ordinary Python affine/fade examples now demonstrate the default language surface.

Validation:
- `NOON_WEB_PREFLIGHT_ONLY=1 NOON_SKIP_PLAYGROUND_TEST=1 bash scripts/check.sh web` passed web source/unit checks and all 174 Python tests.
- `node scripts/shared-authoring-smoke.mjs` passed actual default and async source continuation, scalar/composition/callback/fade midpoint/final pixels, early context reuse, and static/live source cases.
- The same browser run verifies ordinary affine+wait export has the reconciled current transform and duration without a semantic execution descriptor or continuation registration; async export rejects before setup.
- A targeted worker fixture forces JSPI unavailable, verifies the clear capability error, confirms no player activation or time advance, and restores the capability function. Test sentinels avoid Python class name mangling.
- Exact-base architecture ratchet and diff check passed. The six worker lifecycle tests include the previously qualified callback-session retirement regression.
- Reused #1144's exact qualified Rust/WASM package (145 contracts); no Rust changes require another Cargo or direct backend run for this source-only slice. #1144 passed native-window, Rust, browser and architecture CI at publication time, with recovery still pending.

### DIFF
--- a/scripts/shared-authoring-smoke.mjs
+++ b/scripts/shared-authoring-smoke.mjs
@@ -164,6 +164,39 @@ assert builtins.__noon_export_boundary_setup_count == 0
 result = Scene()
 `;
 
+const unsupportedJspiSource = `from noon import *
+import builtins
+import pyodide.ffi
+
+builtins.__noon_unsupported_jspi_original = pyodide.ffi.can_run_sync
+pyodide.ffi.can_run_sync = lambda: False
+
+class UnsupportedJspiContinuation(Scene):
+    def construct(self):
+        circle = Circle(radius=0.4)
+        self.add(circle)
+        builtins.__noon_unsupported_jspi_scene = self
+        self.play(circle.animate.shift((2.0, 0.0, 0.0)), run_time=1.0, rate_func=linear)
+`;
+
+const restoreUnsupportedJspiSource = `from noon import *
+import builtins
+import pyodide.ffi
+
+try:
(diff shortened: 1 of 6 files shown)


yongkyuns:

PR #1276, opened 2026-09-08
fix(architecture): package-aware dependency checks and a unified local gate (#1272 R3/R4)

yongkyuns:
## Final status — R3/R4 landed

**Merged into `master` on September 8, 2026 at 23:42:54 UTC** as **`0b1e4bba26ef4ff7b3299a7b5bb0e33e085411ed`**. This supersedes the earlier review-ready/pending-CI description.

Final source head: **`a463142644ebe4ac47ee40cddac697536c2c273a`**.
Implementation base: `178e3bcdc78a257e7db8285e862b2c1ba978b1f4`.
Three commits separate R3, R4 and documentation/export-test completion; eleven changed files.

All four final-head PR workflows completed successfully. The [maintainer review/validation comment](https://github.com/yongkyuns/noon/pull/1276#issuecomment-5593468199) recorded no blocking findings before merge. The independent [Codex review](https://github.com/yongkyuns/noon/pull/1276#issuecomment-5593552180) subsequently completed on `a463142644` at approximately 23:51 UTC and reported no major issues. This automated result is not a formal GitHub APPROVED review and is not represented as having preceded the merge.

**Overlap resolved:** R3-only alternatives #1274 and #1275 are both closed/unmerged. This is the selected R3/R4 implementation; do not merge those duplicate scanners on top.

## Roadmap ownership

Implements **#1272 R3 and R4**, accepted under **#961 A6.12**, with dependency policy under #960 and timing coordination with #1265. Tooling/tests/documentation only: no engine, API, provider or feature changes. R1/R2/R5 are outside this PR; consult the current #1272 disposition rather than the historical handoffs for those workstreams.

### R3 — package identity rather than dependency key

Replace manifest-key grep with `cargo metadata --no-deps --format-version 1 --offline`. Inspect **declared `packages[].dependencies`**, never the activated `resolve.nodes` graph. All 20 existing forbidden direct edges retain their policy. **Normal/build/dev declarations obey the same direction**, including aliased inline/table declarations, workspace inheritance, optional dependencies and inactive targets. Allowed packages with forbidden-looking aliases remain allowed.

Unreadable/malformed metadata, Cargo failure and missing/misidentified protected workspace packages fail closed: exit 0 checked/pass, 1 forbidden edge, 2 unable to check. Retain the shell entrypoints; replace incomplete manifest fixtures with valid dependency-free Cargo workspaces. This does not replace transitive-graph or native/WASM feature qualification.

### R4 — a common gate before compilation

[...]

Issue #1272, opened 2026-09-08
[Architecture review] Tighten lifecycle/API contracts, development guardrails, and provider isolation


yongkyuns:
## Status and scope

**Accepted refinements are being implemented within the existing Phase A owners.** Architecture authority remains `docs/architecture.md`; this issue does not authorize a separate architecture.

Current disposition (September 9 UTC):

| Refinement | Disposition and implementation |
| --- | --- |
| R1 | Accepted and implemented in #1279 (`1cd9fa46`): exclusive ownership states and recovery, qualified by full CI and PR checks. |
| R2 | Accepted under #958 / #61; public API narrowing and structured authoring failures remain open. |
| R3 / R4 | Accepted and implemented together in #1276 (`0b1e4bba`): package-aware ratchet and shared local architecture gate; full CI plus scanner fixtures passed. Superseded #1274 / #1275 are closed. |
| R5 | Accepted and implemented in #1278 (`74eb25cf`): provider features, explicit font resources, minimal consumer and native/WASM matrix. Controlled cold/warm build cost measurement remains #1265; cached footprint evidence is recorded there and is not a cold-build speedup claim. |

Related frontend consolidation since the audit: #1271, #1273, #1280, #1281 and #1282 merged. #1283 is qualifying the next animation/lifecycle deletion. Python facade work remains #61/#959.

Follow-up to the September 8, 2026 architectural health check. Architecture authority remains `docs/architecture.md`; #953 remains the Phase A tracker. This issue records concrete acceptance gaps and review decisions within existing workstreams, not a second roadmap or a request for a broad rewrite.

Audit baseline: `a14b2f46ddcd0835932c75bd29ea29b09ecf124e`.

Follow-up checked against master at `178e3bcdc78a257e7db8285e862b2c1ba978b1f4`, including the intervening file comparison, relevant issue bodies, #960/#961 progress comments, and focused issue/PR searches. Existing broad ownership is distinguished below from a dedicated, reviewable case. No dedicated case was found for the five refinements below; this is not a claim that their general architectural principles were previously untracked.

## Existing coverage — do not duplicate

| Health-check finding | Existing owner/cases | Disposition |
| --- | --- | --- |
| Replace Python monkeypatch synchronization, parallel bookkeeping and dual checkpoints with the canonical handle facade | #61 A3.1/A3.2/A3.7; #959 A4.2/A4.8 | Already explicitly tracked. Delete by migrated vertical slice; do not create another Python migration project. |
[...]

PR #1279, opened 2026-09-08
refactor(web): make canonical player ownership explicit (#1272 R1)

yongkyuns:
All applicable PR checks passed, including Chromium, Firefox and mobile WebKit. Merged as `1cd9fa46da1236bffc18fbd948b77a6e3db16a55`.

## Scope

Implements **#1272 R1 only**, following the accepted identity refinement under #969. This does not close the R1–R5 umbrella, change the architecture direction, or touch R3/R4 guardrails and R5 providers.

## Changes

- Replace the optional player plus two independent ownership flags with `PlayerOwnership::{Unstarted, Active, Transferred, Returned}`. Document phase rights/exits beside the type; use enum matching internally while preserving the Python/browser `none/active/transferred/returned` observations.
- Keep active local players distinct from returned dormant players. Only a stale returned player may be replaced at an explicit rerun boundary; queries do not silently create another runtime.
- Validate returns against the **existing store identity, scene root, runtime incarnation, and transport session**. A matching revision/session number cannot authorize a foreign or different-runtime player. Valid publication while leased remains permitted because revisions are not frozen into the lease.
- Reject absent/duplicate/foreign/stale returns without replacing the rightful lease. Rejections retain the supplied player instead of dropping it: Rust returns a typed failure owning that player; the WASM rejection exposes `message`, `toString()`, and consuming `takePlayer()` for recovery to its rightful context. Success-path ownership continues to be a move, not a clone.
- Keep callback termination, segment completion, publication, wake clocks, transport state, and GPU retirement in their existing owners. Continuation and recovery reuse the same player/runtime.

The only `noon` API addition is public read-only access to its already-existing opaque runtime identity. Transport identity getters remain crate-private. No new identity allocator, scheduler, runtime, or semantic implementation is introduced.

## Coverage

13 focused Rust regressions cover phase observations, transfer overlap/query/edit rejection, foreign stores, different roots within one store, noncanonical candidates, same-revision/different-runtime candidates, stale transport rejection, rejected-player recovery, absent/duplicate returns, failed handoff atomicity, publication during a lease, stale rerun recovery, callback termination, segment completion, encoder continuity, and runtime move-versus-clone identity.

The existing Manim browser smoke also exercises the **real consuming WASM return boundary**: both same-store/different-root and foreign-store players are rejected, recovered from the thrown ownership error, still produce valid snapshots, and return successfully to their rightful contexts.

## Validation

Integration head `7c78bd49167ff56e07e4ff2c535b19a1bf50e439` includes the merged export deletion (#1273) and local architecture gate (#1276). [Full hosted CI 34292217832](https://github.com/yongkyuns/noon/actions/runs/34292217832) passed all seven Rust/browser jobs. All eight applicable PR workflows also passed on that same head: PR Fast Gate, Architecture Ratchets, Layer Dependency Ratchet, Native Host Smoke, Shared text authoring, Playground authoring recovery, Playground Product Gate, and [Playground cross-browser matrix 34292220242](https://github.com/yongkyuns/noon/actions/runs/34292220242). The cross-browser matrix passed Chromium desktop, Firefox desktop, and mobile WebKit, including the mobile main-thread fallback. No final-head PR qualification remains pending; the PR is merged.

- Local: architecture migration/growth ratchet and renderer-host boundary ratchet pass.
- Local: semantic engine endpoint tests **40/40 pass**.
- Local: focused Python scene-bound handle/updater/live-text regressions **5/5 pass**.
- Isolated exact-source CI: focused Rust ownership tests, `bash scripts/check.sh fast`, and authoring-only WASM compilation passed in [R1 qualification](https://github.com/yongkyuns/noon/actions/runs/34291409393).
- Normal PR Rust, static/unit, browser, and architectural checks passed on the integration head above.

No temporary qualification workflow, compressed patch, generated WASM, or helper script is part of this PR.


yongkyuns:
Reviewed R1 at `983943d59f4945e8cb52cc3df7a4142fe7d572a6`. No blocking design findings: ownership is represented by one exclusive enum, return validation uses the existing store/root/runtime/transport identities, and rejected consuming returns preserve the actual player. The new tests cover stale/foreign returns, callback termination and publication/encoder continuity; the real WASM return-boundary test is particularly useful. Focused hosted R1 qualification 34291409393 passed.

Integrated current master (#1273 mixed-export deletion and #1276 R3/R4) on this PR as `7c78bd49167ff56e07e4ff2c535b19a1bf50e439`. Resolved the single import conflict by preserving the ownership modules and the narrowed post-export imports; the new ownership test module imports its own Vec2. The export projection remains deleted.

On the integration head, `cargo fmt --all -- --check`, `bash scripts/check.sh architecture origin/master`, JavaScript syntax and `git diff --check` passed. Full hosted CI has been dispatched for the integration; no local Rust/full gate pass is claimed. Keep pending qualification distinct from the earlier source-head pass. #1272 remains open for R2/R5.

PR #1282, opened 2026-09-09
refactor: remove Python composition scheduling and timing fallbacks (#959)

yongkyuns:
Python retained a recursive composition scheduler that built tracks/time maps and performed rollback, even though normal composition playback already uses the shared Rust operation. Delete that scheduler, its Scene.play replacement, and its unused JSON/WASM scheduling bridge. Composition classes now carry inert requests only. Also remove selection between Python and Rust time cursors and all legacy native-binding fallbacks: time, waits, admission, scheduling and publication stay with the existing Rust engine.

Advances #61/#959 (A3/A4), with absence ratchets under #961. No new authority, scheduler, identity, transport or temporary seam. The existing uniform family bridge still has other callers under #61; this removes only the now-unused nonuniform bridge. Explicit diagnostic/checkpoint consumers remain separate deletion work. Time queries remain O(1); actual composition work uses the existing shared recursive compiler and locality rules.

The semantic-ownership inventory moves composition from Python duplication to shared Rust (37 shared operations; seven remaining duplicate classifications). Enforce zero budgets for the deleted Python scheduler, cursor/track writes and fallback names; the generated WASM package rejects the retired export.

Validation:
- All 215 Python tests passed after the combined deletion. The updated inert-install regression also passed in a focused run: `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s web/python -p 'test_manim_show_passing_flash.py'`.
- `bash scripts/check.sh architecture origin/master`, `python3 scripts/semantic_ownership_check.py`, `cargo fmt --all -- --check`, JS syntax checks and `git diff --check` passed.
- Existing paired Rust/Python timed/ordinary/mixed-scalar composition and automatic-wait examples exercise the normal shared execution path. Shared browser regression poisons the obsolete Python cursor before zero-time and wait/play checks; unsupported native bindings cannot append legacy tracks to an empty Scene.
- Full hosted CI is dispatched for the combined final head, because the new shared-authoring regression is outside the PR fast gate. The queued previous-head full run is superseded, avoiding two full qualification runs for adjacent deletion slices. Applicable PR checks also qualify before merge. No local Rust/full build pass is claimed with constrained disk space.

Base #1281 is merged; this PR targets master.


yongkyuns:
Full CI [34296149316](https://github.com/yongkyuns/noon/actions/runs/34296149316) passed all seven jobs on `bdadf49bb9cb0e2d1b58ec6205e877269dd2c644`, including the shared-authoring regressions. The separate product comparison failed only the FPS threshold: baseline 50.9, candidate 40.6, floor 40.7. Fixed-frame pixel difference was 0.00%; shell/cold/warm/edit latency checks passed (231/5784/1167/2248 ms baseline;190/5942/1149/2233 ms candidate).

This is not yet classified as proven runner noise. I requested one rerun of only the failed product job [34296153667](https://github.com/yongkyuns/noon/actions/runs/34296153667), with source and thresholds unchanged. Other passing checks are retained. If the failure repeats, investigate the rendering/sample evidence rather than repeatedly retrying or weakening the gate.

PR #1299, opened 2026-09-09
Use direct Rust recovery fixtures and redraw changed surfaces

yongkyuns:
Advances #959 / #953 / #969: platform recovery and camera-density qualification now runs typed Rust scenes directly through `ExecutionSession` and the existing canvas renderer. The browser harness no longer constructs an EngineScenePlayer, reconciles scene JSON or forwards execution deltas within one WASM context. Its unused incremental/scene replacement APIs and final JS fixture loader/corpus are deleted.

This exposed a platform bug: changing the canvas size did not request a new frame from a clean direct session, while assigning identical dimensions could clear an otherwise idle canvas. Resize now leaves identical dimensions untouched and requests one platform frame after a change. The existing recovery-pending flag becomes surface-pending state and is included in the direct wake directive. Camera/surface invalidation does not change semantic/runtime authority or authored revision. Clean unchanged frames remain idle.

The density test now exercises resize-driven presentation directly, verifies an identical resize produces no frame, and compares the runtime's actual scene revision and playhead before/after resize. WebGL/WebGPU recovery, validation-error and initialization-failure checks use the direct scenes with their existing assertions and image thresholds. The harness's explicit render-at-time diagnostic requests a platform camera refresh for repeated failure probes; ordinary scheduling remains event driven. A debug-only scalar scene-revision getter is a read-only diagnostic boundary.

Layer ownership: `noon-web` owns platform invalidation and lifecycle; shared Rust examples author the test scenes; the existing runtime and retained renderer remain authoritative. No new migration seam or engine protocol. Rust legacy codec/player definitions remain for a subsequent deletion under #959 now that their last browser consumers are removed.

Validation: `cargo fmt --all`; `git diff --check`; `bash scripts/check-architecture.sh` passed; `NOON_WEB_PREFLIGHT_ONLY=1 bash scripts/build-web-demo.sh` passed (139 Python tests plus included JS/static checks); `node --check web/browser-smoke.js`; `node --check scripts/camera-density-smoke.mjs`. Hosted full Rust/WASM/browser and required final-head PR checks must pass before merge. No local Rust/GPU execution claimed; CI remains the primary build as requested.


yongkyuns:
Full CI passed at 696615c5, but the product gate exposed a release-only harness failure: its static import required a debug-only fixture. Current head 366b81a3 provides the typed recovery fixture through an explicit `renderer-smoke` feature for release qualification. Normal release packages omit it, and enabling it does not include the other debug example factories. The release product gate enables this feature when building the candidate and checks real fallback/initialization behavior. No visual/performance thresholds changed. Updated head is running qualification before merge.

PR #1303, opened 2026-09-09
Replace callback method installation with direct facade dispatch

yongkyuns:
Callback behavior depended on a final installer replacing Mobject/VMobject methods and capturing the methods installed earlier. Public methods now enter the callback adapter directly. An active callback uses its existing staged property view; otherwise the adapter calls the concrete shared semantic helper. Delete the callback installer and all of its captured-method globals, and remove earlier assignments that would override the stable facade.

Advances #61 A3.6/A3.7. The existing Rust callback ordering, registration, effective batch, and publication protocol are unchanged; Python still owns callable identity and invocation. No new scene, scheduler, mutation protocol, serialization boundary, or migration seam is introduced. Dispatch checks are constant-time; operations retain their existing locality. The semantic-handle installer still owns constructor/layout setup and is tracked by #61 for the next deletion.

Validation: `NOON_WEB_PREFLIGHT_ONLY=1 bash scripts/build-web-demo.sh` passed (145 Python tests, all frontend checks); `bash scripts/check-architecture.sh` and `python3 scripts/semantic_ownership_check.py` passed. All 17 callback tests now run without an installer, including ordered writes, sparse reads, paint operations, copying callback-bound targets, and rejection before mutation. Existing ordinary affine/callback/paint paired Rust/Python examples qualify through hosted full browser CI. No local Rust/WASM build due to disk constraints.

### DIFF
--- a/compat/semantic-ownership-v1.json
+++ b/compat/semantic-ownership-v1.json
@@ -51,6 +51,7 @@
     {"path": "web/python/_manim_rotate.py", "token": "_snapshot_for_object_at", "maximum": 0, "migration_issue": "#61"},
     {"path": "web/python/_manim_semantic_handles.py", "token": "copy.deepcopy", "maximum": 1, "migration_issue": "#61"},
     {"path": "web/python/_manim_semantic_handles.py", "token": "_ir.Mobject(", "maximum": 1, "migration_issue": "#61"},
+    {"path": "web/python/_manim_updaters.py", "token": "_ORIGINAL_", "maximum": 0, "migration_issue": "#61"},
     {"path": "web/python/_manim_updaters.py", "token": "copy.deepcopy", "maximum": 0, "migration_issue": "#61"},
     {"path": "web/python/_manim_updaters.py", "token": "_ir.Mobject(", "maximum": 0, "migration_issue": "#61"},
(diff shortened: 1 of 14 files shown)


yongkyuns:
CI exposed the MovingDots callback entering the live semantic session during Rust execution. The remaining shared-geometry installer was overwriting the stable `set_x`/`set_y` dispatcher. Removed those assignments; ordinary coordinate operations explicitly reach the existing Rust placement adapter, while callback coordinates remain staged effective writes. Added an installer-order regression that performs both coordinate writes and reads their updated callback center.

At `56cf90ee`, all 146 Python tests pass, including directional shared placement; the frontend preflight checks and architecture gate pass. Hosted checks and full CI are running on the corrected head. The old full run also reported one Reset-source assertion in the viewport test; its cause is being tracked separately from the deterministic callback failure. No visual/performance limits changed.

yongkyuns:
The callback fix is included in #1304 at `83055b2e06f3e9b7e0dc7a0408869cc2e8885e29`, which passed full CI and every PR/product/cross-browser check. This smaller head passed full CI and product checks; its sole Firefox failure was Uncreate failing to fetch `https://cdn.jsdelivr.net/pyodide/v314.0.5/full/pyodide.mjs` (CORS/network failure), before any revision or samples. MovingDots passed. We are landing the fully qualified combined head in #1304 rather than spending another full gallery run on this superseded head. No test or performance threshold changed.

PR #1310, opened 2026-09-09
Remove the unconsumed slotted runtime wrapper

yongkyuns:
Advances #953/#959/#960: delete `SlottedSceneInstance`, an unconsumed legacy wrapper around `SceneInstance` maintained only by regression fixtures. The real product already uses `ExecutionSession`, the shared runtime, and the durable `ExecutionSlotTable`. Keeping the wrapper preserved duplicate scene-patch decoding, transaction preflight, dirty aggregation, spatial synchronization, and an independent frame-compaction protocol.

Keep the actual slot allocator and membership implementation, normalized into `execution_slots/table.rs`. Preserve direct stale-generation/exhaustion tests. Replace the wrapper's churn fixtures with shared semantic transaction tests on `ExecutionSession`: 1,000 single-object replacements, a 128-object rotating working set, 1,000 atomic 32-object swaps, contraction from 4,096 live objects to eight followed by 1,000 temporary-object cycles, and failed replacement rollback/recovery. Assertions cover durable slot capacity/reuse, stale semantic and execution handles, unaffected objects, one coherent publication, local spatial retirement, and no hidden full seeks/group rebuilds. Existing session property/timeline/viewport regressions cover the wrapper's other useful claims.

Remove tests specific to the unused wrapper's compaction heuristic and layout handle protocol. This does not claim production frame-tombstone reclamation is solved: the wrapper's compaction was never used by `ExecutionSession`. A coherent maintenance/reclamation barrier in the actual engine remains owned by #955 under the existing architecture's memory contract. No replacement wrapper, compatibility layer, scene authority, or transport is introduced. No public common semantic behavior is added, so paired Rust/Python examples remain unchanged.

Validation: `cargo fmt --all`; `bash scripts/check.sh architecture`; `bash scripts/architecture-ratchet.test.sh`; `git diff --check`. Hosted Rust and product qualification pending before merge. The architecture guard rejects restoration of the wrapper and its obsolete compaction types across the full Rust tree.

Local execution qualification: `CARGO_TARGET_DIR=/tmp/noon-geometry-check CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 cargo test -p noon --no-default-features --lib execution_session::publication::tests:: -- --nocapture` passed all 22 publication regressions, including the new churn/atomicity cases, in 2.74 seconds after the small initial build. The tests use the existing semantic transaction contract: additions and edits precede the terminal removal suffix; runtime membership still retires exits before reusing slots for entries. The first hosted revision caught a missing test import; it is corrected and the superseded full workflow was canceled.


yongkyuns:
#1311 contains all commits in this PR plus removal of the shared runtime's remaining legacy patch-codec entrypoints. It also includes the merged #1309 renderer fixture migration and resolves their overlapping architecture-guard additions. Qualification and merge will use that combined head, avoiding a separate rebuild of #1310 solely for the guard conflict. The combined local check passed 153 runtime tests, lint, and the minimal public Noon consumer. This PR's full CI 34323667872 also passed. This PR will land through the combined merge once its required hosted checks pass.

PR #1333, opened 2026-09-09
Remove raw Mobject construction and unused Python factories

yongkyuns:
Advances #61 (Phase A3). Public shape constructors already use typed shared Rust options, but the base Mobject still accepted the obsolete raw geometry/transform/style snapshot constructor. Its only remaining callers were two color test fixtures. Remove that entry point, its raw-to-options adapter, and the now-unused `_noon_ir` shape factories. Mobject/VMobject are base types; callers construct concrete shared shapes such as Circle, Rectangle, Line or Path.

Also remove ShowPassingFlash's unreachable snapshot-geometry fallback: admission now requires the existing typed Line observation. Explicit inspection/export snapshots remain inert and cannot be fed back through the removed constructor. No replacement scene representation, compatibility adapter, or new serialization boundary is introduced.

The color fixtures now use the public typed Rectangle constructor while retaining independent fill/stroke opacity assertions. A boundary regression protects absence of the raw constructors. The ownership test now detects the next Python function directly rather than naming a deleted neighboring function; its Rust ownership assertions are preserved.

Validation: `NOON_WEB_PREFLIGHT_ONLY=1 bash scripts/build-web-demo.sh` (including the Python suite and JS structural/unit checks), `bash scripts/check.sh architecture`, and `git diff --check`. No Rust implementation changed, so no additional local Rust build was run. Existing paired shape/color examples continue to use unchanged shared Rust operations. Normal PR browser/product checks qualify this Python-only deletion; no redundant manual full Rust workflow is requested.

The browser authoring smoke also asserts raw factory absence instead of saving/poisoning the deleted factory. `node scripts/manim-compat-smoke.mjs` passed locally against the CI-built package from 34347774147 (Rust source is unchanged) with the current Python worker. Python suite: 152 passed.

### DIFF
--- a/compat/semantic-ownership-v1.json
+++ b/compat/semantic-ownership-v1.json
@@ -56,7 +56,7 @@
     {"path": "web/python/_manim_updaters.py", "token": "_ORIGINAL_", "maximum": 0, "migration_issue": "#61"},
     {"path": "web/python/_manim_updaters.py", "token": "copy.deepcopy", "maximum": 0, "migration_issue": "#61"},
     {"path": "web/python/_manim_updaters.py", "token": "_ir.Mobject(", "maximum": 0, "migration_issue": "#61"},
-    {"path": "web/python/_noon_ir.py", "token": "copy.deepcopy", "maximum": 4, "migration_issue": "#61"},
+    {"path": "web/python/_noon_ir.py", "token": "copy.deepcopy", "maximum": 3, "migration_issue": "#61"},
     {"path": "web/python/_noon_ir.py", "token": "_snapshot_for_object_at", "maximum": 0, "migration_issue": "#61"},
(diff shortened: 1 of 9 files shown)


yongkyuns:
R3 coordination (#1292): I rechecked master `d6734d2a` and #1331's merged target cleanup before editing. My Python scope will be a new `_noon_errors.py`, explicit membership/continuation error catches in `_manim_scene.py`, public exception exports in `noon.py`, module-list registration, and tests. No `_noon_ir.py` constructor/factory, ShowPassingFlash, `_manim_compat.py` or target-state cleanup changes are intended. Rust changes stay in noon-web's error projection/conversion paths; R2 owns the shared producers. I will preserve your constructor deletion and reconcile the actual combination before final qualification.

PR #1286, opened 2026-09-09
feat: retain typed handle and scene membership errors (#1272 R2 slice)

yongkyuns:
## Scope

Typed handle validation and Scene membership errors for **#958 / #1272 R2**, now reconciled with the merged public-facade work in #1290. **This does not complete all R2 error producers or Python exception mapping.**

Current head: **`d1c72e55f03817b4bb2efe815c80e8b9f32b8747`**. Integrated master: **`7e3ee4feab9aa281efd7f256e292cffcc18265fa`**. The reconciliation commit retains both the previous PR head (`6e97a36a`) and current master as parents; the PR branch was fast-forwarded, not force-pushed. Current-base delta: 43 files, mostly explicit caller adaptations around the focused error implementation and tests.

## Changes

- Add explicitly exported `AuthoringError` for `Mobject::validate`, `MobjectFamily::validate`, and authored Scene `add`, `remove`, batch, clear, replace and membership-edit operations. Preserve semantic/transaction causes and rejected node/resource identities.
- Carry handle/membership failures through `LiveSessionError::Authoring`, preserving existing foreign-store, publication, callback and segment categories. Expose nested causes through `std::error::Error::source`; callers need not match diagnostic strings.
- Update still-String-returning internal/example and existing browser-boundary callers explicitly. There is no blanket `From<AuthoringError> for String`, compatibility result alias, or new session-error framework.

No membership semantics, mutation ordering, store/runtime ownership, serialization boundary, scheduler, renderer, provider feature or successful-path traversal changes are intended. Handle checks remain local; rejection precedes atomic publication.

## Reconciliation with current architecture

The public root/prelude exposes `AuthoringError` deliberately. The independent consumer imports raw fixture types through `noon::integration` and uses `integration_store()`. The narrowed API, existing public-facade tests, ordinary `shared_authoring` example and provider workflow from #1290 are preserved.

The semantic scene-operation cause implementation resides in its current `noon-core::semantic_store` module; the retired reactive path is not restored. Current compiler-owned publication preparation and intervening family/grid, renderer-fixture and cleanup changes are retained. Recent String-returning example factories receive only explicit error conversion at their membership calls.

There are **no Python implementation, renderer-crate, manifest, production-workflow or architecture-document changes** in this PR. Development helpers remain outside it on the separate prep branch.

## Regression coverage

[...]

PR #1336, opened 2026-09-09
fix: retain semantic transaction causes through Error::source (#1272 R2)

yongkyuns:
## Scope

Closes the concrete lower-level cause-chain gap found during #1286 review, under #958 / #1272 R2 / #1292 R2. Coordinated as a disjoint three-file patch with the active Rust producer and WASM/Python mapper owners. Does not replace their implementation or claim all R2 is complete.

Base: `d6734d2a0626f5c9926e4043ad2017a94f5c8c5a`.
Head: `48ee4fcdd84f4444b56205524839e2f9409f270e`.

## Changes

`SemanticMutationTransactionError::source()` now borrows the existing nested Signal, SignalTrack, Object, Family, AnimationTarget or Node error. `SemanticTransactionReadError::source()` exposes its Existing scene-operation error. Leaf rejections have no invented cause. No new enum variants, diagnostic wording, allocation, validation policy, preflight, commit, publication, identity or successful-path traversal change.

Four external core integration tests cover every immediate cause-bearing variant, leaf termination, borrowing the original cause rather than cloning it, a real rejected prepare preserving state/resources/revision/counters followed by successful recovery, and a failed staged read followed by a valid read/drop without publication. Tests inspect types and values, never diagnostic substrings.

## Qualification

Hosted development run **34351140285**, attempt 1, passed before source publication:

```
cargo fmt --all
git diff --check
cargo test -p noon-core --test transaction_error_causes
cargo test -p noon-core --lib semantic_transaction
cargo clippy -p noon-core --all-targets -- -D warnings
bash scripts/check.sh architecture d6734d2a0626f5c9926e4043ad2017a94f5c8c5a
cargo test -p noon-core --test transaction_error_causes --target wasm32-unknown-unknown --no-run
```

**4/4 new tests and 139/139 existing transaction tests passed**, strict Clippy and architecture gate passed, WASM test executable compiled. Compilation is not browser execution. Final PR checks remain separately required. The editing container has no Cargo; it inspected the returned exact candidate diff and logs, not a claimed local Rust run.

## Architecture / landing

Read AGENTS and architecture; no new authority, runtime, codec, dependency or migration seam. Only existing error references are exposed. Development helpers are confined to the separate prep branch and are absent from this PR. No renderer/Python/manifest/workflow changes. The user authorized merging after successful review/qualification; no bypass or auto-merge is requested.

Other domain errors may still need their own source forwarding and producer conversion; the active R2 owner retains that work. This patch is intentionally independent of #1286 so both can land without a competing AuthoringError design.

### DIFF
--- a/crates/noon-core/src/semantic_store/semantic_transaction.rs
+++ b/crates/noon-core/src/semantic_store/semantic_transaction.rs
@@ -2746,7 +2746,19 @@ impl std::fmt::Display for SemanticMutationTransactionError {
     }
 }
 
-impl std::error::Error for SemanticMutationTransactionError {}
+impl std::error::Error for SemanticMutationTransactionError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::Signal { error, .. } => Some(error),
+            Self::SignalTrack { error, .. } => Some(error),
+            Self::Object { error, .. }
+            | Self::Family { error, .. }
+            | Self::AnimationTarget { error, .. } => Some(error),
+            Self::Node { error, .. } => Some(error),
+            _ => None,
+        }
+    }
+}
 
 #[cfg(test)]
 mod base_tests;
--- a/crates/noon-core/src/semantic_store/semantic_transaction/prepared.rs
+++ b/crates/noon-core/src/semantic_store/semantic_transaction/prepared.rs
@@ -25,7 +25,14 @@ impl std::fmt::Display for SemanticTransactionReadError {
     }
 }
 
-impl std::error::Error for SemanticTransactionReadError {}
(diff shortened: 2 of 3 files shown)


yongkyuns:
@codex review

Please review the exact three-file delta at `48ee4fcdd84f4444b56205524839e2f9409f270e` against `d6734d2a0626f5c9926e4043ad2017a94f5c8c5a`. The change only exposes references to causes already held by SemanticMutationTransactionError and SemanticTransactionReadError. Review borrow/cause lifetimes, completeness of cause-bearing variant handling and the real prepare/read rejection/recovery tests. No semantic change or new error framework is intended. The user has authorized merge after sound review and applicable qualification; no approval or bypass is requested.

PR #1352, opened 2026-09-09
refactor: expose animation requests without import-time replacement

yongkyuns:
Public animation imports previously returned different classes before and after `_manim_animate` loaded; an early `Create(..., run_time=...)` could therefore reject valid options. Define the animation requests directly in their existing module and expose them through Noon's existing static export table. Remove eight captured base classes, their duplicate records, and the two public-export mutation loops. Move ScaleInPlace/ShrinkToCenter requests beside the other animation requests and leave the option module responsible for adapting the shared Rust resolver.

Advances #1292 R6b / #61 under Phase A #953. Python owns signatures, coercion and inert request fields; Rust option resolution, activation, lifecycle and playback are unchanged. No new semantic state, scheduler, protocol, migration seam or runtime work. Public class identity remains stable across subsequent module imports; the ordinary Scene consumer uses the defining request class directly.

Validation:
- `PYTHONPATH=web/python PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s web/python -p 'test_*.py'`: 168 run, nine browser-only skips.
- `bash scripts/check.sh architecture`: passed.
- `NOON_WEB_PREFLIGHT_ONLY=1 bash scripts/build-web-demo.sh`: passed, including the existing web checks and the same Python suite.
- `git diff --check`: passed.

The existing constructor regression now creates an option-bearing public request before Scene/bootstrap imports and verifies class/instance identity afterward. Existing fade, transform, create/uncreate, indication, scale and import-without-host tests pass. Existing paired Rust/Python examples cover unchanged semantics; no new semantic feature requires a duplicate example. Actual browser qualification is now green in PR CI. This Python-only cleanup does not trigger a separate full Rust rebuild or claim that R2/R3 are complete.

Final qualification at `89b014daf972068bee53de271d82b2fbfb1d7576` (tested merge `abcbf56c99d6d5eb447b9a8dd70db0d2696a4441`): all applicable checks passed, including architecture, Rust/web fast gates, production recovery, Chromium/Firefox/mobile WebKit, and product comparison. Source review confirms consumers use the same defining classes directly, preserve all inert request fields/options, and do not depend on the deleted dataclass machinery. No per-frame execution path changes.

Product run34375140530 attempt1 failed FPS49.3→38.4 against floor39.4 despite matching pixels and passing latency checks. One bounded **measurement-only** retry (attempt2) reused the same successful build/artifact job and passed FPS40.2→43.8, fixed-frame pixel diff0.00%, and all latency limits. No source, thresholds, provenance or assertions changed; no further retry was needed. Logs are retained by that run. The first failure is not erased or called a pass.

This completes the animation import-order seam. R6 remains open for the source-level family layout/translation residuals recorded in #61 comment5605218051, and R2/R3 errors remain with their owners.

### DIFF
--- a/web/python/_manim_animate.py
+++ b/web/python/_manim_animate.py
@@ -14,27 +14,17 @@ import _manim_compat as _compat
 import _manim_rate_functions as _rate_functions
 
 
-_ORIGINAL_TRANSFORM = _base.Transform
-_ORIGINAL_REPLACEMENT_TRANSFORM = _base.ReplacementTransform
-_ORIGINAL_TRANSFORM_FROM_COPY = _base.TransformFromCopy
-_ORIGINAL_TRANSFORM_MATCHING_SHAPES = _base.TransformMatchingShapes
-_ORIGINAL_CREATE = _base.Create
-_ORIGINAL_UNCREATE = _base.Uncreate
-_ORIGINAL_FADE_IN = _base.FadeIn
-_ORIGINAL_FADE_OUT = _base.FadeOut
 _PURE_YELLOW = _base.color_from_hex("#FFFF00")
 
 
 def _store_animation_args(animation: object, kwargs: dict[str, Any]) -> None:
     """Attach Manim Animation kwargs for the shared option resolver.
 
-    Noon's original animation records are frozen/slotted dataclasses. Subclassing them
-    gives the compatibility facade a small Python ``__dict__`` for authoring metadata
-    while preserving the existing low-level fields and isinstance-based lowering.
-    Validation remains centralized in ``_manim_animation_options`` at play time.
+    These inert requests carry Python call-shape metadata only. Validation and
+    execution remain in the shared Rust option/animation operations.
     """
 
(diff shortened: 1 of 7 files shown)


yongkyuns:
All functional/fast/recovery and Chromium/Firefox/mobile WebKit checks passed at89b014da. Product comparison attempt1 used identical visible-pixel counts217414 and fixed-frame pixel diff0.00%; shell/cold/warm/edit latency checks passed. Only effective FPS failed: baseline49.3, candidate38.4, floor39.4. The changed files contain only request construction/static export adaptation, with no Rust or per-frame algorithm change. This is insufficient evidence to dismiss the failure as noise.

Initiated one bounded rerun of comparison job102548566065 in run34375140530 using the same immutable production artifacts; the successful build job is not rerun. No threshold, source, artifact provenance or test assertion changed. If that measurement fails again, investigate rather than repeatedly retry. No merge while this check is unresolved.

Issue #954, opened 2026-09-02
[Roadmap B] Common 2D semantics on the consolidated architecture


yongkyuns:
## Authority

Phase B execution tracker under `docs/architecture.md`. Phase B becomes the default feature priority only after the Phase A exit gate in #953.

This umbrella owns cross-feature invariants and sequencing. Child issues own API/detail. Existing pre-consolidation implementation remains useful only as specification/evidence where it conforms to the current architecture.

## Phase invariant

Every supported 2D feature:
- authors persistent identity/structure/content/declarations into the same Semantic Scene from Rust/Python;
- exposes current live values from coherent **effective Runtime state** when drivers are active;
- lowers through the same execution/runtime path;
- keeps heavy content/resources immutable, retained and versioned;
- uses effective driver/resource updates for transient live output rather than semantic mutation every frame;
- reconciles persistent final results through normal semantic mutation only when semantics require it;
- preserves direct-seek/deterministic behavior where applicable;
- preserves work proportional to affected/active/visible state;
- qualifies authored structure and effective observables separately against pinned ManimCE v0.21;
- uses paired executable Rust/Python cases for significant semantics supported in both frontends.

No feature may create a Python scene engine, feature-specific runtime, renderer-owned semantic type, second identity/camera/resource authority, in-process serialized bridge or target-specific semantic implementation.

---

## B1 — core object/family/value/query semantics
Primary: #74.

- stable family/order/alias semantics;
- layout-accurate authored values/bounds;
- normal live getters observe effective transforms/styles/bounds/current values;
- persistent mutators/layout/state/copy/target behavior use shared semantic operations;
- copy/target/animation activation explicitly define when current effective state is snapshotted/reconciled.

**Gate:** Python/Rust share behavior and live-query semantics; no frontend bounds/family/layout engine.

## B2 — geometry/imported resources
Parents: #76/#77/#78/#79; resource foundation #53.

- common shapes, lines/arrows, paths/VMobject operations and SVG/images;
- immutable retained content/versioning;
- transform/style/effective driver changes do not rebuild topology/content;
- persistent vs transient resource replacement follows #368.

**Gate:** representative semantic/effective/raster fixtures pass with no geometry cloning or renderer semantic authority.

## B3 — animation/lifecycle
Parents: #80/#82; dynamic breadth #81; family text cases #705.

- deterministic animation intent remains semantic until lowering/activation;
- activation captures current effective start state when semantics require it;
[...]

Issue #83, opened 2026-08-24
[Roadmap B4] Text, math, and numeric text APIs


yongkyuns:
## Roadmap

Phase B4 text/math/numeric-text parent under #954. Architecture authority: `docs/architecture.md`.

## Goal

Deliver common ManimCE v0.21 text/math/numeric-text behavior on one shared semantic/resource architecture with correct persistent-vs-effective update semantics.

## Text/resource contract

- authored text/math source, formatting and semantic part identity belong to the Semantic Scene;
- immutable shaped/compiled artifacts are versioned retained resources shared where valid;
- persistent source/format edits use semantic content mutation;
- tracker/reactive/animation-driven numeric or text content that changes transiently may use C4/#368's effective versioned resource/content lane rather than semantic `ReplaceContent` every frame;
- if a dynamic final value must persist as authored state, reconcile it through normal semantic mutation.

### Native Text/MarkupText
Common font/size/weight/style, multiline/layout and range/substr styling through #364.

### Typst
Typst/MathTypst as first-class source kinds normalized into the shared backend-neutral retained representation.

### LaTeX
Real Tex/MathTex/core SingleStringMathTex semantics through #365, including source/part identity for indexing/coloring/Write/Create/matching.

### Numeric text
DecimalNumber, Integer, Variable and common formatting/update behavior. Static values are ordinary authored content. Live tracker/native/host-driven numeric changes use effective resource/content updates where possible and do not automatically mutate semantic source every frame.

### Code-like text
Optional breadth; split if dependency/highlighting work is large. It must not block core B4.

## Architecture constraints

- one semantic Text object unless observable semantics require semantic children;
- shared Rust shaping/layout/part semantics;
- one backend-neutral compiled-resource boundary #369;
- retained atlas/vector rendering with lazy outlines;
- no backend-specific scene/runtime/renderer;
- no frame-time compiler work for static text;
- transient dynamic text/resource output is versioned/local and not synonymous with semantic mutation;
- Python remains a thin handle/live-query facade; JS/TS #259 is not a Phase B dependency.

## Acceptance

- representative Text/MarkupText/Typst/MathTypst/Tex/MathTex/numeric-text cases author through shared semantics;
- Python/Rust equivalent authored and live observable behavior agrees where both APIs expose it;
- identical immutable text can share compiled resources;
- static text has no host/per-frame compiler work;
- tracker-driven numeric text can update locally without semantic revision churn;
[...]

PR #1371, opened 2026-09-09
Share stable text source-part queries across Rust and Python

yongkyuns:
Text objects can select authored UTF-8 source spans and repeated substrings through shared Rust operations, with the same immutable part observations exposed to Python. Source spans and keys remain stable across repaint, movement and recompilation of the same source; cluster/vector ranges describe the current retained text resource. This advances #83/#364 under #954 and supplies the query substrate for #1376.

The shared Semantic Scene/resource layer owns matching and projection. A substring query traverses glyph/vector records once, binary-searches the sorted match boundaries, and routes each record only to intersecting matches. A prefix pass resolves insertion positions for source-only spans such as newlines. It does not rescan the entire resource per match or add a retained frontend index. Duplicate/reordered clusters are normalized within the selected ranges; invalid UTF-8 boundaries and non-contiguous coverage fail explicitly. No scene/resource mutations or new semantic IDs are allocated by a query.

Python performs argument/result adaptation and releases every typed WASM part/list owner, including on conversion errors. Paired `text_source_parts` examples run through shared Rust sessions, native windows, direct Rust/WASM factories and Python, demonstrating repeated Unicode source selection and immutable resource reuse across presentation edits.

Validation:
- `cargo test -p noon-core text_parts --lib`: 8 passed, including 10,000 UTF-8 matches/newlines, reordered duplicate glyphs and disconnected vector coverage.
- `cargo test -p noon --all-features text_part_authoring --lib`: 3 passed, including the paired execution/seek example.
- `PYTHONPATH=web/python python3 -m unittest web/python/test_manim_text_source_parts.py`: 4 passed, including WASM owner cleanup on success and failure.
- `NOON_WASM_PROFILE=dev bash scripts/check.sh full`: passed (full Rust workspace, Python/JS checks and 220 generated browser API contracts).
- `node scripts/manim-differential-noon.mjs`: 73 real Rust/WASM/Python semantic fixtures executed.
- `node scripts/shared-authoring-smoke.mjs`: passed, including the new source-part example, transferable/shared execution, live membership and persisted scenes. This run also exposed stale visible-fill `get_color()` expectations in the older paired live-geometry example; both Rust/Python assertions now follow the existing shared color semantics.
- GitHub exact-head qualification remains required before merge.

There is no migration seam or new scene/text/runtime authority. This slice does not claim complete native Text/MarkupText selection APIs, range coloring, Typst/Tex/numeric text breadth or B4 completion; those remain in #83 and its child tracks. Source-part observations are authored resource queries, not a claim to expose transient glyph/morph render output.


yongkyuns:
Root integration update: I am driving #954 through completion and will integrate qualified helper changes. My current B1 replacement batch is on `codex/b1-family-replacement`: shared `LayoutAnchor::replace_layout`, `LiveSession::replace_layout`, the corresponding WASM layout methods, and `_replace` wrapper routing. It touches `authoring_mobject.rs`, `canonical_authoring_scene.rs`, `semantic_execution_player.rs`, and `live_session/family_layout.rs` in distinct replacement-method blocks; please preserve those operations when updating your branch after integration.

Please finish the claimed feature batch, resolve the current format/lint failure, and update the PR body/evidence to match the final implementation head. Root will self-review source and merge when the batch qualifies; do not request automated Codex reviews. Continue to record follow-up scope in the owning issue so subsequent helpers can claim independent work.

yongkyuns:
Source review of `efdfbf16`: the ownership direction is sound (source matching and coverage projection live on the retained Rust resource). Two items to include before finalizing this batch:

1. `source_parts_for` calls `source_part` for every match, and each call scans/sorts all glyphs and vectors. Repeated one-character matches make this quadratic in the text length. Please batch the projection or use a reusable derived source index; avoid storing a frontend map or a second text authority. Add one focused repeated-match structural regression.
2. The public Rust/Python query is now exposed, so include equivalent executable Rust and Python examples on the ordinary execution/rendering path, and update the body to describe the wrapper API that is already in the diff. Query observations may stay immutable derived values; do not present a source span as surviving arbitrary source edits unless that behavior is actually defined.

The current CI failure is plain `cargo fmt --all -- --check` (run 34419401209), including missing trailing newlines and wrapping. Apply formatting with the complete feature batch to avoid another formatting-only CI cycle. Root is continuing B1 and will integrate your qualified final head.

yongkyuns:
Integration snapshot at `1e67cedc`: the latest format gate still fails (run34421850890), and the query implementation still calls a full cluster/vector scan per match. The earlier requested single-pass/multi-range query and paired executable Rust/Python evidence remain merge requirements. Please include formatting, the locality fix, examples and final PR scope together in the next source push to avoid repeated CI builds. #1376 can stay stacked; the color query/rewrite should consume that same corrected shared range selection.

PR #1376, opened 2026-09-10
B4: add Rust-owned native Text range colors

yongkyuns:
## Summary

Restacked directly on current `master` through prerequisite #1414. Implements the bounded #83 / Roadmap B4 range-color slice for native `Text` while consuming #1371's landed source-part APIs rather than carrying its old stack.

- adds immutable `TextResource::with_source_fills` projection that preserves shaped glyph/cluster/layout identity and splits runs only where intrinsic fill changes
- keeps #1371's `text_part_authoring` implementation untouched; range styling delegates substring/source-span selection to those landed Rust queries
- rejects UTF-8/cluster-cut spans and Manim-ambiguous conflicting overlaps before object publication
- resolves ManimCE v0.21.0 substring and `[start:end]` selectors in Rust against the original source (including whitespace, negative endpoints, UTF-8 byte projection)
- publishes styling as a replacement immutable text resource plus semantic content-handle swap; object transform/style identity is unchanged
- projects one typed selector/color batch through WASM; Python validates only argument/color shape and forwards selectors unchanged
- preserves #1371's WASM-wrapper `free()` ownership cleanup and also frees the new fill-batch wrapper
- exposes `Text(..., t2c=...)` and the v0.21 `text2color` precedence alias
- explicitly rejects styled `Text` construction after live execution starts until content-resource replacement is routed through the owning live session

## Geometry invariant

The Rust semantic regression snapshots object transform/layout bounds/width/height/center plus retained resource bounds, baseline, layout artifact, vector items, run font/stroke/transform and every positioned glyph (source cluster, origin, advance and glyph bounds), then verifies the range-color edit changes paint/run partitioning only.

## Scope / boundaries

This PR does **not** implement `t2f/t2s/t2w/t2g`, `MarkupText`, Typst/Tex part styling, dynamic text, generic resource lifetime, or renderer threshold changes. Native `Text` point sizing was corrected independently in prerequisite #1414; this PR carries no sizing/metrics delta relative to its current base. Raster thresholds remain unchanged.

## Qualification

Focused Rust/WASM/Python, Fast Gate, both renderer backends, and the pinned `text-range-colors` ManimCE v0.21 raster fixture are being rerun on the restacked exact head. The PR is not merge-ready until the existing raster ratchet passes without threshold relaxation.

Refs #83
Part of #954

### DIFF
--- a/.github/workflows/manim-raster-differential.yml
+++ b/.github/workflows/manim-raster-differential.yml
@@ -7,6 +7,12 @@ on:
       - "parity/manim-v0.21/core-examples/**"
       - "web/semantic-preview-session*"
       - "web/manim-raster-host.js"
+      - "web/python/_manim_typst.py"
+      - "web/python/examples/text_range_colors.py"
+      - "crates/noon-core/src/resources/text*.rs"
+      - "crates/noon/src/text*.rs"
+      - "crates/noon-web/src/text_parts.rs"
+      - "parity/manim-v0.21/text_range_colors*"
       - "scripts/semantic-preview-*"
       - "scripts/agent-preview-capture*.mjs"
       - "scripts/manim-differential.py"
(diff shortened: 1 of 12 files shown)


yongkyuns:
Qualification update: the post-#1414 residual raster mismatch is now isolated to native bundled font-face selection, not range-color semantics. `text-range-colors` reached exact 36 px height parity, but Noon remained 3 px wider/heavier because family lookup selected the first `DejaVu Sans Mono` asset; `typst-assets` orders the Bold face before Regular. Prerequisite #1417 now prefers normal/upright/normal-stretch while preserving the old first-family fallback. I will restack this PR after #1417 qualifies/merges and rerun the pinned raster with unchanged thresholds.

PR #1377, opened 2026-09-10
Bind preview provenance to loaded worker and WASM bytes

yongkyuns:
Advances #1197 / R4c by binding preview provenance to the runtime bytes the browser actually loads. It does not expose rendering through MCP and does not add scene/runtime/renderer authority.

## Contract

`scripts/build-runtime-identity.mjs` creates deterministic `runtime-build-identity.json` from the generated Python worker, wasm-bindgen glue, WASM bytes, and runtime verifier. A clean checkout contributes its exact Git revision; dirty or non-Git builds record `sourceRevision: null` rather than guessing.

The generated Python worker loads that identity, validates the envelope/build digest, hashes the exact served worker/WASM/glue/verifier bytes, and passes the verified WASM bytes directly into wasm-bindgen rather than allowing an implicit second WASM fetch. The existing worker `ready` message carries the verified identity.

`ProvenancedPythonAuthoringClient` is a thin observer over the canonical `PythonAuthoringClient`: it consumes only that existing ready message and returns the verified identity from `ready()`. `SemanticPreviewSession` retains the identity in its coherent snapshot when the supplied authoring client provides one; legacy/test clients preserve the old snapshot shape.

## Qualification corrections on this branch

Three red paths were traced to test/package seams rather than runtime semantics:

- the gallery no-JSPI smoke had been rewriting `python-worker.js` after the identity was generated, so the verifier correctly rejected the mutated worker hash. It now launches a test-only bootstrap worker that disables JSPI and dynamically imports the untouched, attested production worker;
- the WebKit mobile fallback smoke had the same invalid worker-rewrite technique in its `automatic-no-jspi` variant. It now delays a test-only bootstrap, verifies that redirect actually occurred, and imports the untouched production worker after disabling JSPI;
- optimized production/replay artifact uploads omitted `runtime-build-identity.json`, causing the later replay host to receive HTTP 404. Both artifacts now include the generated identity, and workflow path filters cover the provenance generator/verifier inputs.

The verifier itself was not weakened for any of these fixes.

## Boundaries / remaining integration

- no process-owner/R2 changes, retry, reroute, or bypass;
- no source-execution MCP registration;
- no new scene graph, scheduler, renderer, semantic authority, or serialization authority;
- byte identity is provenance, not malicious-server attestation or a sandbox guarantee;
- Docker/process isolation remains the already-landed R2 foundation;
- the persistent Docker runner should consume `ProvenancedPythonAuthoringClient` after this PR lands and bind its observed build identity into frame-artifact provenance. That runner/capture plumbing remains the next R4 integration slice rather than being duplicated here.

Per `AGENTS.override.md`, qualification uses repository CI and self-review only; no automated/Codex review is requested or relied upon.


yongkyuns:
Cross-workstream observation from the current exact-head cross-browser failure: normal Chromium gallery cases pass through `compatible-text-family-reveal`; the failures are specifically forced **no-JSPI** variants (`compatible-timed-composition`, `parity-moving-dots`, `parity-rotation-updater`, `compatible-indicate-square`), each surfacing `Runtime failed`. Firefox normal matrix passes; Chromium/WebKit fail at the no-JSPI gallery step. This strongly localizes the regression to the generated worker/provenance bootstrap interaction with the no-JSPI continuation path rather than general rendering or build-identity generation. Please preserve the exact current failure artifacts when repairing; #1386 will consume this provenance path only after it is qualified. No automated review requested.

PR #1268, opened 2026-09-08
Integrate preview PNG capture with bounded artifact storage

yongkyuns:
## Scope

#1197 / R4b capture→artifact integration. #1202 and #1266 are already on master. This remains a bounded host-tooling slice: actual preview PNG bytes enter the existing bounded `FrameArtifactStore` with source/time/backend provenance. It does not touch #1246/R1 process ownership, isolation, rendering MCP exposure, or engine semantics.

## Current master reconciliation — 2026-09-10 UTC

Current head: **`d0b51b7941c67a2e0c9bb962e5efce00961a4434`**. Current master **`7edd6b41feee02355943868d48754bf0e23a706b`** is the second parent of this non-force reconciliation. The compare remains exactly six intended files:

- `.github/workflows/manim-raster-differential.yml`
- `.github/workflows/noon-agent-foundation.yml`
- `scripts/agent-preview-capture.mjs`
- `scripts/agent-preview-capture.test.mjs`
- `scripts/agent-preview-capture-smoke.mjs`
- `web/manim-raster-host.js`

The raster workflow preserves current-master semantic-differential coverage and adds only the R4b capture contract/smoke hooks. No old master-only workflow behavior is reintroduced.

## What changes

- `PreviewFrameArtifactSession` computes the submitted-source SHA-256 itself, opens one private `FrameArtifactStore` scope, and retains already-captured PNG bytes with observed requested/published time and renderer backend.
- Build, scene and frame revisions remain explicit `null` until genuinely observed. No checkout/git SHA is substituted for loaded-build evidence.
- `web/manim-raster-host.js` preserves historical `time` and exposes the `requestedTime` / `publishedTime` already recorded by `SemanticPreviewSession`.
- Deterministic unit coverage checks source identity, observed metadata, independent artifact ownership, closed-scope cleanup, failed/unpresented samples, accessor-backed sample rejection and source bounds.
- Real Playwright WebGL2 qualification screenshots the actual canvas, stores those exact bytes through `FrameArtifactStore`, reads them back byte-for-byte, verifies provenance, writes evidence, then closes ownership.

## Qualification

The previous `4e98f687...` combination had all eight workflows green, including the real raster capture→store smoke. Because master advanced substantially afterward, that evidence is historical only. Exact-head CI for **`d0b51b...`** is running now. Merge requires those current-head gates to pass and a current self-review of the six-file delta. Per repository policy, no Codex/automated review is requested or relied upon.

## Limits

[...]

PR #1379, opened 2026-09-10
Add bounded transport-independent preview session registry

yongkyuns:
Advances #1197 / R3 without touching blocked R1/#1246 or exposing source execution through MCP.

This draft adds an ownership/control registry over an injected preview-session factory compatible with `SemanticPreviewSession`. It owns only transport scopes and opaque session IDs; all rendering/authoring semantics remain delegated to the existing preview session.

Implemented contract:
- bounded global/per-scope sessions and bounded transport scopes;
- private in-process scope capabilities plus opaque UUID session IDs;
- cross-scope and stale session rejection even when an ID is known;
- open returns only after the injected session's first-ready result;
- forward sample and coherent inspect delegate to the owned session;
- conflicting in-flight operations are rejected before duplicate engine work;
- AbortSignal cancellation closes and invalidates the owned session;
- transport disconnect rejects even a non-cooperative pending open and closes all owned sessions;
- registry shutdown closes all scopes/sessions;
- failed opens release quota so a fresh session can recover;
- invalid inputs/factories/reasons do not partially publish ownership.

The committed focused tests cover all of the above. The existing `noon-agent-foundation` workflow is currently being changed by #1268, so this PR intentionally does **not** modify that overlapping file yet. Once #1268 lands, this branch will reconcile onto master and add these tests to the existing agent-foundation job rather than creating a new workflow category.

This is session ownership/control only. It is not the R2 sandbox, browser/process launcher, CLI, MCP registration, scene graph, scheduler, renderer, or source rewriting. No execution surface is safe to expose until R1/R2 qualification. Per `AGENTS.override.md`, no Codex/automated review is requested or relied upon.

### DIFF
--- a/scripts/agent-preview-sessions.mjs
+++ b/scripts/agent-preview-sessions.mjs
@@ -1,0 +1,216 @@
+import { randomUUID } from "node:crypto";
+
+export class AgentPreviewSessionRegistry {
+  #createSession;
+  #scopes = new Map();
+  #sessionCount = 0;
+  #disposed = false;
+  #limits;
+
+  constructor({ createSession, maxScopes = 32, maxSessions = 64, maxSessionsPerScope = 16 } = {}) {
+    if (typeof createSession !== "function") {
+      throw new TypeError("session registry requires a preview-session factory");
+    }
+    for (const [name, value] of Object.entries({ maxScopes, maxSessions, maxSessionsPerScope })) {
+      if (!Number.isSafeInteger(value) || value <= 0 || value > 65_536) {
+        throw new RangeError(`${name} must be a positive bounded integer`);
+      }
+    }
+    if (maxSessionsPerScope > maxSessions) {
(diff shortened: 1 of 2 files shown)


yongkyuns:
R2 integration coordination: the current Docker isolation owner in #1383 necessarily has **asynchronous close/cleanup** because it must `docker rm --force <exact-container-id>` and surface cleanup failure. This registry currently calls `entry.session.close(reason)` synchronously inside `#release`, and `closeScope()`/`dispose()` do not await returned promises. Before R3 is used for real execution, please make the ownership contract async-safe: cancellation/disconnect/shutdown must not report completed cleanup or drop the only cleanup result while a container removal is still pending. Keep rejection/accounting atomic, but preserve a bounded awaited cleanup path (or an explicit pending-cleanup result that the outer runner must await). No need to touch #1383 files; I’m keeping R2 separate.

Issue #74, opened 2026-08-24
[Roadmap B1] Core Mobject/VMobject family, query, layout, and state semantics


yongkyuns:
## Roadmap

Phase B1 parent under #954. Architecture authority: `docs/architecture.md`.

## Goal

Complete common Manim-style object/family/query/layout/state semantics on the authoritative Semantic Scene while preserving the architecture's authored/base vs effective-runtime distinction.

## Query contract

The Semantic Scene owns authored/base state. The Runtime owns the latest coherently published effective state after active timeline/reactive/host/editor drivers.

Normal live compatibility getters such as center/edge/size/transform queries observe the latest coherent **effective** state. An explicit authored/base inspection API may exist, but it is a different query and must not silently substitute for the live value.

Layout or authoring operations that intentionally edit persistent state consume/reconcile the appropriate authored values through shared semantic operations.

### B1.1 — family identity/order
- persisted ordered family membership/traversal;
- add/remove/insert/reorder and indexing/slicing;
- alias/reference behavior where observable;
- family-aware painter/order semantics.

### B1.2 — queries/transforms
- center/edge/corner/start/end/critical-point queries;
- width/height/depth-compatible getters/setters;
- `match_*`, `set_x/y/z`, `set_coord`, stretch/fit, flip and rotation helpers;
- live getters observe coherent effective state;
- persistent mutators update authored/base state through shared semantic operations.

### B1.3 — layout
- `next_to`, `align_to`, `to_edge`, `to_corner`;
- `arrange` / `arrange_in_grid`;
- layout-accurate semantic bounds from #62, never frontend approximations;
- define explicitly whether an operation uses authored/base or current effective state as its input when active drivers are present.

### B1.4 — state/target/copy
- copy with new semantic identity and shared immutable resources;
- `generate_target`, save/restore, `become`, `replace`;
- every supported mutator participates automatically in shared target-state/`.animate` behavior;
- activation-time animation semantics may capture current effective state where required.

### B1.5 — style/order
- family style propagation;
- z/painter ordering;
- fill/stroke/object opacity and stroke scaling through #62.

## Acceptance

- representative ManimCE v0.21 family/layout/state/query cases run through shared semantic handles;
- family membership/order exists only in Semantic Scene state;
- Python and Rust equivalent operations produce equivalent observable results;
- live getters return coherent effective values during/after active drivers;
- authored/base inspection, where exposed, is explicit;
- `.animate`/Transform use shared target-state semantics and correct activation-time state;
[...]

PR #1381, opened 2026-09-10
Complete common paint queries and per-member gradients

yongkyuns:
`get_color()` previously always selected stroke and returned paint opacity inside the color. That disagreed with pinned Manim's visible-fill selection and could change opacity when a caller set the queried color back. Shared authored/live color queries now return RGB independently of opacity. Ordinary fill/stroke RGB and stroke-width getters are available through the same Rust query path.

This advances #74 under #954 and is stacked on #1380. It also adds per-member `set_color_by_gradient` / `set_submobject_colors_by_gradient`, with interpolation, ordered unique-leaf selection and atomic style publication in Rust. A one-leaf source receives the first stop; multiple leaves include the endpoint colors. Empty/invalid inputs fail before mutation. Native Rust exposes the reusable `color_gradient` value helper.

The Semantic Scene owns authored paint and family identity; LiveSession reads current effective paint and publishes persistent updates through the existing mutation lane. Gradients reuse the family style transaction with an indexed callback. Work is proportional to color stops plus affected unique leaves; resources, geometry and unrelated objects remain unchanged. No persistent frontend palette/state mirror, renderer gradient engine, transport boundary or migration seam is introduced. Python only converts arguments and dispatches typed calls.

Reference semantics: [pinned VMobject paint queries](https://github.com/ManimCommunity/manim/blob/v0.21.0/manim/mobject/types/vectorized_mobject.py) and [pinned per-member gradients](https://github.com/ManimCommunity/manim/blob/v0.21.0/manim/mobject/mobject.py). The existing live-geometry example now checks RGB and opacity separately while keeping its rendered opacity unchanged.

Paired native/direct-WASM Rust and Python examples exercise aliased family gradients, paint queries, and color round trips through normal retained rendering. Pinned semantic/raster fixtures qualify selection, interpolation and opacity. Existing callback capture limitations, background strokes, non-rendered family style, per-pixel gradient paint resources, and text's per-glyph styling remain with their owning B1/B4 work; this PR does not claim those capabilities or complete B1.

Validation:
- `CARGO_TARGET_DIR=/Users/ykshin/Dev/me/noon/target CARGO_PROFILE_TEST_DEBUG=0 CARGO_BUILD_JOBS=2 cargo test -p noon --no-default-features --test paint_queries_gradients` — 4 passed, including the paired session.
- Same environment, `cargo test -p noon --no-default-features --lib manim` — 4 passed.
- Same environment, `cargo test -p noon --no-default-features --lib live_line` — 2 active-runtime query tests passed.
- `PYTHONPATH=web/python python3 -m unittest discover -s web/python` — 205 tests, 180 passed, 25 existing host-dependent skips.
- Shared target/two jobs, `bash scripts/check.sh fmt-lint` — architecture/format/workspace check/clippy passed before the final existing-example/test expectation adjustments.
- `cargo check -p noon-web --target wasm32-unknown-unknown` — passed before the final example/assertion adjustments.
- Changed JavaScript syntax, differential Python compilation and `git diff --check` passed.

Full local linking remains disk-limited; required GitHub Rust/browser/native/provider/reference/product checks qualify this head. No local full-pass claim or premature support promotion.


yongkyuns:
Paint-query and gradient commits, examples and reference cases are now preserved in the integrated #1378 branch. Closing the separate PR to qualify overlapping B1 work once. #1378 owns completion and merge.

PR #1402, opened 2026-09-10
Share vector subpath and Bezier construction with endpoint queries

yongkyuns:
Custom vector objects can now start disconnected subpaths, append line/quadratic/cubic segments, and close or extend contours through shared Rust authoring operations. First/last endpoint queries include unfinished anchors; length-based proportion sampling remains a separate observation. This advances #78 under #954 and uses merged #1401.

Both Mobject and LiveSession prepare new immutable path content, then use #1401's existing semantic content transaction and coherent publication. Appends preserve the previous world-space geometry and paint; extending a closed contour retains its closing edge and earlier subpaths. The shared path query keeps a derived endpoint snapshot alongside its existing prepared curve measure. There is no frontend path engine, new semantic identity, runtime, mutation protocol or transport seam. Preparation is proportional to the affected path and reuses the normal resource/lowering/renderer path.

Paired `path_construction` examples exercise disconnected subpaths, quadratic/cubic curves, closure and live contour extension through Rust native/direct-WASM and Python. Pinned Manim semantic/raster fixtures cover these operations and unfinished anchors. Python also frees every query snapshot.

Validation:
- `cargo test -p noon --all-features --test path_construction --test path_editing --test path_queries`: 13 passed.
- `PYTHONPATH=web/python python3 -m unittest web/python/test_path_editing.py web/python/test_path_queries.py`: 5 passed.
- `NOON_WASM_PROFILE=dev bash scripts/check.sh full`: architecture, format/check/clippy, all Rust workspace tests and Python/JavaScript checks passed. The generated-export check caught a stale local WASM artifact from the shared worktree cache.
- After invalidating affected local source timestamps, `NOON_WASM_PROFILE=dev bash scripts/check.sh web` passed, including the WASM build and 221 generated package contracts.
- `node scripts/manim-differential-noon.mjs /tmp/noon-path-construction-observations.json`: 75 shared Rust/Python semantic fixtures executed successfully.

GitHub raster/browser/provider checks remain required before merge. Remaining #78 work includes broader anchor/handle/subpath and partial-path observations, reverse/subdivide/alignment, smoothing, booleans and derived morph/reveal queries. Transient callback resource production/reclamation remains C4/#368. This does not close Phase B.

The shared stroke tessellator now handles exact line retracing under miter joins with a finite bevel at the reversing endpoint, preventing Lyon's zero-normal cusp from collapsing stroke width. Other joins and progress attributes are preserved. The regression checks full-width triangle coverage for open and closed retraces; the integrated geometry suite passes (47 library tests plus integration tests).

Focused rendering against the exact pinned CI reference confirms the defect is repaired on both WebGPU and WebGL: max channel error falls from 255 to 76 and mean error from 0.112476 to 0.068945. Remaining differences follow antialiased edges; bounds differ by at most one pixel. The new path-construction fixture records differing-ratio 0.0038 (measured 0.00369985), tightens mean error to 0.07, and retains its one-pixel bounds limit. No existing landed fixture tolerance changes.

Post-fix validation: `cargo test -p noon-geometry` passed; `cargo clippy -p noon-geometry --all-targets -- -D warnings` passed in the isolated geometry target. Web preflight passed and the corrected development package build passed all 223 generated contracts (`NOON_WASM_PROFILE=dev NOON_RENDERER_SMOKE=1 NOON_SKIP_WEB_PREFLIGHT=1 bash scripts/check.sh web`). The final package-only invocation reused that completed preflight. Focused local raster compared the same CI reference PNG on both backends. A mistakenly selected release build was cancelled; the installed wasm-pack does not accept --no-opt, so the successful dev build omitted that flag.


yongkyuns:
This root-owned #78 batch is being consolidated into #1405 together with the dependent path-query/spline work. #1405 contains these commits and will receive the combined full qualification. Redundant CI runs here are being cancelled to reduce the runner queue. This PR remains open until the code lands through #1405.

yongkyuns:
Included in the qualified combined path implementation merged as #1405.

PR #1403, opened 2026-09-10
Share partial-path selection and contour reversal across Rust and Python

yongkyuns:
Partial-path extraction and curve reversal now use shared Rust semantic operations from both Rust and Python. `pointwise_become_partial` accepts a finite ordered interval `0 <= a <= b <= 1`, replaces the destination's world-space points, and retains its identity, paint and priority. `reverse_direction` reverses this object's curves and subpath order while preserving contour closure. This advances #78 under #954 and builds on #1401/#1402.

Geometry owns the curve algorithms. Authored partial extraction reuses the existing reveal algorithm while retaining degenerate boundary curves, because later curve-count selections observe them. Render-only reveal continues omitting those zero-length curves. Subpath breaks, reversed cubic controls and implicit closing edges are preserved. Empty-source partial edits are no-ops except for full-range replacement. Invalid intervals fail before resource allocation.

The Semantic Scene owns persistent results through the existing immutable-resource and atomic mutation transaction. Live operations capture coherent runtime operands and reject active render overrides through the established capture contract. Python only adapts arguments and dispatches to typed shared methods; the WASM authored operand supports self-aliasing without a mutable JS borrow conflict. No migration seam, second authority or new mutation protocol is introduced. Work is proportional to the selected path, with no unrelated-scene traversal.

Paired `path_selection` examples are registered for Rust native windows, direct Rust/WASM and Python execution. A pinned Manim fixture compares disconnected cubic/line extraction and reversal. Broader anchors/handles, get_subcurve/wrapping, subdivision/alignment, smoothing and boolean operations remain in #78.

Validation:
- `cargo test --workspace --all-features --test path_selection`: 5 passed, including active-render-override rejection.
- `cargo test -p noon-geometry --lib`: 46 passed before the parent cusp fix; the integrated stack now passes 47 library tests plus geometry integration tests.
- `PYTHONPATH=web/python python3 -m unittest web/python/test_path_editing.py`: 5 passed.
- `NOON_WASM_PROFILE=dev bash scripts/check.sh full`: the consolidated #1404 stack passed architecture, format/check/clippy, every workspace test and web validation, including this PR's final rejection regression.
- `node scripts/manim-differential-noon.mjs /tmp/noon-path-selection-observations.json`: 76 real WASM/Python semantic probes executed successfully.
- Generated package validation passed 222 contracts for this batch and 223 for the consolidated stack after the parent renderer correction.

This head includes the parent #1402 exact-retrace miter repair and reviewed new-fixture baseline, verified by focused WebGPU/WebGL comparison against the pinned CI reference. It also aligns the paired live-geometry example's Rust/Python fill inputs; its focused native continuation test passes. GitHub qualification remains required before merge.


yongkyuns:
This root-owned #78 batch is being consolidated into #1405 together with the dependent path-query/spline work. #1405 contains these commits and will receive the combined full qualification. Redundant CI runs here are being cancelled to reduce the runner queue. This PR remains open until the code lands through #1405.

PR #1404, opened 2026-09-10
Share path control queries and deterministic curve subdivision

yongkyuns:
Vector objects now expose curve counts, cubic control points, start/end anchors and anchors/handles through shared Rust path snapshots. `insert_n_curves` subdivides existing curves while preserving their shape, discontinuities, closed joins and unfinished anchors. This advances #78 under #954 and builds on #1402/#1403.

The geometry layer promotes line/quadratic controls exactly and uses Manim's integer curve-count distribution for subdivision. The Semantic Scene publishes persistent refinement through the existing immutable-resource and atomic mutation operation; live queries continue using one coherent effective snapshot. Python adapts scalar/array results, frees each query owner and routes mutations to Rust. Native and direct-WASM execution use the same operations. No new scene model, query authority, runtime or migration seam is introduced; work and storage follow the selected path and requested output size.

Public surfaces: `curve_count` / `get_num_curves`, `curve_points` / `get_nth_curve_points`, start/end anchors, anchors/handles, and `insert_n_curves`. The paired `path_refinement` examples show an original curve, its subdivided copy and dots placed from the shared anchor query. Native, direct-WASM, Python and pinned Manim fixtures are registered. Further subpath/subcurve APIs, alignment, smoothing and boolean operations remain in #78.

Validation:
- `cargo test --workspace --all-features --test path_refinement`: 5 passed, covering exact quadratic promotion, coherent snapshots, count distribution across discontinuities, closure/singletons, cubic polynomial invariance, identity/resource reuse and the paired runtime example.
- `PYTHONPATH=web/python python3 -m unittest web/python/test_path_queries.py web/python/test_path_editing.py`: 9 passed, including owner release and argument errors.
- `NOON_WASM_PROFILE=dev bash scripts/check.sh full`: passed architecture, format/check/clippy, all workspace tests, Python/JavaScript checks and 223 generated package contracts on the consolidated path stack.
- `node scripts/manim-differential-noon.mjs /tmp/noon-path-refinement-observations.json`: all 77 real WASM/Python semantic probes executed successfully.
- After the isolated parent #1402 miter-cusp repair: integrated `cargo test -p noon-geometry` passed 47 library tests and all geometry integration tests. The development web package passed 223 contracts, and focused WebGPU/WebGL raster comparison confirmed the corrected retrace against the exact CI reference. The unchanged full suite was not repeated for that isolated renderer fix.

[...]

PR #1405, opened 2026-09-10
Share VMobject path editing, queries and atomic spline operations

### LINKED ISSUE #78: [Roadmap B2] VMobject paths, partial paths, smoothing, and boolean geometry

yongkyuns:
Custom VMobjects can now construct and edit retained paths, select/reverse/refine curves, inspect coherent path structure, and smooth or straighten individual paths and families through shared Rust semantics. This advances #78 under #954. It consolidates #1402, #1403 and #1404 into one qualification run; those PRs remain open until this code lands.

The public Rust/Python surfaces cover:
- subpath start/line/quadratic/cubic append, closure and unfinished-anchor endpoints;
- partial-path replacement and curve/subpath reversal;
- curve counts, cubic controls, anchors/handles, subpaths and closure observations;
- deterministic shape-preserving subdivision and detached subcurve copies, including closed-seam wrapping;
- smooth/jagged anchor modes and smooth-corner construction, including atomic unique-leaf family edits.

Geometry owns curve extraction, reversal, subdivision and the linear-time f64 natural/periodic spline solver. The Semantic Scene owns all persistent results through the existing immutable-resource and atomic transaction boundary. Runtime queries/copies capture one coherent publication and retain the established callback, pending-animation and render-override restrictions. Python only adapts inputs/results, preserves wrapper metadata and releases query owners. Work is proportional to selected paths/family leaves and requested output; no unrelated scene traversal, new identity allocator, runtime, mutation protocol, migration seam or in-process serialization boundary is introduced.

Copies retain paint/priority while receiving a new identity and insertion order. Full-range copies reuse immutable content. Family edits validate all candidates before publication and reclaim unpublished resources on allocation/commit errors. Explicit retained contour breaks and closure survive smoothing; the solver is not run per frame.

The renderer also fixes exact line-retracing miter cusps, which previously collapsed stroke width. A coverage regression protects open and closed retraces. Focused WebGPU/WebGL comparison against the exact pinned CI reference confirms the defect is repaired: maximum channel error fell from 255 to 76 and mean error from 0.112476 to 0.068945. The new path-construction fixture records the measured AA baseline (ratio limit 0.0038, mean tightened to 0.07, bounds retained at one pixel). No existing landed fixture tolerance changes.

Paired `path_construction`, `path_selection`, `path_refinement`, `path_subcurves` and `path_smoothing` examples are registered for native Rust, direct Rust/WASM, Python and pinned Manim semantic/raster qualification. Reference quirk: Manim v0.21's spline closure helper uses a signed start-coordinate tolerance that can classify exactly closed negative-coordinate contours as open. Noon preserves explicit topological closure and translation-invariant periodic smoothing; a Rust regression protects that behavior rather than copying the reference bug.

Validation:
- Construction: 13 focused Rust tests, 5 Python tests and WASM compilation passed.
- Selection: 5 focused Rust tests, Python dispatch tests and WASM compilation passed.
- Refinement: 5 focused Rust tests, 9 Python tests, the full local gate and 77 real WASM/Python semantic probes passed.
- Subcurves: 6 focused Rust tests, 5 Python query tests, the full local gate (224 WASM contracts) and 78 real WASM/Python probes passed.
- Smoothing: 5 focused Rust tests, 50 geometry library tests, 7 Python dispatch tests and `cargo check -p noon-web --all-features --target wasm32-unknown-unknown` passed.
- The consolidated `NOON_WASM_PROFILE=dev NOON_RENDERER_SMOKE=1 bash scripts/check.sh full` is running after integrating landed text source-part work (#1371). Final combined probes and GitHub qualification remain required before merge.

Remaining #78 work includes path alignment, booleans and effective morph/reveal queries. Transient callback resource production/reclamation remains C4/#368. This does not close #78 or Phase B.


yongkyuns:
The combined head passed every active check except the newly added subcurve fixture's provisional pixel-ratio limit. I inspected the downloaded reference, WebGPU result and difference images from run 34458591166: both backends have exactly the same bounds, centroid, foreground pixel count (2,430), duration and semantic state as the reference. Differences follow the antialiased stroke edges; the result is the intended square and selected wrapped subcurve.

Measured on both backends: differing ratio 0.0031134259259259257, mean absolute channel error 0.03922260802469136, maximum channel error 65. The **new** fixture is calibrated to ratio 0.0032 while tightening mean error from 0.1 to 0.04 and bounds from 1 pixel to zero. No landed fixture baseline or implementation changed. Replaying `NOON_MANIM_RASTER_ARTIFACTS=/tmp/noon-path-completion-raster node scripts/manim-raster-enforce.mjs` against the complete downloaded report passes all fixtures.

Full local validation and all 79 real WASM/Python probes passed on the identical Rust/Python implementation; a new CI run qualifies the manifest-only adjustment.

Issue #79, opened 2026-08-24
[Roadmap B2] SVG and image resource support


yongkyuns:
## Roadmap
Phase B2 child under #954. Architecture authority: `docs/architecture.md`.

## Goal
Support common SVG and raster-image authoring as immutable semantic resources that participate in the same scene, bounds, transform and retained-rendering architecture as native geometry.

### SVG
- supported file/string parsing;
- transforms/viewBox/aspect behavior;
- style inheritance, groups/subpaths and common primitives;
- stable imported element/subpath identity where needed by matching/hot reload;
- deterministic diagnostics for unsupported constructs.

### Raster images
- browser-supported PNG/JPEG/WebP-style resources;
- dimensions/layout/opacity/scene transforms;
- URL/blob/packaged-asset loading boundaries appropriate for browser authoring;
- resource cache/deduplication and lifecycle.

## Architecture constraints
- decode/parse at author/compile/resource-install time, never on the render frame path;
- imported content is immutable retained resource data referenced by semantic identity;
- repeated identical assets may share immutable resources;
- large image/SVG payloads never flow through per-frame semantic snapshots;
- imported SVG paths reuse B2 path/Create/Transform/matching semantics;
- frontend asset access adaptation must not become frontend-owned geometry/rendering semantics.

## Acceptance
- representative supported SVG/image cases author through shared semantic handles;
- repeated assets share immutable resource data where valid;
- SVG paths participate in bounds/layout/Create/Transform/matching through shared primitives;
- local transform/style changes do not reparse/redecode content;
- packaged and explicitly supported runtime asset-loading paths have browser tests;
- unsupported content fails explicitly rather than silently changing appearance;
- representative output passes B7 backend/raster qualification.

## Dependencies
A1/B1/#53 resource ownership and B2/#78 path semantics. C4/#368 later handles localized live asset replacement.

yongkyuns:
### Acceptance criteria amendment — exact Manim output parity

This breadth issue must use the exact-output gate from #185, not only API/semantic/example coverage. For **every slice promoted to supported**:

- add a source-equivalent ManimCE **v0.21.0** fixture to the canonical #176 raster + timeline corpus (no Noon-only geometry/style/timing tuning; only the minimum import/browser wrapper adaptation);
- compare semantic state **and rendered pixels** against real Manim at the canonical profile, with explicit/ratcheted raster tolerances and categorized diff artifacts;
[...]

Issue #53, opened 2026-08-24
[Roadmap A1/B2] Immutable content/resource arena and no geometry cloning


yongkyuns:
## Roadmap

A1/B2 immutable-resource infrastructure under #953/#954. Architecture authority: `docs/architecture.md`.

## Goal

Keep heavy content outside mutable authored/runtime state so paths, SVGs, text and future meshes are referenced by stable immutable resource versions rather than cloned through authoring, animation, execution and rendering.

## Ownership

- Semantic Scene owns authored content/source identity/specification and persistent resource references;
- compiler/resource preparation produces immutable versioned content artifacts;
- Runtime may select a current **effective resource version** for transient dynamic content without rewriting authored/base content every frame;
- Renderer owns backend packing/residency only and never semantic resource authority.

Cheap analytic primitives may remain compact analytic content; uniformity does not justify heap/path expansion.

### Resource identity/versioning
- stable/content-addressed identity where appropriate;
- immutable `ResourceVersion` or equivalent exact version identity;
- deterministic replacement/removal/deduplication;
- stale preparation result cannot install over a newer semantic/effective version.

### Geometry/resource references
- semantic/execution objects reference handles rather than embedding repeated heavy payloads;
- transform/morph plans reference source/target resources + prepared correspondence data;
- local transform/style/effective driver changes leave immutable content untouched.

### Persistent replacement
Persistent source/content edits use the shared semantic `ReplaceContent` path from #368 with local impact/lowering.

### Transient effective content
Host/native dynamic producers may install a versioned effective runtime resource result through #368 without advancing semantic revision every frame. If the final result must persist, reconcile it into authored/base state through semantic mutation.

### GPU lifetime/retirement
Replacing the live version does not invalidate resources referenced by in-flight submissions. Renderer/backend retirement uses submission/fence serials; old versions remain resident until safe to reclaim. Long-running churn has bounded reclamation/accounting.

### Instrumentation
Track resource count/versions, retained bytes, allocation/cloning/regeneration, dependent-slot invalidation, effective-version churn and deferred retirement.

## Acceptance

- one large path referenced many times is retained once where sharing is valid;
- transform/morph construction does not clone command arrays through layers;
- local transform/style/effective changes do not duplicate/regenerate content;
- persistent replacement and transient effective replacement use the same immutable version/lifetime machinery without conflating semantic revision;
- stale async results cannot overwrite newer versions;
- in-flight old versions survive until safe GPU retirement;
[...]

PR #1418, opened 2026-09-10
Add immutable raster image resource foundation

yongkyuns:
Advances #79 under #954 with a deliberately non-overlapping B2 raster-resource foundation.

This first slice adds a backend-neutral `RasterImageResourceArena` for canonical tightly-packed RGBA8 pixel payloads with exact visual-content deduplication, intrinsic-dimension metadata, stable arena-qualified handles, retained-byte/pixel-byte instrumentation, explicit validation errors, and clone/re-namespace tests that preserve shared immutable payload allocations. The resource contract is exported from `noon-core` but is not yet wired into semantic object content or rendering.

The canonical-pixel choice follows pinned ManimCE v0.21 `ImageMobject`: both file and ndarray inputs are normalized into an RGBA pixel array before image behavior. PNG/JPEG/WebP parsing/decoding therefore remains an off-frame loader/resource-preparation concern rather than renderer semantic state. Sampling/resampling policy also remains object/renderer-facing metadata, not part of immutable pixel identity.

Architecture boundary: this PR owns immutable resource identity/cache data only. It does **not** touch #78 SVG/path semantics, #1416 unequal-family Transform correspondence, renderer texture residency, public `ImageMobject`, URL/blob loading, or browser/native decoders. There is no migration seam or second scene/runtime/renderer authority.

Locality: repeated identical pixel content shares one immutable allocation; transform/style/runtime work is outside this resource contract and cannot cause image decoding or pixel duplication here. Content lookup uses a hash index so admission is linear in payload size rather than byte-wise tree comparison across the resource set.

Validation in this slice is focused Rust unit coverage for deduplication, dimension-sensitive identity, invalid admission atomicity, arena provenance, immutable payload sharing across cloned owners, and accounting. GitHub qualification provides the repository-level executable gate because this session has no mounted/networked Noon checkout.

Base at branch creation: `9064df5008800e13f5d3568fee354eaf5001854a`.

Follow-up after this foundation qualifies: SemanticStore/image-content ownership plus decoder/authoring integration, then retained renderer texture installation and browser/package-loading tests. SVG remains blocked on the active #78 path work rather than being mixed into this PR.

### DIFF
--- a/crates/noon-core/src/resources/image.rs
+++ b/crates/noon-core/src/resources/image.rs
@@ -248,8 +248,7 @@ fn expected_rgba8_len(width: u32, height: u32) -> Result<usize, RasterImageResou
         .checked_mul(u64::from(height))
         .and_then(|pixels| pixels.checked_mul(4))
         .ok_or(RasterImageResourceError::ImageTooLarge { width, height })?;
-    usize::try_from(pixels)
-        .map_err(|_| RasterImageResourceError::ImageTooLarge { width, height })
+    usize::try_from(pixels).map_err(|_| RasterImageResourceError::ImageTooLarge { width, height })
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
--- a/crates/noon-core/src/resources/image.rs
+++ b/crates/noon-core/src/resources/image.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::BTreeMap,
+    collections::HashMap,
     mem::size_of,
     sync::{
         atomic::{AtomicU64, Ordering},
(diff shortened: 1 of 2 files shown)


yongkyuns:
Self-review note: the final foundation shape retains canonical tightly-packed RGBA8 pixels, not encoded PNG/JPEG/WebP bytes. This matches pinned ManimCE v0.21's normalization of file/ndarray inputs to an RGBA pixel array and avoids making platform decoder behavior part of renderer semantics.

Known boundary: this arena is append-only in this slice. It establishes immutable content identity/dedup/accounting/provenance only; it does **not** claim #79's resource-lifecycle/reclamation acceptance. Explicit long-running asset reclamation/churn remains coordinated with #53 before #79 can close. Likewise, resampling policy, persistent pixel replacement, semantic image content, loaders and GPU texture residency remain follow-ups.

PR #1425, opened 2026-09-10
Add identity-free derived display publication

yongkyuns:
## Summary

Adds the next bounded B3 layer for unequal-family `Transform`, stacked on #1421.

- introduces `DerivedDisplayObjectState`, an identity-free renderer-facing effective row that reuses retained content/resources without entering `FrameState.objects` or `CompiledScene`;
- introduces `DerivedDisplayObject`, which carries only an existing stable-slot anchor for painter placement plus a deterministic occurrence ordinal — the anchor is provenance/order, not the copy's identity;
- extends `RendererPublication` additively with a borrowed derived-display slice; ordinary runtime publications still default to an empty slice;
- makes attachment fallible and preflighted: anchor range/presence, unique occurrence ordinals, finite transform/style/z-index, and normalized appearance/reveal/morph are checked before publication is enriched;
- keeps semantic topology, stable execution slots, and persistent `become()` state untouched.

## Qualification

The preflight implementation at `439feb2bd49ac8e2b4ec401989a9a07f5af7c3bf` passed all 5 focused `noon-runtime` derived-display tests. That probe identified exactly one rustfmt line-wrap; `79bf6f242e3dc4360d0afdb78e2df737afebdca9` applies that formatting correction and no semantic change. The temporary diagnostic workflow has been removed from the final PR diff.

Final production diff: only `crates/noon-runtime/src/renderer_publication.rs`.

## Stack / remaining B3 work

Base: #1421 / `codex/b3-unequal-family-transform`.

This still does **not** claim end-to-end unequal-family Transform rendering. #1426 consumes this publication shape in renderer-owned transient buffers; the remaining activation layer must materialize derived rows from #1421's correspondence descriptors at animation evaluation time and retire them at segment completion.

Refs #82.

PR #1421, opened 2026-09-10
Add derived unequal-family Transform correspondence

yongkyuns:
## Summary

Adds the first bounded B3 layer for unequal-family `Transform`, stacked on #1416.

- derives ManimCE-style repeat-index correspondence recursively across semantic families;
- marks repeated source/target occurrences as transient padding descriptors rather than creating semantic nodes;
- preserves authored family membership and scene revision during correspondence derivation;
- rejects empty-family and authored-alias cases fail-closed until their derived-display behavior is defined;
- leaves `ordered_family_leaf_pairs()` unchanged and strict.

This layer intentionally carries authored leaf references plus padding flags only. It does **not** create synthetic `SemanticNodeId` values, stable execution `ObjectId` values, or persistent topology.

## Qualification

Exact implementation `907bb74b3e37646184328076debce0eb87bde2f3` passed the focused compiler probe:

- `cargo test -p noon-compile family_transform --lib` — 6 passed, 0 failed;
- `cargo fmt --all -- --check` — passed.

The temporary probe workflow used to retain diagnostics has been removed from the final PR diff.

## Stack / remaining B3 work

Base is #1416 / `codex/b1-unequal-become` at `f08cd6301072204dbff627d08ec86a77ff07d949`.

This PR does **not** claim end-to-end unequal-family Transform rendering yet. The next stack layer must consume these occurrence descriptors into plan-local derived display instances through runtime publication/render preparation, then route unequal `FamilyTransformTo` through that lane. Those display instances must remain outside Semantic Scene identity and stable execution-object identity; persistent topology reconciliation stays owned by explicit `become()`.

Refs #82.

Issue #82, opened 2026-08-24
[Roadmap B3] Transform family and matching-parts semantics


yongkyuns:
## Roadmap

Phase B3 transform-family parent under #954. Architecture authority: `docs/architecture.md`.

## Goal

Complete high-value ManimCE v0.21 Transform variants on one shared target-state/morph/lifecycle implementation, with activation-time source state taken from the correct coherent live value rather than stale declaration-time state.

## State/activation contract

Authored animation/target intent belongs to the Semantic Scene. When a transform activates in a live session, source-side values that Manim observes as “current” are captured from the coherent **effective Runtime state** at activation time.

Target construction may use authored target/state operations, but the source snapshot for interpolation must not silently fall back to stale authored/base state while another driver has already changed the object. Persistent final results reconcile through the normal authored-state contract when required by the animation/lifecycle semantics.

## Scope

- MoveToTarget;
- ApplyMatrix and deterministic geometric transform helpers;
- FadeToColor, ScaleInPlace, ShrinkToCenter;
- Restore using B1 save/target state;
- Swap/CyclicReplace;
- clockwise/counterclockwise/path-arc options;
- deterministic path-function variants where representable;
- TransformMatchingParts/Shapes;
- TransformMatchingTex once B4 source/part identity exists;
- preserve/harden Transform, ReplacementTransform and TransformFromCopy as common substrate.

Arbitrary Python ApplyFunction/ApplyPointwiseFunction/ApplyComplexFunction behavior is classified explicitly: lower supported deterministic forms when possible; otherwise use C1 staged host-callback semantics rather than pretending the function is a deterministic seekable track.

## Architecture constraints

- target/state construction uses shared B1 semantic operations;
- activation-time current source values come from coherent effective Runtime state;
- morph/path correspondence is prepared below the semantic layer using immutable resources;
- matching uses stable semantic/resource/source-part identity rather than frontend list position alone;
- lifecycle/source-target membership is shared semantic behavior;
- deterministic transforms require no Python during playback;
- host-function variants may emit effective driver/resource output and persistent semantic mutations according to #70/#368; they do not get a second transform engine.

## Acceptance

- representative transform variants match source/target/lifecycle behavior;
- chained/overlapping activation starts from the visible effective source state;
- MoveToTarget/Restore use explicitly defined B1 snapshot/state semantics;
- path-arc/direction and intermediate effective morph states are differentially qualified;
- matching-shape/text cases reuse one matching/morph pipeline;
[...]

PR #1426, opened 2026-09-10
Pack identity-free derived analytic display rows

yongkyuns:
## Summary

Adds the renderer-side B3 layer for transient unequal-family `Transform` copies, stacked on #1425.

- packs identity-free derived Circle/Rectangle/Line rows into renderer-owned transient instance arrays with **no `ObjectId` side tables**;
- emits `PreparedDerivedDisplaySlot` rows keyed only by stable source anchor + plan-local occurrence ordinal;
- composes painter order as stable anchor followed by its derived occurrences in deterministic occurrence order, without relocating or extending stable `FramePreparer` slots;
- requires the copy to retain the source anchor's z-index for this ordinary Transform slice;
- fails closed for text, VectorPath and external geometry until their retained-resource/tessellation transient lane exists, rather than fabricating a stable execution/render identity;
- preserves existing stable render preparation unchanged.

Focused renderer tests on the production implementation passed before stack cleanup, together with full rustfmt. The branch has now been reconstructed onto #1425's formatted final production head without changing `render_order.rs`; the temporary renderer probe workflow is absent from the final tree.

Final production diff: only `crates/noon-render-wgpu/src/render_order.rs`.

## Stack / remaining B3 work

Base: #1425 / `codex/b3-derived-display-publication`.

This PR does not yet activate unequal-family `Transform`. The remaining integration layer must materialize these rows from #1421's correspondence descriptors at animation evaluation time, attach them to the coherent `RendererPublication`, and retire them at segment completion. Path/text support remains a bounded follow-on before broad family Transform can be marked supported.

Refs #82.

### DIFF
--- a/crates/noon-render-wgpu/src/render_order.rs
+++ b/crates/noon-render-wgpu/src/render_order.rs
@@ -803,3 +803,362 @@ mod tests {
         assert_eq!(preparer.visible_projection_stats().projections, 2);
     }
 }
+
+/// Packed primitive kind for one identity-free derived display occurrence.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DerivedDisplayPrimitive {
+    Circle,
+    Rectangle,
+    Line,
+}
+
+/// One lookup row into the transient derived-instance buffers.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct PreparedDerivedDisplaySlot {
+    pub anchor_object_index: u32,
+    pub occurrence_index: u32,
+    pub primitive: DerivedDisplayPrimitive,
+    pub instance_index: usize,
+}
+
+/// Painter-level merge descriptor. Stable entries retain their execution slot;
+/// derived entries retain only their plan-local occurrence ordinal.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DisplayPainterItem {
+    Stable { object_index: u32 },
+    Derived { occurrence_index: u32 },
+}
+
+/// Transient packed analytic geometry for one renderer publication.
+///
+/// These arrays deliberately contain no `ObjectId` side tables. They are rebuilt
+/// from the publication's identity-free derived rows and discarded independently of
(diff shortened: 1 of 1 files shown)


yongkyuns:
Merge-readiness reconciliation after clean restack onto the latest master:

- current head `48aa64920ff334750f77bd001e06f26f5990392e` is exactly 1 commit / 1 renderer file (`crates/noon-render-wgpu/src/render_order.rs`) / +359/-0 on master `9640d71581d9a6514b3c3a1030212f2942b5b698`;
- current master’s pre-feature `render_order.rs` blob is byte-identical to #1426’s original parent, so this is an exact feature replay with no overwritten renderer work;
- the production blob is exactly the previously focused-qualified final implementation (`731d5cf048222866c66a9ecbd53865ccec99a800`), whose derived-display renderer tests and rustfmt passed before stack cleanup;
- historical exact-head qualification was green for Rust unit tests, fmt/clippy, web preflight, Architecture, Layer Dependency, Native Host, Provider isolation, Playground authoring recovery, Product Gate, and cross-browser;
- the sole historical PR Fast failure was the known stale direct `family_state` WebGPU sample mismatch (`direct family state at 0ms missed expected members`), after deterministic replay, typed error-boundary coverage, and Manim authoring had already passed. That same inherited probe failure was independently diagnosed/fixed in the B1 lineage and is not caused by this renderer packing slice;
- this PR does not activate derived rows yet: it only adds identity-free packing/preparation entry points. Ordinary direct browser scenes do not exercise the new path until later B3 activation layers;
- current master contains the corrected family-state/probe lineage, and the only intervening master advance after #1425 is disjoint physical-performance workflow/script work;
- fresh workflows on `48aa649...` are queued because of runner saturation, with no failure signal.

Given the exact tested one-file blob, broad/static historical green matrix, identified unrelated historical PR Fast blocker now resolved on current master, and the fact that the new renderer path is not yet activated, this is a justified qualification transfer. Safe to promote/merge the current head; fresh queued runs can remain post-merge confirmation.

Issue #1198, opened 2026-09-07
Noon MCP: thin local stdio adapter over shared capabilities and preview runner


yongkyuns:
Parent / cross-session entry point: #1194. Depends on #1195 and the qualified #1197 runner; no independent render implementation. `docs/architecture.md` remains the architecture authority.

## Current handoff — 2026-09-08

**Reopened because only the read-only discovery subset is complete.** Closing this issue after #1204 did not complete its rendering/session acceptance criteria; the [existing completion comment](https://github.com/yongkyuns/noon/issues/1198#issuecomment-5576689769) explicitly said to keep the issue open.

Merged subset: #1204, merge `fac4c9ae3b4411afe90414de433734e28da1a653`. The optional checkout-bound package at `tools/noon-mcp` implements `noon_capabilities` and `noon_reference` using the official SDK's local stdio transport. [Discovery CI 34169658806](https://github.com/yongkyuns/noon/actions/runs/34169658806) qualified locked installation, boundary tests and actual SDK-client interoperability on Linux/macOS. These tests do not qualify scene execution or rendering.

Rendering/session operations and rendered-image delivery are still unfinished. The immediate dependency is the runner work in #1197, including the [#1246 ownership blockers](https://github.com/yongkyuns/noon/pull/1246#issuecomment-5585216203). Do not expose source execution before the relevant isolation/cleanup gates are qualified.

**Next action after the runner contract is qualified:** extend the existing `tools/noon-mcp/src/server.mjs` rather than creating another server, protocol implementation, scene model or scheduler. Preparation of schemas/client tests may proceed against an explicit contract without registering an unsafe execution surface.

## Scope

Optional Node/TypeScript tooling package wrapping the same runner used by the CLI. Proposed operations: capabilities, reference lookup, open_scene, sample_frames, inspect, close_scene. Final names follow the implemented runner contract.

The existing server/discovery modules and package README already provide the discovery path and setup; extend them. Rendering operations must preserve runner validation, limits, errors, capabilities and provenance. Route request cancellation and client/server shutdown to the owned runner sessions, not just the outer response. Inspection starts with supported coherent status/frame observations; do not promote the trusted raster diagnostic into arbitrary public scene introspection or claim seek when the session reports it unavailable.

[...]

Issue #1199, opened 2026-09-07
Agent integration: reproducible packaging and deterministic/agent evaluation


yongkyuns:
Parent: #1194. Follows #1195, #1196, #1197, and #1198.

## Acceptance

- [ ] Package compatible skill, capability metadata, and runner versions with source hashes and clear environment requirements.
- [ ] Verify setup from a clean environment; do not install unrelated unpinned dependencies or execute installation hooks implicitly.
- [ ] Preserve provenance/license notices for any copied third-party material and independently review external scene/assets corpora.
- [ ] Fixed deterministic task corpus covers square-to-circle initial/middle/final state, sequential barriers, fades, supported composition/text, callbacks, unsupported plotting/LaTeX/3D, cancellation, and repeated runs.
- [ ] Semantic assertions plus backend-qualified raster evidence, not blank/nonblank-only checks; never weaken existing gates or widen tolerances to admit an agent fixture.
- [ ] Compare docs-only, skill+runner, and skill+MCP using the same tasks/model settings.
- [ ] Report semantic/visual correctness, unsupported-API suggestions, repair iterations, first-useful-frame latency, and token/tool overhead.
- [ ] Stochastic agent runs are separate reported evaluations, not flaky replacements for deterministic mandatory CI.

Initial distribution may require a Noon checkout. A distributable installed capability snapshot must identify its build and must not silently read unrelated/stale checkout metadata. Retrieval starts with the tested local example index; vector databases and nested agent orchestration are not prerequisites.

yongkyuns:
## Session claim — 2026-09-11

Claiming the first bounded #1199 slice: reproducible checkout-bound agent packaging/source identity only.

Scope:
- deterministic package builder/verifier using only existing repository/toolchain primitives;
- package the existing `skills/noon-authoring` files rather than defining a second skill;
- embed the canonical `scripts/noon-capabilities.py` output rather than creating a second capability matrix;
- manifest source hashes and the existing MCP/runner package identity, while explicitly keeping actual loaded worker/WASM/build identity runtime-observed per preview artifact;
- update stale skill/verification text to the now-landed CLI/MCP preview path;
- clean-environment/package determinism tests and CI as needed.

Out of scope for this PR: the fixed semantic evaluation corpus, docs-only vs skill+runner vs skill+MCP comparison, stochastic model runs, score reporting, or any new renderer/session/capability authority. Those remain subsequent #1199 slices.

[...]

PR #1427, opened 2026-09-10
Wire retained preview rendering into existing MCP server

yongkyuns:
Advances #1198 directly on the landed shared preview stack (#1387 CLI, AgentPreviewService/session registry, Docker runner, retained artifacts, and observed runtime provenance). The only execution authority remains `AgentPreviewService`; MCP is a transport adapter over that qualified path.

## Mapping and lifecycle

- registers `noon_open_scene`, `noon_sample_frames`, `noon_inspect`, and `noon_close_scene` on the existing `McpServer`;
- rendering registration is enabled only when trusted startup configuration explicitly supplies `NOON_PREVIEW_RUNTIME_CONFIG`; discovery-only startup remains unchanged otherwise;
- one `AgentPreviewService` is shared by the server process and each stdio server instance receives an opaque service scope;
- tool arguments cannot select checkout, Docker image, isolation profile, executable, or command;
- request AbortSignals flow into shared open/sample operations;
- the public sampling contract admits at most 31 samples per call, matching the landed 32-frame retained scope including the initial frame; cumulative capacity remains preflighted by the shared service before advancement;
- disconnect/SIGINT/SIGTERM await scope/session/container cleanup and service disposal;
- open/sample return standard MCP `image/png` blocks plus JSON-only structured snapshots/artifact descriptors with observed build/source/time/backend provenance;
- inspect is metadata-only; stale/cross-scope rejection remains owned by the shared registry/service;
- if artifact delivery fails after open/sample advances state, the adapter retires that session before returning the bounded structured error.

## Isolation qualification corrections

During qualification the old Docker workflow was found to be false-green because `node ... | tee ...` did not propagate the producer failure. This PR makes those preview pipelines fail closed and installs the exact locked MCP dependencies in the preview job so the real SDK client is actually exercised.

The real Chromium/Pyodide route also exposed two product requirements rather than papering over them:

- the preview image now contains the exact pinned Pyodide 314.0.5 distribution and the host routes only the matching CDN URLs to those local bytes, so `--network=none` remains an actual product boundary rather than relying on CI/network access;
- Noon derives the pinned Playwright seccomp profile to admit Chromium sandbox `chroot` at the syscall filter while retaining outer `--cap-drop=ALL`, `no-new-privileges`, read-only mounts/rootfs, private namespaces, and explicit seccomp verification; derivation fails closed if the upstream pinned rule shape changes;
- the default preview memory ceiling is 2 GiB. The previous 1 GiB limit was demonstrated to OOM-kill the real Chromium + Pyodide + Noon WASM workload; the higher value remains a hard cgroup bound while CPU, PID, shm and tmpfs limits remain bounded.

## Qualification

Focused adapter tests cover registration, cancellation propagation, image association/order, scoped artifact lookup, structured errors without stacks, and fail-closed retirement after image-delivery failure. Discovery-only stdio remains unchanged without preview configuration, while invalid configured runtime startup fails before protocol stdout.

The real Docker MCP client smoke opens actual source, verifies real PNG bytes and observed build/source/time/backend provenance, exercises a real source failure and same-transport recovery, rejects cross-transport and stale handles, rejects oversized sampling without advancing state, cancels a stuck post-first-frame sample and verifies authoritative cleanup/stale-handle recovery, reproduces deterministic fresh-run hashes, verifies a real source revision changes output, checks CLI/MCP artifact equivalence, and verifies disconnect cleanup. The pinned MCP SDK 2.0 wraps deliberate AbortSignal cancellation as `SdkError`/`REQUEST_TIMEOUT`, so the smoke keys its client-side assertion to an explicit abort reason while retaining the authoritative server-side cleanup, stale-handle and recovery checks.

No second server/protocol, runner, session registry, artifact store, scene model, scheduler, renderer, or HTTP transport is introduced. Repository CI/self-review only per `AGENTS.override.md`; keep draft until this exact head is fully green.


yongkyuns:
### Handoff — 2026-09-10/11 — current-base requalification queued

Active branch: `codex/agent-preview-mcp-rendering`
Current head: `493e04b4cfa20d8702d9ce945763d922f80dd96c`
Current base/master: `3240c2f1ac24155b2cb91d5d0bcc2c88c6c6251f` (#1411)

This continuation fixed three real Docker qualification blockers without weakening the isolation boundary:

1. `preview.Dockerfile` had an invalid nested shell quote in the Pyodide version check; replaced it with a direct `node -e` assertion.
2. The pinned `pyodide-core-314.0.5.tar.bz2` extraction assumed `bzip2`, which the pinned Playwright image did not contain; the image now installs the required extractor. The Pyodide archive remains SHA-256 verified (`f528dccea95fa8ec54295fd65bf86dd61183d11f0e52563dc8eadda45e0f78d6`) before extraction, and the Playwright base remains pinned by OCI digest.
3. Once the real sandboxed browser actually ran, cgroup evidence showed an OOM kill at the inherited 1 GiB memory ceiling (`oom_kill 1`, peak ~1.074 GB) while loading local Pyodide plus the current Noon WASM. The bounded default is now 2 GiB with `--memory-swap` equal to `--memory` (no swap). Network-none, private IPC/PID, read-only root/mounts, dropped capabilities, no-new-privileges, seccomp, tmpfs bounds, PIDs and CPU limits are unchanged.

`setup-preview-runtime.mjs` diagnostics were also corrected to retain the useful tail of Docker stderr. A speculative build retry was removed after the deterministic Dockerfile cause was identified.

Evidence before the final master integration: exact head `a5d77dd5b270d8c35e140e054b784150ca4d6b26` reached successful **Prepare content-addressed preview runtime**, **Verify isolated real-browser preview**, and **Verify persistent shared runner** under the 2 GiB/no-swap bound; GitHub then cancelled that stale run during the CLI step when the branch moved. Treat that only as diagnostic evidence, not the merge gate.

Master subsequently advanced via #1411. There was no direct file overlap with MCP/Docker code, but #1411 changes the browser/Python/WASM package consumed by the real preview smoke, so I merged current master into this branch through integration-only #1434 rather than relying on stale-base evidence.

Final-head qualification now running on `493e04b4…`:
- Architecture Ratchets `34553300148`: **passed**.
- Layer Dependency Ratchet `34553300233`: **passed**.
- PR Fast Gate `34553300226`: queued at handoff.
- Agent discovery MCP `34553300177`: queued at handoff.

No review threads are open. PR remains draft intentionally.

**Next concrete action:** when the two queued workflows execute, inspect the exact-head results. If the real Docker workflow passes isolation → shared runner → CLI → configured MCP smoke and PR Fast is green, re-read current master once, mark #1427 ready, merge with expected head `493e04b4…`, then update #1198 and start #1199 packaging/evaluation from the merged interface. If a job fails, fix only the observed blocker and requalify the exact final combination.

PR #1429, opened 2026-09-10
Add authored semantic family Transform intent

yongkyuns:
## Summary

Adds the authored Semantic Scene carrier for B3 unequal-family `Transform`, stacked on #1426.

- adds `SemanticAnimationIntent::FamilyTransformTo { source, target_state }` as one real authored animation declaration;
- both endpoints are real semantic families; no padding object/family nodes are created;
- adds the equivalent transaction vocabulary and `create_family_transform_animation(...)` constructor;
- preflights both endpoints as families and rejects same-endpoint declarations before allocation/publication;
- records source/target-state semantic references through the existing reference graph;
- treats family Transform as a leaf animation intent with source/target-state accessors, while correspondence remains compiler-derived below the Semantic Scene.

## Qualification

The exact production tree was qualified fail-fast before reconstruction:

- `cargo fmt --all -- --check` — passed;
- `cargo test -p noon-core --test family_transform_intent` — 3 passed, 0 failed;
- `cargo test -p noon-core semantic_animation --lib` — 14 passed, 0 failed.

The tested production blobs were then rebuilt as connector-authored commit `f9014bf6afee63493f807773650ebbedc60102f0` directly on #1426 exact head `334693e127390dc6fe8622d06e7a058c34abf798` so regular PR CI can qualify the linear stack. Net diff is exactly five core/test files; all temporary patch workflows/scripts are absent.

## Scope / remaining B3 work

This PR owns authored meaning only. It does **not** assign an execution object ID or expand semantic children. The next stack layer projects this intent through the existing composition scheduler into a family-Transform timing record with no `ObjectId`; activation will then combine that timing with #1421 correspondence and #1425/#1426 transient display publication/rendering.

Persistent unequal topology reconciliation remains exclusively explicit `become()` from #1416.

Refs #82.

### DIFF
--- a/crates/noon-core/src/semantic_store/semantic_animations.rs
+++ b/crates/noon-core/src/semantic_store/semantic_animations.rs
@@ -279,6 +279,13 @@ pub enum SemanticAnimationIntent {
         /// Ordinary Transform leaves source priority unchanged.
         complete_priority: bool,
     },
+    /// Transform one semantic family toward another through compiler-derived visual
+    /// correspondence. The two family identities remain authored scene state;
+    /// unequal padding occurrences never receive semantic identities.
+    FamilyTransformTo {
+        source: SemanticNodeId,
+        target_state: SemanticNodeId,
+    },
     /// Temporarily scale and recolor one object around a shared activation center.
     /// The compiler captures the effective source and lowers a restoring track.
     Indicate {
@@ -358,6 +365,7 @@ pub enum SemanticAnimationIntent {
 impl SemanticAnimationIntent {
     pub const fn target(&self) -> Option<SemanticNodeId> {
         match self {
+            Self::FamilyTransformTo { source, .. } => Some(*source),
             Self::ObjectPropertyTrack { target, .. }
             | Self::TransformTo { target, .. }
             | Self::Indicate { target, .. }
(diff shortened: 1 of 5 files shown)


yongkyuns:
Superseded by the atomic integration in #1430, merged as `39bee93fa5cf1887c03d5d22eedf5654811aee72`.

Standalone #1429 was intentionally not landed because adding `SemanticAnimationIntent::FamilyTransformTo` without the scheduler projection leaves `animation_schedule.rs` non-exhaustive and makes the product/workspace baseline uncompilable. #1430 was retargeted and collapsed onto healthy master so the authored intent plus required scheduler handling landed as one compilable 7-file unit.

The #1429 semantic-intent content is preserved byte-for-byte in that merge; this PR is closed only to avoid representing the deliberately invalid intermediate state as independently mergeable.

yongkyuns:
Superseded by the atomic #1430 integration, exactly as planned in #1430's PR body. #1430 landed the authored `FamilyTransformTo` intent from this PR together with the required scheduler projection in commit `965699046dceb2c7dae3e68cf6f831b61c59bc30`, then merged to master as `39bee93fa5cf1887c03d5d22eedf5654811aee72`. #1431 has since landed on top. Rebuilding this stale stacked PR would duplicate already-integrated semantics, so closing it without merge.

PR #1430, opened 2026-09-10
Project family Transform through shared scheduler

yongkyuns:
## Summary

Atomic integration of the B3 authored family-Transform intent (#1429) plus its required scheduler projection.

This PR is intentionally retargeted directly to `master` rather than landing #1429 alone. The authored `FamilyTransformTo` enum addition makes the pre-existing scheduler matches non-exhaustive until this scheduler layer is present; therefore the two logical slices must land as one compilable integration unit.

### Authored intent from #1429

- adds `SemanticAnimationIntent::FamilyTransformTo { source, target_state }` and equivalent transaction vocabulary;
- both endpoints are real semantic families; no padding semantic nodes are created;
- records source/target-state semantic references and validates family endpoints;
- adds focused semantic-intent regressions.

### Scheduler layer

- projects authored `FamilyTransformTo` through the existing recursive animation scheduler;
- adds published and prepared family-Transform timing records carrying source family, target-state family, root timing/time-map, finish map, and resolved options;
- deliberately carries **no stable execution `ObjectId`** for the family Transform record;
- uses the same composition scheduling/time-map recursion as object/scalar leaves;
- adds focused published/prepared/nested-sequence scheduling regressions.

## Architecture

The integration adds authored meaning plus execution-plan timing only. It does not allocate padding `SemanticNodeId`s, execution `ObjectId`s, stable renderer slots, or persistent topology. Identity-free correspondence/publication/render preparation remains owned by the already-landed #1421/#1425/#1426 layers. Persistent unequal topology reconciliation remains explicit `become()` from #1416.

## Why combined

Standalone #1429 is intentionally not mergeable as a product state: its historical workspace/product baseline fails with three non-exhaustive matches in `animation_schedule.rs` for `FamilyTransformTo`. #1430 is exactly the missing scheduler projection. Historical #1430 Product Gate failed while building the **#1429 baseline**, before building the candidate, for those same three non-exhaustive matches. Retargeting this PR to healthy `master` allows the actual 7-file atomic integration to be qualified normally.

## Current exact tree

- base: current `master` `968022149f2f7e3ecfdd5d21f987d72ac5bf7d64`;
- head: `341ee210d307cba1ba3a9f6e8f8957fb4853dbbe`;
- 2 clean logical commits, 7 files, +581/-5;
- all seven files were proven unchanged on master relative to their original B3 parents before exact blob replay.

After this lands, #1429 will be closed as superseded by this atomic integration and the stack advances to #1431.

Refs #82
Refs #1429.

PR #1431, opened 2026-09-10
Prepare activation-effective family Transform correspondence

yongkyuns:
## Summary

Adds the next bounded B3 activation-preparation layer, stacked on #1430.

- consumes the shared prepared family-Transform schedule from #1430;
- derives recursive unequal-family correspondence through #1421's compiler-owned correspondence path;
- resolves only real source leaves through the existing stable `SemanticExecutionIndex`;
- captures each distinct source execution row's coherent effective state exactly once at activation preparation;
- repeated source padding occurrences reuse that captured value as their initial visual state rather than receiving synthetic semantic/execution identity;
- carries deterministic occurrence ordinals plus shared timing/time maps/options for later derived-display materialization;
- publishes nothing and mutates neither semantic topology nor runtime state.

## Qualification

The exact production implementation before probe removal passed:

- `cargo test -p noon-compile --test family_transform_activation` — success (2→3 expansion/capture reuse and 3→2 contraction coverage);
- `cargo fmt --all -- --check` — success.

The diagnostic workflow's final bookkeeping condition reported failure despite both substantive steps succeeding; that temporary workflow has now been removed from the production diff.

## Scope / safety

This PR intentionally does **not** switch `ExecutionSession::FamilyTransformTo` away from its current strict equal-family path yet. Doing that before derived channel/materialization publication exists would create a no-op family Transform. The next layer will reuse the canonical Transform channel lowerer for ordinary source occurrences and release-only derived channels for repeated source occurrences, then materialize padding through #1425/#1426 identity-free display publication.

Persistent topology reconciliation remains exclusively explicit `become()` from #1416.

Refs #82.

### DIFF
--- a/crates/noon-compile/src/semantic_lowering/animation_payload.rs
+++ b/crates/noon-compile/src/semantic_lowering/animation_payload.rs
@@ -1,5 +1,6 @@
 mod affine;
 mod family_transform;
+mod family_transform_activation;
 mod prepared_composition;
 mod scheduled_captures;
 mod text_write;
@@ -7,6 +8,7 @@ mod transform_payload;
 
 pub use affine::*;
 pub use family_transform::*;
+pub use family_transform_activation::*;
 pub use prepared_composition::*;
 pub use text_write::*;
 pub use transform_payload::*;
--- a/crates/noon-compile/src/semantic_lowering/animation_payload/family_transform_activation.rs
+++ b/crates/noon-compile/src/semantic_lowering/animation_payload/family_transform_activation.rs
@@ -1,0 +1,211 @@
+use std::collections::HashMap;
+
+use noon_core::{
+    ObjectId, PreparedSemanticMutationTransaction, SemanticNodeId, SemanticTransactionNodeRef,
+};
+
+use super::super::{PreparedSemanticAnimationScheduleProjection, SemanticExecutionIndex};
+use super::affine::EffectiveAnimationProperties;
+use super::family_transform::{
+    derive_family_transform_correspondence, FamilyTransformCorrespondenceError,
+};
+
+/// One activation-time family Transform occurrence before execution publication.
+///
+/// The source/target fields are real authored leaf identities. `source_padding` and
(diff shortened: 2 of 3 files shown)


yongkyuns:
Merge-readiness reconciliation after clean restack onto merged #1430:

- current head `45b3ddfc538939d7abe7368e199ddd6eff888473` is exactly 1 commit / 3 compiler-test files / +376/-0 on master `39bee93fa5cf1887c03d5d22eedf5654811aee72`;
- the three production/test blobs are exact replays of the focused-qualified #1431 implementation; `cargo test -p noon-compile --test family_transform_activation` and `cargo fmt --all -- --check` passed historically on those blobs;
- historical normal matrix was green for Native Host, Architecture, Layer Dependency, Provider isolation, Playground Product Gate, and cross-browser;
- historical PR Fast's Rust format/lint, Rust unit tests, and web static/unit checks were all green; its sole red was primary WebGPU with the inherited stale `family_state` direct-WASM sample (`direct family state at 0ms missed expected members`), the already-diagnosed B1-era probe mismatch now corrected on current master;
- deterministic replay, typed WASM/Python error boundaries, and Manim-style authoring had all passed before that stale probe failed;
- this slice only prepares activation-effective correspondence/captured state; it publishes nothing, allocates no synthetic semantic/execution identity, and does not switch `ExecutionSession` onto the unequal-family path yet;
- fresh exact-head workflows exist but remain entirely queued due repository Actions saturation, with no fresh failure signal;
- the existing-file base (`animation_payload.rs`) was proven byte-identical between the original #1430 parent and the merged atomic #1430 tree before replay, so no newer compiler work was overwritten.

Given the exact focused-qualified 3-file tree, six historical broad green workflows plus all Rust/static PR Fast jobs, the conclusively unrelated stale browser-probe failure now fixed in the base lineage, and this preparation-only/non-publishing scope, this is a justified qualification transfer. Safe to promote/merge the current head; queued runs may serve as post-merge confirmation.

PR #1436, opened 2026-09-11
Lower unequal family Transform execution channels

yongkyuns:
## Summary

Adds the next bounded B3 layer for unequal-family `Transform`, stacked on #1431.

- lowers activation-effective family correspondence through the existing canonical `validate_affine_payload` / `lower_transform_channels` path; there is no second Transform interpolation implementation;
- keeps real source occurrences on the ordinary `PreparedSemanticAnimationTrack` vocabulary, including existing driver-conflict and authored completion behavior;
- represents repeated **source-padding** occurrences separately as identity-free derived channel bundles anchored only to their real source execution row for capture/painter provenance;
- adds the Manim-style faded-copy appearance lane (`0 -> 1`) only to the derived source-padding occurrence;
- equal-family correspondence produces only ordinary stable tracks;
- deliberately fails closed for **target-padding contraction** until its execution-only endpoint-hold policy is installed, rather than authoring padding opacity or letting ordinary `Release` pop the source visible again.

No semantic topology, `SemanticNodeId`, or stable execution `ObjectId` is allocated for padding copies. `ordered_family_leaf_pairs()` remains unchanged and strict.

## Qualification

Exact implementation before temporary-probe removal passed:

- `cargo test -p noon-compile --test family_transform_channels` — 3 passed, 0 failed;
- `cargo test -p noon-compile --test family_transform_activation` — 2 passed, 0 failed;
- `cargo fmt --all -- --check` — passed.

The temporary branch-only probe workflow has been removed from the production diff.

## Stack / remaining B3 work

Base: #1431 / `codex/b3-family-transform-activation`.

Next is the narrow contraction completion policy: retain only the padded-target **appearance** endpoint as effective execution state while canonical geometry/transform/style channels complete normally. After that, the derived-channel evaluator will materialize #1425 identity-free display rows and feed #1426 renderer-owned transient buffers. Only then will `ExecutionSession::FamilyTransformTo` switch away from its current strict equal-family expansion.

Refs #82.

### DIFF
--- a/crates/noon-compile/src/semantic_lowering/animation_payload.rs
+++ b/crates/noon-compile/src/semantic_lowering/animation_payload.rs
@@ -1,6 +1,7 @@
 mod affine;
 mod family_transform;
 mod family_transform_activation;
+mod family_transform_channels;
 mod prepared_composition;
 mod scheduled_captures;
 mod text_write;
@@ -9,6 +10,7 @@ mod transform_payload;
 pub use affine::*;
 pub use family_transform::*;
 pub use family_transform_activation::*;
+pub use family_transform_channels::*;
 pub use prepared_composition::*;
 pub use text_write::*;
 pub use transform_payload::*;
--- a/crates/noon-compile/src/semantic_lowering/animation_payload/family_transform_channels.rs
+++ b/crates/noon-compile/src/semantic_lowering/animation_payload/family_transform_channels.rs
@@ -1,0 +1,404 @@
+use std::collections::{hash_map::Entry, HashMap};
+
+use noon_core::{
+    ObjectId, PreparedSemanticMutationTransaction, Property, SemanticNodeId,
+    SemanticObjectProperty, SemanticTransactionNodeRef, TimelineError, TrackDefinition, TrackId,
+    TrackValues,
+};
+
+use super::affine::{
+    completion_at_endpoint, driver_key, lower_transform_channels, transform_driver_conflict,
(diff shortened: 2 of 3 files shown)


yongkyuns:
Clean-restack / qualification checkpoint

- Recovered the true child delta from direct parent `9b64da90a7d3b8d912b7a532b88489a48234449f`: exactly 3 files, +628/-0.
- Current master `a81f6fb3b9498834db17ca97a7b3be53ddaaebcb` had the existing export blob byte-identical to that parent and neither new #1436 file present, so the tested production blobs were replayed without semantic conflict.
- Clean head is `7bd12a867cd4e0ebd458dc53b1c1a04b9690a672`, one commit directly on master; compare confirms exact merge base and exact 3-file diff.
- Original focused qualification passed: `family_transform_channels` 3/3, `family_transform_activation` 2/2, and rustfmt.
- Original broad CI: Layer, Architecture, Native Host, Product Gate, cross-browser, Provider, Rust tests, fmt/lint, web preflight, deterministic replay, typed error boundary, and Manim authoring all passed. The only PR Fast failure was the pre-existing `browser-smoke.mjs` direct `family_state` 0ms probe (`right` rendered black), the same stale probe independently seen on earlier B3 slices and fixed by intervening master work; it fails before/independently of this compiler-only family-Transform channel path.
- Fresh workflows on the clean head are currently runner-queued with no negative signal.

Treating this as qualification transfer for the byte-identical production change on the corrected master integration head.

PR #1438, opened 2026-09-11
Retain contraction padding appearance as effective state

yongkyuns:
## Summary

Adds the bounded unequal-family Transform contraction policy on top of #1436.

- real source leaves continue to use canonical Transform geometry/affine/style channels;
- when correspondence maps a real source to a padded target copy, only an execution-domain `Appearance 1 -> 0` channel represents that padding fade;
- that appearance channel is marked `retain_effective`; canonical geometry/transform/style channels keep ordinary authored completion behavior;
- a later Transform that maps a previously hidden real source to a real target naturally emits `Appearance 0 -> 1` with ordinary completion;
- repeated source padding remains identity-free derived display data.

The retain marker is execution-release metadata, not semantic animation meaning: no authored opacity, membership, node identity, or topology is created for alignment padding.

## Qualification

Exact production content passed before workflow cleanup:

- `cargo test -p noon-compile --test family_transform_channels` — 4 passed;
- `cargo test -p noon-compile --test family_transform_activation` — 2 passed;
- `cargo clippy -p noon-compile --tests -- -D warnings` — passed;
- `cargo fmt --all -- --check` — passed after the formatter-only correction.

The branch has been reconstructed as one production commit `be70c498224952335ffa25dd2b402473d44b8b43` directly on #1436; temporary diagnostic workflows are absent.

## Remaining integration

The child session layer supplies the generic `retain_effective` release policy, and the derived-evaluator layer evaluates source-padding copies outside stable `FrameState.objects`. `ExecutionSession::FamilyTransformTo` remains on the strict equal-family expansion until those pieces are wired and qualified together.

Refs #82.

PR #1439, opened 2026-09-11
Retain execution-only presentation endpoints at completion

yongkyuns:
## Summary

Adds the live-session release policy required by unequal-family Transform contraction, stacked on #1438.

- adds `retain_effective` to session completion entries as execution-release metadata;
- ordinary tracks continue to reconcile back to authored base state exactly as before;
- a retained presentation endpoint skips only the final `ReconcileTrack`, leaving the execution track authoritative;
- existing RevealLifecycle and ZIndex completion behavior is preserved;
- all existing completion-entry construction defaults `retain_effective` to false, so this changes no current animation behavior until family Transform wiring opts in.

This deliberately does **not** add another `SemanticAnimationCompletion` variant: retaining an effective padding endpoint is execution release policy, not authored animation meaning.

## Qualification

Exact production content passed:

- focused `retained_presentation_endpoint_skips_only_execution_reconciliation` test;
- `cargo check -p noon`;
- `cargo clippy -p noon --lib -- -D warnings`;
- full `cargo fmt --all` was applied before the focused gate.

The branch is reconstructed as one production commit `489606ac8ddd68e2e3f5974166487698ab62de2c` directly on #1438; the temporary patch workflow is absent.

## Remaining integration

The family Transform activation layer must propagate #1438's per-track `retain_effective` marker into the corresponding completion entry. Derived source-padding copies remain separate identity-free renderer publication data.

Refs #82.

### DIFF
--- a/crates/noon/src/execution_segment.rs
+++ b/crates/noon/src/execution_segment.rs
@@ -45,6 +45,10 @@ pub(crate) struct SegmentCompletionEntry {
     pub property: Property,
     pub track: TrackId,
     pub end_time: f64,
+    /// Keep this execution endpoint authoritative instead of reconciling the track
+    /// back to authored base state at segment completion. Used only for derived/
+    /// presentation state with no authored field to receive the endpoint.
+    pub retain_effective: bool,
 }
 
 #[derive(Clone, Copy, Debug)]
--- a/crates/noon/src/execution_session.rs
+++ b/crates/noon/src/execution_session.rs
@@ -1219,6 +1219,7 @@ impl ExecutionSession {
                 property: track.property,
                 track: track_id,
                 end_time,
+                retain_effective: false,
             });
             definitions.push(definition);
             next_track_id = raw_id.checked_add(1);
@@ -3243,6 +3244,7 @@ impl ExecutionSession {
                             property,
                             track,
                             end_time,
+                            retain_effective: false,
                         }
                     },
                 )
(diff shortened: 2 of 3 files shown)


yongkyuns:
Clean-restack / qualification checkpoint

- #1439 is now one commit directly on merged #1438 master `4582e47d8d61c2b2201aa13b2fdef030e4f8ba0a`.
- Clean head `e369c0300eade4cce25996b190b33b1c5ad10cdd` is exactly the previously qualified 3-file child delta: `execution_segment.rs`, `execution_session.rs`, and `execution_session/completion.rs` (+53/-6).
- All three input blobs on current master were byte-identical to #1439's original parent, so replay introduced no semantic merge/overwrite risk.
- Original focused qualification passed: retained-presentation endpoint completion test, `cargo check -p noon`, clippy `-D warnings`, and rustfmt.
- Original broad CI was green for Architecture, Native Host, Layer, Product Gate, Provider isolation, cross-browser, Rust tests, fmt/lint, web checks, deterministic replay, typed error boundary and Manim authoring. The sole PR Fast failure was the known stale direct `family_state` browser probe, unrelated to this completion-policy slice and fixed by intervening master work.
- `retain_effective` defaults to false for all existing completion entries, so current animation behavior remains unchanged until the later family-Transform activation layer opts in.
- Fresh workflows on `e369c030...` are currently runner-queued with no negative signal; master/head are unchanged and GitHub reports mergeable=true.

Using qualification transfer for the byte-identical production delta on the corrected integration head.

PR #1449, opened 2026-09-11
Render derived family Transform rows in direct hosts

yongkyuns:
## Summary

Completes the direct retained-host consumption layer for B3 derived unequal-family `Transform`, reconstructed as one clean commit on the landed #1447 renderer layer.

- prepares viewport-filtered identity-free derived display rows from the exact `RendererPublication` used for stable retained preparation;
- uploads derived Circle/Rectangle/Line instances into #1447's separate transient GPU buffers only when present;
- adds profiled mixed geometry encoding without changing the existing host-owned timestamp-query lifecycle;
- opts both native host and WASM direct canvas into mixed retained geometry submission;
- keeps ordinary retained rendering unchanged when no derived overlay exists;
- fails closed when derived rows would require retained text/glyph painter interleaving;
- accounts derived upload bytes in existing web telemetry.

## Restack / reconciliation

Current clean head: `034ebd99d1886442417330c5100ce1d1e3793fc8`, one commit directly on landed #1447 head `502cf0a69db2aa656ca61973f37287f4e021a653`, with exactly the intended six-file surface (+185/-28). The obsolete #1451 ancestry is absent.

Five files were replayed byte-for-byte after confirming their historical base blobs exactly matched the clean #1447 tree. `crates/noon-web/src/execution_canvas.rs` had legitimate newer master-side changes, so only #1449's localized host hunks were reconciled while preserving those additions.

No semantic IDs, stable execution IDs, stable renderer slots, or second painter authority are introduced.

## Exact-head qualification

All 8 fresh master-base workflows are green:
- Layer Dependency Ratchet — run `34758397063` / #2188;
- Architecture Ratchets — run `34758397046` / #2229;
- Native Host Smoke — run `34758397035` / #1130;
- PR Fast Gate — run `34758397017` / #2391;
- Provider feature isolation — run `34758397044` / #720;
- Playground authoring recovery — run `34758397077` / #2215;
- Playground cross-browser matrix — run `34758397090` / #2488, green on Chromium desktop, Firefox desktop, and WebKit mobile including main-thread fallback;
- Playground Product Gate — run `34758397045` / #1103, including baseline/candidate runtime, visual, latency, and FPS comparison.

The next clean layers (#1450 and #1453) have already been reconstructed as one-commit descendants and are qualifying in parallel.

Refs #82.


yongkyuns:
Architecture follow-up: the identity-free transient GPU/storage direction is correct, but the renderer-facing abstraction should not remain shaped around `family Transform` as more consumers arrive.

#1466 now owns the generalization target: stable retained publication + feature-neutral transient presentation occurrences, one authoritative painter projection, generic host consumption, and localized add/update/remove. Before path/text/glyph breadth is added, renderer errors/APIs should become authoring-feature neutral rather than adding more `FamilyTransform`-specific branches or sibling derived pipelines.

This does not invalidate this bounded analytic slice; it establishes the required consolidation point before expansion.

PR #1450, opened 2026-09-11
Redraw after derived family Transform endpoint expires

yongkyuns:
## Summary

Fixes the final transient-overlay retirement seam for B3 unequal-family `Transform`, reconstructed as one clean commit after #1449.

- preserves the derived copy for one coherent exact-endpoint renderer publication;
- when that publication consumes the expiring plan, queues a renderer-only full invalidation for the immediately following publication;
- the follow-up publication contains no derived rows, so a retained physical surface removes the synthetic copy instead of leaving its last pixels indefinitely;
- does not mutate Semantic Scene state, authored family membership, runtime spatial state, or stable execution slots;
- leaves ordinary renderer publication unchanged when no derived overlay is expiring.

## Clean stack

Current head: `0fc01e38118cd7d6f35bd2b89b43e0a38488b8d9`, one commit / three files directly on clean #1449 head `034ebd99d1886442417330c5100ce1d1e3793fc8`. Its tree is `82cf7265e8850a95cb3f4e3aa35f841b41f413c3`.

#1449 has now landed on `master` as `ef337fc1532afa45c4d3a0501bcb2f9a836ffd11`, whose tree is exactly the same `b7a853ebb4cfee69f0e690ddfe2d7257b67f2792` as #1450's parent. Therefore this retarget changes only the base ref, not the effective base tree or tested diff.

Fresh qualification on this clean head is already in progress/mostly green; it must finish before landing.

## Regression

The unequal-family expansion regression verifies that completion leaves a pending renderer frame and that the next publication is a full invalidation with no derived display objects while authored membership remains unchanged.

Refs #82.


yongkyuns:
Architecture disposition after review: keep this as a **bounded flat-analytic proof seam only**. The follow-up renderer `invalidate_all()` preserves correctness, but it must not become the locality contract for broader unequal-family Transform.

#1466 now owns replacement with generic transient-presentation retirement that removes/updates only the affected ephemeral occurrences/painter range while preserving unrelated stable residency. If this PR lands before #1466, please record the full invalidation explicitly as temporary and do not use this implementation to justify nested/path/text/glyph breadth.

The persistent semantic/topology behavior in this PR remains correct; the follow-up is specifically about renderer-local retirement/scalability.

yongkyuns:
Superseded for landing by #1535. GitHub cannot reopen this PR because the head branch was reconstructed after closure. #1535 uses an alias branch pointing at the exact cleaned head `0fc01e38118cd7d6f35bd2b89b43e0a38488b8d9`; no new implementation was introduced.

PR #1451, opened 2026-09-11
Keep nested unequal-family rejection coverage

yongkyuns:
## Summary

Updates one stale canonical-authoring regression after flat unequal-family Transform became intentionally supported in the B3 stack.

- the prior fixture expected a flat 2→1 family Transform to reject;
- flat 2→1 contraction is now a supported B3 case, so that expectation no longer describes the product contract;
- the rejection fixture now uses a nested unequal target family, which remains explicitly unsupported/fail-closed by the current flat-analytic session activation boundary;
- no production behavior changes.

Final diff is exactly one test file: `crates/noon-web/src/canonical_authoring_scene.rs`.

This preserves meaningful crosswrite/rejection coverage without misclassifying newly supported flat unequal-family Transform.

Refs #82.


yongkyuns:
Superseded by #1446. The sole intended content of this PR—the nested unsupported-target rejection fixture—was folded into the repaired #1446 root so the current-master qualification exercised the correct contract. #1446 has now merged at `075c36cb42e77c4cb692e09bb125677b622b59c0` with all normal exact-head workflows green. No additional production or test change remains unique to #1451.

PR #1453, opened 2026-09-11
Verify unequal-family direct seek equivalence

yongkyuns:
## Summary

Adds the explicit B3 deterministic-seek acceptance regression for the bounded flat unequal-family Transform path, stacked on #1450.

The test builds two identical 2→3 family Transform sessions:

- one advances through an intermediate presented frame before reaching `t=0.5`;
- the other performs an absolute deterministic `seek(0.5)` immediately after activation.

It then verifies both sessions produce exactly the same stable `FrameState` and the same identity-free derived-display occurrence, including anchor/order and partial midpoint appearance.

This exercises the real live `ExecutionSession` path rather than compiler-only correspondence helpers. No production behavior changes.

Final diff: `crates/noon/src/execution_session/family_transform_tests.rs` only.

Refs #82.


yongkyuns:
Superseded for landing by #1536. #1536 is the clean reconstruction directly on merged #1534/master and contains exactly the intended one-file +59-line acceptance regression with no stale production ancestry.

PR #1455, opened 2026-09-11
Add source-equivalent unequal-family raster qualification

yongkyuns:
## Summary

Adds the final source-equivalent B3 visual qualification layer for the bounded flat unequal-family `Transform` path.

- adds ordinary ManimCE v0.21 source fixtures for 2→3 `VGroup` expansion and 3→2 `VGroup` contraction;
- samples the existing canonical frame fractions and uses the standard global raster tolerances;
- registers the fixture in the existing Manim raster manifest;
- no runtime/renderer/semantic production behavior changes.

## Clean reconstruction

Current head `ae0fb0df8a886f0f2a04a05dc2673decd1ce50da` is exactly **1 commit / 2 parity files** directly on merged #1536 master `fdb60080127e70b0ff48baea0779f5771c19f105`.

The seek-equivalence test from former parent #1453 is already landed in master, so this reconstruction keeps only the true raster delta:
- `parity/manim-v0.21/core-examples/unequal_family_transform.py`
- `parity/manim-v0.21/manifest.json`

Applying those exact final blobs to the landed #1536 tree reproduces the previously fully-qualified #1455 content tree; no stale stack production files are carried forward.

Refs #82
Refs #1466


yongkyuns:
Qualification scope clarification: this PR may establish exact Manim parity for the bounded flat analytic unequal-family case, but it should not be read as architectural clearance for broader unequal-family Transform.

#1466 is now the prerequisite before nested topology, retained path copies, text/glyph interleaving, or another transient-copy consumer. Keep this fixture narrow and keep broader capability classification `partial` until the generic/localized transient-presentation lane is in place.

PR #1460, opened 2026-09-11
Retain Arrow dependent scale semantics

yongkyuns:
## Summary

Continues #77 after merged #1456 from master `bcd0b65abf8e4613a69f29fe1be826ab78995b4e`.

- retains Arrow shaft/tip dependency roles in authoritative `SemanticObjectState`, including the constructor-time uncapped stroke width and stroke/length cap ratio required by ManimCE v0.21 scaling;
- keeps Arrow leaves ordinary analytic Line / retained VectorPath content; compiler/runtime/renderer remain Arrow-agnostic;
- implements shared Rust `ManimArrow::scale(factor, scale_tips)` with one semantic transaction;
- `scale_tips=false` repositions/reorients retained tips without rebuilding their immutable geometry resources and recaps the shaft from retained authored policy;
- `scale_tips=true` scales all component leaves together, then applies the same stroke recapping rule;
- exposes the operation through the WASM Arrow handle and thin Python `Arrow.scale` facade;
- retains explicit fail-closed behavior for scale pivot kwargs until that breadth is shared Rust semantics.

## Architecture

The previous abandoned scale branch kept `initial_stroke_width` and the cap ratio on `ManimArrow` wrapper state. This PR does not reuse that ownership. The policy now lives on the semantic shaft role, so wrapper cloning does not create a second source of truth. Tip roles are also semantic state and are published in the original Arrow constructor transaction.

No Arrow primitive or policy is added to runtime/renderer state. Mutation stays local to the Arrow dependency closure and immutable tip resources are reused.

## Qualification

Focused Rust tests cover short-arrow shrink/grow stroke recapping, retained tip resource identity, `scale_tips=true`, negative scale direction reversal, DoubleArrow two-tip retention, zero-length no-op behavior, and atomic invalid-input rejection with unrelated state unchanged. Python facade tests verify scale delegates as one Rust operation and does not reconstruct tips/policy in Python.

The canonical source-equivalent `ArrowGeometry` ManimCE v0.21 fixture now exercises preserved-tip scaling, explicit `scale_tips=True`, DoubleArrow scaling, and a short Arrow grown enough to recover its original stroke-width cap. The matching user-facing Noon Python example and paired native Rust example exercise the same retained dependency behavior. Existing global raster tolerances remain unchanged.

Exact-head GitHub CI, including the canonical Manim raster differential, is the merge gate.

Refs #77.


yongkyuns:
Wrap-up / handoff for #1460 (`56830c3bfa59743a165c5393f1155549cd133bdf`):

The retained Arrow scale slice is implemented. Arrow shaft/tip dependency roles and constructor-time uncapped stroke policy live in authoritative semantic state; `ManimArrow::scale(factor, scale_tips)` updates the dependency closure in one semantic transaction, preserves immutable tip resources for `scale_tips=False`, supports `scale_tips=True`, DoubleArrow, negative factors, zero-length no-op, and stroke recapping. WASM and the thin Python `Arrow.scale` facade are wired. Python now explicitly rejects post-bootstrap/live Arrow.scale rather than mutating the authored store behind `LiveSession`; live dependent publication remains separate work.

Qualification on exact head: PR Fast Gate, Architecture Ratchets, Layer Dependency Ratchet, Native Host Smoke, Provider isolation, Product Gate, authoring recovery, cross-browser, Noon agent foundation, and Agent discovery are green. The Manim raster differential is still running, but public Python-vs-Manim semantics, preview/capture checks, canonical Manim rendering, semantic frame capture, and the per-fixture raster ratchet have already passed. The remaining running raster step is shared sparse/dense playback, followed by artifact upload.

Important integration note: master has advanced from the PR base `bcd0b65a…` to `cf32a4ee…` (11 commits), including merged #1454. There is real file overlap in `crates/noon-web/src/authoring_arrow.rs` and `web/python/_manim_arrow.py`, and GitHub currently reports #1460 non-mergeable. Before landing: let the raster run finish, re-read current master, reconcile/rebase those overlapping Arrow changes without losing either slice, then rerun exact-head qualification, mark ready, and merge.

Suggested #77 continuation after #1460 lands:
1. Add shared Rust boundary-point resolution for Mobject endpoints and use it for `Arrow(start=mobject, end=mobject)` / related line-derived constructors. Do not reconstruct bounds in Python; reuse shared boundary/layout semantics.
2. Add live retained-family Arrow mutation publication so `Arrow.scale()` can work after bootstrap through the owning `LiveSession` instead of direct store mutation; keep transferred/stale ownership fail-closed and preserve atomic publication.
3. Then close remaining Arrow mutation breadth (pivot options if justified) and continue the broader #77 vocabulary: DashedLine/DashedVMobject, Elbow/TangentLine/simple line-derived helpers, with tips/dashes retained as dependencies and no renderer-only Arrow primitive.

#77 remains open and its acceptance still includes the broader line/arrow/dash vocabulary; this PR is the retained dependent-scale slice, not the B2 closeout.

yongkyuns:
Architecture disposition: this PR's move away from wrapper-owned `initial_stroke_width`/cap state into authoritative semantic state is the correct direction and should be preserved.

One guard for follow-up work: treat the new Arrow shaft/tip roles as a bounded Arrow implementation, not a pattern for growing `SemanticObjectRole` into a Manim-class registry. If a second unrelated composite needs comparable component/dependency metadata, extract the smallest generic semantic dependency/component contract at that point instead of adding more class-specific role variants.

No change is requested to the core ownership here: Python stays thin, mutation stays one semantic transaction, immutable tip resources are reused, and runtime/renderer remain Arrow-agnostic.

Issue #1366, opened 2026-09-09
Restore WebGPU device-loss recovery and validation diagnostics


yongkyuns:
## Current failures

The Phase A closeout reran the existing renderer recovery checks against the retained real renderer-smoke package from #1364. These are unresolved product/host correctness failures, not successful checks or inferred GPU-retirement work.

- Direct WebGPU device-loss recovery: run34399817109 and browser-only34405592889. The script expects a distinct replacement device and equivalent output after loss; only one device is observed. Requiring replacement-device observation inside the polling loop did not establish recovery. Current direct execution_canvas.rs exposes WebGL context recovery but lacks the equivalent automatic WebGPU replacement path. Reuse retained artifacts before rebuilding.
- Direct recoverable validation diagnostics: the existing browser harness drains recoverable diagnostics to console.warn. Tests that expect renderAt to throw must respect the intended recoverable contract; the attempted corrected diagnostic observation in34405592889 still failed and is not accepted evidence.
- Worker validation diagnostics: safe BigInt serialization exposed the real failure in34405592889: worker control boundaries did not surface the injected diagnostic while retained rendering remained ready/settled. Preserve the diagnostic rather than treating a failed assertion or serialization error as a pass.
- The renderer workflow also needs explicit existing renderer-smoke feature selection before fixture imports can execute; production default exports intentionally omit these fixtures. The exploratory workflow change was removed from the narrow resize #1364 rather than silently treating all recovery checks as green.

## Ownership and scope

Browser platform integration owns loss/recovery policy and platform lifecycle; noon-render-wgpu remains reusable rendering machinery, and the existing runtime remains authoritative. Preserve scene identity, authored/effective time, coherent publication and retained resource semantics during recovery. Do not add a second runtime/scene, retry by rerunning source, or put a serialized intermediate between same-context Rust engine layers.

This is separate from #953 Phase A consolidation and from Phase C GPU submission retirement. The finite Phase A gate proves one authority per engine layer, typed direct targets, shared frontend semantics and supported atomicity/locality/settle-wake. Automatic WebGPU replacement and expanded diagnostic delivery are explicit product correctness follow-ups; their current failures are not waived as successful validation.

## Acceptance

- Actual device replacement after loss, including two successive losses, preserves semantic state and renders the expected output.
- Direct and worker recoverable validation errors reach the established diagnostic boundary without becoming silently lost or falsely terminal.
[...]

PR #1504, opened 2026-09-12
Release retained renderer borrow before GPU diagnostic await

yongkyuns:
Fixes a regression introduced by #1477 and exposed while integrating #1410.

`RetainedExecutionCanvasRenderer::flushGpuDiagnostics` was exported as `async fn (&mut self)`. wasm-bindgen therefore kept an exclusive Rust borrow of the renderer alive while JavaScript awaited WebGPU `popErrorScope()`. Engine-port messages are independent of the main control dispatch and may apply deltas/render during that await, re-entering the same renderer and trapping with `recursive use of an object detected which would lead to unsafe aliasing in rust`.

This change pops/re-arms the validation scope synchronously, clones only the diagnostic mailbox/generation/backend, and returns an owned Promise for the popped scope future. The `&mut self` borrow ends before JavaScript awaits, while #1477's recoverable WebGPU validation-diagnostic contract remains intact.

The Shared text authoring workflow now also includes `crates/noon-web/src/retained_execution_canvas.rs` in its push/PR path filters. #1477 did not trigger this workflow, which allowed the regression to escape its most direct browser qualification.

Isolation/validation:
- diagnostic #1503 reproduced the recursive-aliasing crash on exact master `ba7ae17733b2e14155bba57b873cfeb5e3bec946` without #1410 code;
- exact runtime-fix code at `eed256e4405a94f871f79037848d9d4b10a7e73a` passed the previously failing Shared text authoring smoke;
- PR Fast Gate passed on that same code, including Rust tests, rustfmt/clippy, WebGPU/WebGL and browser authoring checks;
- final `cff3201cb8c32af7c4c80f82f700de50405519e1` changes only the workflow path filter relative to the green runtime code and keeps the Rust file byte-identical.

Refs #1366
Unblocks #1410.


yongkyuns:
Isolation result: temporary PR #1503 ran the same Shared text authoring workflow on exact master `ba7ae17733b2e14155bba57b873cfeb5e3bec946` with no #1410 code and reproduced the wasm-bindgen `recursive use of an object detected which would lead to unsafe aliasing in rust` failure during retained Typst startup. This confirms the blocker is already present on master and is independent of #1410. #1503 is now closed unmerged.

The fix here changes the exported retained-renderer diagnostic flush from `async fn (&mut self)` to a synchronous method returning an owned Promise: pop/re-arm the wgpu validation scope under the short mutable borrow, then await the popped scope future with only cloned diagnostic metadata. Current exact head `eed256e4405a94f871f79037848d9d4b10a7e73a` has rustfmt/clippy and web preflight green; Shared text/browser qualification is running.

PR #1525, opened 2026-09-13
Restore retained morph geometry transport locality

yongkyuns:
## Summary

Restore the retained Dynamic Load morph locality that was lost when production worker playback was unified onto `SemanticExecutionPlayer`.

- Keep the current unified semantic/runtime ownership model; do not resurrect the deleted retained shadow player.
- Publish a transient morph path inline only when a stable transport row first observes that immutable render geometry, or when the geometry identity changes.
- On later incremental frames for the same stable row and same immutable geometry, omit the path payload and reuse the render-worker mirror's already-installed geometry.
- Retain the previously published `Arc<GeometryRef>` in the encoder and compare with `Arc::ptr_eq`, so allocator address reuse cannot alias a different path.
- Clear the reuse identity on endpoint/removal and force a fresh publication on re-entry.
- Bump the retained execution transport version because omitted incremental render geometry now has an explicit v4 meaning for an existing row with prior transient geometry.

## Root cause

PR-era commit `aab908da` (`Unify family worker imports with shared semantic execution`) removed the specialized retained family player and, with it, the production `compiled_render_geometries` / resource-indexed encoder path. `RetainedExecutionDeltaEncoder::with_render_geometries` became test-only, while production `SemanticExecutionPlayer` uses the ordinary encoder with no compiled render-geometry table.

As a result, `transport_object()` fell back to cloning `frame.render_geometries[index]` into every dirty morph row. Dynamic Load therefore repeatedly serialized hundreds of full vector morph paths across the authoring/render worker boundary instead of reusing retained geometry.

Recent A/Bs ruled out the two leading same-day suspects: disabling the persistent WebGPU validation scope changed hosted throughput by only ~4%, and restoring the pre-#1410 morph selection was effectively flat.

## Regression coverage

`continuation_morph_geometry_is_inlined_once_then_reused_by_slot` verifies the broken invariant directly:

1. first morph publication carries the vector path;
2. the next frame for the same immutable path carries no path commands and round-trips to the identical frame in the mirror;
3. a changed path is published again;
4. clearing transient geometry invalidates reuse;
5. re-entering after a clear republishes the path.

Existing snapshot validation continues to reject an omitted render geometry, and new rows cannot reuse a geometry that was never installed.

## Performance evidence

A hosted WebGPU A/B of the initial locality fix reduced Dynamic Load morph phase round-trip cost roughly in half:

- morph A: ~73.9 ms -> ~33.5 ms (-54.7%)
- morph B: ~70.8 ms -> ~33.2 ms (-53.1%)
- overall hosted effective FPS: ~11.36 -> ~12.56 (+10.6%)

Hosted WebGPU is not the physical throughput acceptance environment. Physical WebGPU qualification remains the repository's `dynamic-stress` guard with a 55 FPS floor, against the historical ~59.7-59.8 FPS baseline from #1110/#1111.

Refs #1110, #1111, #1518.

### DIFF
--- a/crates/noon-web/src/retained_execution_transport.rs
+++ b/crates/noon-web/src/retained_execution_transport.rs
@@ -17,7 +17,7 @@ use crate::TransportSlotId;
 /// Object content and family-plan semantic bindings are explicit so geometry and
 /// text share the source identity/order stream across a genuine worker boundary.
 pub const RETAINED_EXECUTION_TRANSPORT_CHANNEL: &str = "noon.execution.retained";
-pub const RETAINED_EXECUTION_TRANSPORT_VERSION: u32 = 3;
+pub const RETAINED_EXECUTION_TRANSPORT_VERSION: u32 = 4;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct TransportTextResourceHandle {
@@ -247,6 +247,10 @@ pub struct RetainedExecutionDeltaEncoder {
     // Retain the Arcs so pointer keys cannot be recycled during this encoder's lifetime.
     render_geometries: Option<Arc<[Arc<GeometryRef>]>>,
     render_geometry_indices: Option<HashMap<usize, u32>>,
+    // Retain the last transient render geometry published for each stable row.
+    // Keeping the Arc alive prevents allocator address reuse from aliasing a different path.
+    // Incrementals may omit an unchanged inline path and let the mirror reuse it.
+    published_render_geometries: Vec<Option<Arc<GeometryRef>>>,
 }
 
 impl RetainedExecutionDeltaEncoder {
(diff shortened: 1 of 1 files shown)


yongkyuns (commented):
Final validation after rebasing onto the then-current `master`:

- PR Fast Gate: green (Rust unit tests, fmt/clippy, web/static, WebGPU, WebGL).
- Architecture and layer dependency ratchets: green.
- Fresh Playground Product Gate baseline-vs-candidate comparison: green.
  - effective FPS: 52.1221 -> 52.9186 (+1.53%)
  - visual diff: 0 pixels
  - cold Run -> applied: 5701.9 ms -> 5606.0 ms
  - warm Run -> applied: 1169.6 ms -> 1180.4 ms
  - edit -> applied: 2654.2 ms -> 2256.8 ms
- Dedicated Dynamic Load hosted A/B from the root-cause investigation reduced the heavy morph phases by roughly 53-55%.
- The deterministic regression test `continuation_morph_geometry_is_inlined_once_then_reused_by_slot` now locks the transport invariant: unchanged transient morph geometry is published once and then reused by stable slot; changed/cleared/re-entered geometry is republished correctly.

The absolute physical WebGPU Dynamic Load qualification (55 FPS floor) is separately queued on the self-hosted runner; the runner has not started that job yet. The merged code does not depend on that validation branch.

PR #1530, opened 2026-09-13
Restore cold morph resource residency

yongkyuns:
## Case

Implements the first performance case established by #1529: `morph-a-activation` / cold morph activation.

A live Transform activation now publishes immutable renderer geometry before the first active morph frame. The renderer installs and preloads the retained suffix transactionally, and the active frame references the installed resource instead of carrying inline vector geometry.

## Implementation

- keeps the unified `SemanticExecutionPlayer` and existing Runtime ownership;
- publishes compiled/effective immutable render geometry as append-only retained-resource additions at the engine-worker -> render-worker boundary;
- stages incremental resource additions transactionally before committing transport state;
- appends retained render-geometry mirror tables without rebuilding the resident prefix;
- preloads newly admitted path meshes into the renderer before first presentation;
- preserves inline fallback for geometry that cannot be retained safely;
- seeds live render-resource indices after any resident prefix, preventing prefix/resource aliasing;
- preserves native build surfaces while keeping renderer-only borrowed views scoped to WASM;
- does not resurrect the deleted specialized retained player.

## Regression coverage

- cold morph activation requires retained resource publication before presentation;
- the first active morph frame must reference retained geometry rather than inline vector geometry;
- incremental render-resource additions preserve transactional append semantics;
- a non-empty resident prefix is covered explicitly: prefix length 1 -> first live resource index 1;
- incremental path-residency append and failure-preservation behavior are covered in renderer tests.

## Current qualification

Restacked as one commit on current master:

- base: `8bbc81ce544a319ca768bc3c7281023ddfcad1e5`
- head: `c51fe309bb26d160c75e6d4e1e5ca91aa6933523`
- diff: 12 files, +871 / -124
- mergeable: yes
- Architecture Ratchets: green
- Layer Dependency Ratchet: green
- PR Fast Gate: green, including WebGPU, WebGL, and canonical callback renderer observations
- Native Host Smoke: green
- Shared text authoring: green
- Playground authoring recovery: green
- Authoring WASM Footprint: substantive build/measurement/upload steps green; post-job cleanup completing

Still intentionally draft while the exact-head cross-browser matrix and Product latency/FPS/visual comparison complete, and while the physical WebGPU master-vs-candidate `morph-a-activation` comparison waits for the self-hosted Mac runner.

Performance budgets remain measurement-only per #1529. The renderer still intentionally invalidates prepared draw descriptors when the immutable resident prefix grows because the old disposable path suffix is dropped; avoiding that first-frame rebuild would require a real suffix-relocation design, not a safe one-line optimization. That larger change is deferred unless the activation measurement shows it is warranted.


yongkyuns:
Blocking review finding on current head `77f6c6f5...`: `RetainedFamilyExecutionDeltaEncoder::new_with_resources()` still initializes `next_render_geometry_resource` to 0 even when the initial retained resource bundle already contains render geometries. A later live append can therefore publish global resource index 0 and alias an unrelated installed prefix resource. The current cold-morph regression starts from an empty render table, so it cannot detect this.

Please initialize the producer counter from the initial bundle's installed render-geometry count (checked `u32` conversion), while leaving `new(session)` at zero. Add a regression with one initial render geometry followed by one incremental live render geometry: the new row must reference global resource 1, the addition preparation remains local resource 0 on the wire, and the receiver must rebase that preparation exactly once to 1. This is a merge blocker independent of the generic CI matrix.

yongkyuns:
Performance acceptance concern in addition to the resource-index correctness blocker: the new `commit_path_mesh_append()` sets the geometry preparer `initialized = false`, and `append_preload_path_meshes()` clears `prepared_generation_ready`. The next referencing frame can therefore take the full retained-geometry rebuild path. #1529 defines `morph-a-activation` around a 600-object morph, so this can turn resource admission into scene-scale CPU preparation even though tessellation/GPU residency was moved earlier.

This may be acceptable only if measurement proves the 600-object activation is at the best attainable state (all 600 rows are genuinely dirty), but it should not be assumed. Please add/retain instrumentation or a focused regression that distinguishes resource preload work from the first post-append frame's `full_rebuilds` / instance repack work, and use the `morph-a-activation` / `morph-b-activation` case results as the merge gate. If the post-append frame is doing unrelated stable-scene rebuild work, preserve the existing prepared state or separate immutable residency from disposable frame packing rather than accepting the full rebuild.

Issue #1571, opened 2026-09-14
B3: deterministic TransformMatchingShapes


yongkyuns:
## Scope

Implement ManimCE v0.21 `TransformMatchingShapes` on Noon's consolidated deterministic animation path.

### Required semantics

- Match family members by normalized point geometry, equivalent to ManimCE v0.21: center the geometry, normalize height, round coordinates to 3 decimals, then compare the resulting shape key.
- Shape matching is translation/scale invariant and rotation-sensitive.
- Preserve deterministic/stable grouping when duplicate keys occur.
- Matching and start-state capture happen at animation activation, so a later child in `Succession` observes the current effective geometry rather than request-construction state.
- Reuse Noon's existing family-transform activation/channel/completion machinery; do not add a frontend frame loop or renderer-specific path.
- Keep unmatched source/target members explicit. Default mismatch behavior must follow Manim's directional fade-out/fade-in semantics; `transform_mismatches`, `fade_transform_mismatches`, and `key_map` must either be implemented faithfully or fail closed and remain documented as partial.
- Public Python surface should follow ManimCE v0.21 call shape and defaults for the supported subset.
- Update compatibility evidence only for behavior actually qualified.

### Qualification

- Native tests for translation/scale invariance, rotation sensitivity, duplicate-key stability, and explicit leftovers.
- Activation-time regression: `TransformMatchingShapes` after an earlier `Succession` child must match the effective geometry seen at its own begin time.
- Python adapter/API tests for supported options and explicit rejection of unsupported options.
- Paired executable evidence where the promoted behavior is exposed through both native/direct-WASM and Python paths.
- Exact-head CI before merge.

Parent: #82

yongkyuns:
Activation-staging note after #1564 landed:

- `TransformMatchingShapes` should stay on the existing family-transform semantic/compiler/runtime path; no frontend frame loop or parallel lifecycle.
- #1564's `effective_semantic_object` / effective-boolean path is useful precedent for ownership, but intentionally rejects render-geometry overrides, so it is not the mechanism for Succession matching after a prior geometry Transform.
- The exact seam is `ScheduledCaptures`: today it can carry completed affine/style endpoints forward, but deliberately bails out when a previous `Property::Morph` or `Property::Transform` geometry driver exists.
[...]

PR #1606, opened 2026-09-15
B3: execute matching-shape family transforms through session

yongkyuns:
Stacked continuation of #1571 on #1599.

This is the source-only execution-session seam for deterministic `TransformMatchingShapes`.

## What changed
- select structural vs matching family lowering from the authored `SemanticFamilyTransformMode`;
- feed already-lowered prior object tracks into activation-time shape matching so a later Succession child observes completed morph/affine endpoints;
- reuse the existing family stable-track path and the single generic transient-presentation plan;
- attach source -> original authored target replacement to the existing exact segment-completion token;
- preserve Manim cleanup ordering at exact end: same-z `[before, source, after] -> [before, after, target]` in both semantic membership and runtime painter order;
- publish only the final semantic ContentMorph endpoint per object when a sequence contains multiple completed morph tracks, while still reconciling every execution track;
- fail closed for required callbacks, scalar-leaf composition, invalid replacement topology, and other unsupported states;
- keep ordinary structural family Transform behavior unchanged.

## Architecture
- no frontend matching or frame loop;
- no second lifecycle/state machine;
- no matching-specific renderer/runtime lane;
- shape keys/correspondence remain compiler-owned activation-time data;
- transient padding/leftovers reuse the existing `DerivedDisplayAnimationPlan` facility;
- completion uses the existing atomic semantic -> execution -> runtime publication boundary.

## Validation
The exact source patch passed before flattening:
- `cargo check -p noon`;
- strict `cargo clippy -p noon --all-targets --all-features -- -D warnings`;
- matching-by-shape rather than position + source-to-authored-target replacement;
- Succession activation after a completed prior content morph;
- matching target transient tests;
- matching completion topology tests;
- same-z exact-end semantic/runtime painter ordering and next-publication stability.

The branch is flattened to one source-only commit over its stack base; temporary qualification workflows/scripts are no longer part of the PR.

Exact-head CI is green on the architecture ratchets, layer dependency ratchet, native host smoke, and PR fast gate.


yongkyuns (commented):
Architecture review of the current integration plan: the product seam is aligned with #1571 once the one-shot qualifier materializes and removes itself. Matching mode reuses the existing semantic family-transform selector, activation-time effective-state capture, stable family tracks, the single generic transient presentation plan, and the existing exact segment-completion replacement token; no second renderer or lifecycle authority is introduced.

Two qualification blockers from the earlier attempts are now correctly addressed in the self-cleaning run: (1) the public LiveSession hook constructs `MatchingFamilyTransformTo`, eliminating the internal dead-code/clippy failure, and (2) sequential content-morph completions publish only the final authored content endpoint while still reconciling every execution track, matching the existing property/priority completion contract.

Do not treat the present staging-only head as reviewable/mergeable product code. The acceptable head is the resulting source-only commit with the three `.github` qualifier files removed, focused matching + Succession regressions passing, and normal exact-head CI restarted on that product-only tree.

yongkyuns (commented):
Follow-up on the source-only qualification: the product patch itself passed `cargo check`, strict all-target/all-feature clippy, matching-session shape/replacement, Succession activation-state, target-transient, and completion tests. The prior publish failure was staging-only.

I found and fixed a second staging hazard before it could land: the shortened publisher invoked `.github/b3_matching_session_qualifier.py` but did not execute the generated `/tmp/b3-qualify.sh`, which would have committed only the LiveSession/completion subset and omitted the execution-session matching path/tests. `ff17a578` now makes the preparer execute the complete already-qualified patch and revalidate it before cleanup.

I also added the remaining #1571 exact-end ordering proof to materialization: with same-z `[before, source, after]`, completion must publish semantic and runtime painter order `[before, after, original_target]`, retire the source execution row, and preserve that order on the next deterministic publication.

Keep this PR draft until the self-cleaning source-only commit has landed and the three staging files are gone; then I can review the exact product diff.

yongkyuns (commented):
Public-adapter follow-up after this source-only seam lands: keep `TransformMatchingShapes` as a thin selector only. The canonical Rust/Web authoring path already has `OrdinaryCompositionChild::FamilyTransformTo -> noon::AnimationCompositionRequest::FamilyTransformTo`, and Python ordinary family Transform already delegates to `appendFamilyTransformTo`. The next slice should add the matching sibling at those two boundaries and let the #1606 engine own keys/correspondence/lifecycle.

Do not reuse `web/python/test_manim_shared_shape_matchers.py` as animation evidence; it covers static SurroundingRectangle/BackgroundRectangle shared bounds helpers. Add dedicated animation adapter tests instead. Unsupported `key_map`, `transform_mismatches`, and `fade_transform_mismatches` should remain explicit/fail-closed unless their existing native correspondence semantics are wired faithfully in that same slice.

yongkyuns:
[P1] Active matching painter order is now a reproduced raster blocker, not only a proposed extra test.

Integrated #1616 source `9890ad9d88f87e10f6c65287e6b857a256600f9a`, focused run https://github.com/yongkyuns/noon/actions/runs/35028891302, artifact `10421265816`, has successful real WebGPU/WebGL captures. At t=0, pixel (696,270) inside the red-kite/right-circle overlap is Manim RGB (233,109,99), Noon (234,184,180). That establishes red-over-white versus white-over-red independently of easing and antialiasing. Both Noon backends are pixel-identical.

The source roots were authored [before, source, after]. During matching, Manim moves the temporary animation group after surviving roots; Noon's published painter rows remain [before id0, source triangle id2, source kite id3, after id1]. The existing `matching_family_completion_appends_target_in_same_z_painter_order` only proves the final target append, not activation/interior placement. Completion/post-completion frames pass; begin and interior samples expose the gap.

Detailed timing-scope and painter-order findings are recorded on the existing adapter stack: https://github.com/yongkyuns/noon/pull/1614#pullrequestreview-5216368077 . Keep this implementation on the shared activation/publication path; do not add frontend cleanup, change authored z-index, reorder the fixture to avoid overlap, or widen thresholds. Preserve existing identity, locality, rejected-activation atomicity and source-restoration behavior. Current additional fixture assertions at `5b32772` do not constitute a fix for these product discrepancies.

yongkyuns:
Taking the independent active same-z matching painter-order P1 from comment 5688848938. I will work against the current integrated #1616 tree `e4240f66752ba6023d5397077d3017b9c9231f0c` to preserve the completion/source-restoration, Wait-root assertion, and raster-host fixes. The correction will remain in the shared activation/publication machinery, using the existing transient presentation/order representation; no Python reorder, z-index workaround, new renderer path, or threshold change. I will add focused active/interior/cleanup and rejected-activation coverage and qualify it before integrating product-only changes. This does not take or duplicate the separately owned #1614 timing-scope work.

PR #1607, opened 2026-09-15
Keep complex filled SVG transforms continuous

yongkyuns:
## Summary

Fix the Ghostscript tiger → rocket → tiger demo so it performs an actual filled-shape `Transform` instead of fading to outlines, popping the rocket in, and restoring late tiger fragments.

#1583 is now merged into `master`; this PR has been integrated and revalidated on top of that merge.

## What changed

- preserve authored paint throughout the real tiger ↔ rocket transforms
- add canonical `PreparedPathInterpolation` for ordered cubic path correspondence
- support complex filled path transforms with a renderer-local sampled/tessellated fallback while retaining the fixed-topology GPU fast path
- reuse one sampled mesh per stable row / derived occurrence without placing sampled meshes in immutable shared or mega-mesh caches
- keep sampled-path updates local to the affected geometry and painter-order chunk
- interpolate unequal-family padding paint exactly once, matching Manim's faded-copy semantics rather than multiplying paint and `Appearance` fades
- match Manim SVG fitting/centering semantics more closely for cubic control-point bounds
- add paired Rust/Python SVG morph examples and regressions for continuity, direct seek parity, family restoration, padding paint, and locality
- document the renderer-local specialization in `docs/architecture.md`

## Review follow-up

Review found a mixed-scene renderer bug where a sampled path could update a mega-path attribute buffer it did not own. The final head now:

- requires actual mega-segment ownership before `update_mega_path_instance()` touches that buffer
- includes permanent regressions for opacity-only sampled updates beside a static vector path and for growing sampled tessellation beside a static mega path
- preserves the newer `master` transient-renderer refactor while layering the sampled-path realization into it
- preserves `path_arc` from current `master`

## Validation

Prior qualification on the same Tiger behavior passed:

- real Manim 0.21 differential
- actual Ghostscript tiger → rocket → tiger browser gallery playback
- mobile-sized retry
- direct Rust/WASM browser smoke
- Python unequal-family round-trip browser smoke

The exact master integration commit `e4ca94c16238b3aa6f62b938a82a821606b136b6` additionally passed rustfmt and focused compiler/renderer regressions before publication. Normal PR CI is running on that exact head.

`bash scripts/check.sh test` previously reported three failures that were independently reproduced on #1583's exact parent (`family_state=101 resource_atomicity=101 lifecycle=101`), so they are pre-existing baseline regressions rather than Tiger changes:

- `unequal_family_become_reuses_receiver_identity_and_never_imports_target_ids`
- `invalid_publication_context_rejects_before_resource_import`
- `lifecycle_churn_presence_transitions_keep_morphed_paths_resident_on_the_exact_sample_grid`


yongkyuns (commented):
## Review verdict: do not merge this revision

Reviewed `06f1590eae6c0cb921dd9e443c140a6769df23da`, including renderer cache ownership, local draw-order updates, padding paint interpolation, the SVG placement change, and the #1583 dependency. The shared Rust approach is appropriate, and all 12 existing PR workflows passed, but an additional mixed-scene regression reproduces a new renderer panic.

### P1 — sampled paths can write into a mega-path buffer they do not own

`rebuild_mega_path_draws()` now excludes sampled fills, leaving their `mega_path_segments[batch]` as `None`, while `mega_path_detached[batch]` is initialized to `false`. `update_mega_path_instance()` only checks the detached flag and whether any mega indices exist; it never checks whether this batch actually owns a mega segment.

A scene containing one sampled concave-to-curved fill and one unrelated static vector rectangle is enough. When tessellation grows, local path replacement allocates a larger vertex range, then the instance update indexes the fixed mega attribute buffer using that range. The unchanged PR engine panics at `mega_mesh.rs:281`:

```
range end index 92 out of range for slice of length 12
```

A separate opacity-only update unnecessarily dirties mega attributes (`[0..8]`) for the sampled row. The existing locality test's 1,200 unrelated circles do not exercise this mixed vector/mega-path case.

### Reproduction and verified correction

Isolated probe branch: `review/1607-mixed-sampled-mega`, commit `4ed3f1593e0939f4f32d705ea606c886ef693bcc`, directly parented by the PR head. Its first test step leaves engine code unchanged.

Run: https://github.com/yongkyuns/noon/actions/runs/34982270915/job/104425304809

- Original PR engine: existing 3 tests pass; both new mixed-scene tests fail, including the panic above.
- Add an actual-segment-ownership guard in `update_mega_path_instance()` before accessing the attribute buffer: all 5 tests pass.

The guard is equivalent to returning early when `!self.mega_path_segments.get(path_batch_index).is_some_and(Option::is_some)`, retaining the existing detached check. It fixes the ownership invariant rather than adding a tiger-specific condition.

The workflow intentionally continues after the original failure to test the correction, so its overall green status is not a pass for the original implementation. Artifact `review-1607-mixed-paths` contains `original.log`, `guard.log`, and `proposed-guard.patch`:
https://github.com/yongkyuns/noon/actions/runs/34982270915/artifacts/10401828548

The correction has only been tested in the isolated job workspace; it is not part of this PR yet. Add the guard and retain both new regressions in the permanent renderer tests, then revalidate before merging. This is distinct from the three already-reproduced parent-baseline failures.

crates/noon-render-wgpu/src/mega_mesh.rs
```diff
@@ -26,8 +26,12 @@ impl FramePreparer {
         let eligible = self
             .path_batches
             .iter()
-            .map(|batch| {
-                batch.instance_range.end == batch.instance_range.start + 1
+            .enumerate()
+            .map(|(index, batch)| {
+                self.path_mesh_cache[self.path_batch_cache_indices[index]]
+                    .sampled
+                    .is_none()
```
yongkyuns, line 33: [P1] Excluding sampled paths from mega eligibility also requires an ownership check in `update_mega_path_instance()`. These batches retain `mega_path_segments[batch] = None` and `mega_path_detached[batch] = false`. With any unrelated static vector path present, an incremental growing sampled fill therefore reaches `mega_path_vertex_instances[range]` and panics (`range end index 92 out of range for slice of length 12` at line 281). An opacity-only update also dirties unowned mega attributes. Both failures are reproduced on unchanged PR engine code in run 34982270915; both pass when the updater returns early for batches with no mega segment. Please include that guard and mixed static/sampled regression coverage before merging.

yongkyuns: Verified this review finding is addressed in final head `e4ca94c16238b3aa6f62b938a82a821606b136b6` before merging. `update_mega_path_instance()` now returns early when `mega_path_segments.get(path_batch_index).is_some_and(Option::is_some)` is false, retaining the existing detached check. Permanent coverage is present in `crates/noon-render-wgpu/tests/complex_filled_morph_mega.rs`: `sampled_fill_paint_does_not_dirty_unowned_mega_attributes` and `growing_sampled_fill_does_not_index_the_fixed_mega_attribute_buffer`. Rechecked GitHub Actions for that exact head: all 12 PR workflows are completed successfully. Resolving this addressed P1 thread; the earlier review verdict applies to the old revision, not this corrected final head.

yongkyuns:
Post-merge follow-up: the `Test coverage` push run on merge commit `036c23a2db8d96e14122a85c9f34af25e18861e6` failed in the new `complex_filled_family_round_trip_restores_padding_without_repainting` regression (actual Appearance 1.0 vs expected 0.0 after contraction completion). Actual job: https://github.com/yongkyuns/noon/actions/runs/35018191917/job/104547196724.

This is a separate integration mismatch, not one of the three old baseline failures. #1609 landed after #1607's tested base and persists completed padding invisibility in authored opacity while releasing Appearance. The existing neighboring round-trip test already follows that contract; the newly added complex-filled test still expected the former execution overlay.

Opened cleanup PR #1617 from the exact failing master commit. It changes only the stale regression, preserves both production fixes, and strengthens the single-fade midpoint / hidden hold / fully restored paint checks. Head `d9c4e1a25382e8cc6dc1ba432af0fa2adae5e342`; candidate CI is not yet qualified.

yongkyuns:
Qualification follow-up from #1615: a merge candidate against master `036c23a2db8d96e14122a85c9f34af25e18861e6` fails the complex filled family round-trip unit test added here.

Evidence: PR Fast Gate run https://github.com/yongkyuns/noon/actions/runs/35025238732, Rust unit job `104570738730`, checked-out merge commit `abb66a1bfc05f46e73887e61924ed6a0b26b0c28` (`0b8238a` into `036c23a`). `cargo test --workspace --all-features --lib` reports 374 `noon` tests passed and one failed:

`family_transform_renderer_publication_tests::complex_filled_family_round_trip_restores_padding_without_repainting`

At `crates/noon/src/family_transform_renderer_publication_tests.rs:259`, after contraction completion/drain, the test expects `appearance == 0.0` but receives `1.0`.

The adjacent `renderer_publication_drain_preserves_restored_family_appearance` test already requires neutral `Appearance == 1.0` with hidden authored `style.opacity == 0.0`. The complex-path test still checks the old Appearance-only contraction and recovery contract. This looks like a test-contract mismatch with the merged persistent completion changes, but it needs targeted verification rather than simply deleting/changing the expected value. The replacement assertions should preserve hidden effective paint after contraction, continuous return opacity, endpoint restoration, and post-drain capture.

#1615 does not change this file. I have not run master alone locally; its failed merge-candidate gate remains a blocker, not an ignored baseline failure.

PR #1609, opened 2026-09-15
Persist unequal-family Transform endpoint semantics

### LINKED ISSUE #1601: [Cleanup B3] Persist unequal-family Transform endpoint semantics
Parent: #82. Related locality work: #1466/#1499.

## Why this cleanup exists

Current master `7bd54eecc971d820e3b2ad5e0aa0afe529a0ad87` proves unequal-family `Transform` with identity-free padding occurrences during interpolation, but the completion test currently codifies a transient visual endpoint that is removed on the following publication while the source family's authored membership remains unchanged.

For a source family with 2 leaves transforming to a target family with 3 leaves, `unequal_family_transform_publishes_one_identity_free_expansion_copy` currently expects:

1. one derived presentation during interpolation;
2. the derived presentation still visible at the exact endpoint;
3. a following presentation-only publication that removes it;
4. the source family to retain its original 2-member authored topology.

That makes the visible exact endpoint and persistent semantic state disagree. It also means a subsequent `wait`, query, copy, family traversal, or animation can observe a different object structure than the transform endpoint that was just rendered.

Manim's ordinary `Transform` aligns source/target families before interpolation and the transformed source retains the aligned result. Noon should preserve that observable semantic behavior without turning temporary renderer occurrences into fake semantic identity.

## Goal

Define and implement one persistent unequal-family `Transform` completion contract that preserves source identity/alias semantics while reconciling the final aligned topology into the Semantic Scene before transient presentation resources retire.

## Required design constraints

- Do not keep a renderer-only transient occurrence indefinitely as the persistent endpoint.
- Do not replace the source root with the target root; ordinary `Transform` mutates the source rather than behaving like `ReplacementTransform`.
- Existing source semantic identities and aliases must remain valid where Manim/source semantics retain them.
- Any additional persistent receiver-owned leaves required by alignment must be legitimate Semantic Scene nodes, not `ObjectId`/renderer identities promoted upward.
- Target authored identities remain distinct; ordinary `Transform` must not steal target nodes merely to satisfy alignment.
- Contraction/expansion, nested topology, aliases, and fail-closed unsupported cases must be explicit.
- Completion must use the normal atomic semantic -> execution -> runtime publication path.
- The transient-presentation renderer remains a temporary interpolation mechanism only; #1466 owns its locality.

## Acceptance

- 2 -> 3 and 3 -> 2 family Transform cases match Manim observable source topology/content after completion.
- Exact endpoint and the next publication are visually/semantically identical except for removal of implementation-only transient resources.
- `wait()` after completion does not visually or semantically revert the aligned endpoint.
- family member count/order, layout queries, copy/save state, and a second animation observe the persistent endpoint.
- target family remains independently valid and unchanged where Manim keeps it independent.
- aliases/shared leaves preserve defined identity semantics without duplicate mutation.
- direct seek to completed deterministic state agrees with forward completion.
- no fake semantic identity is allocated for purely temporary interpolation copies.
- focused native + direct-WASM/Python parity evidence covers lifecycle, not only midpoint raster output.

## Regression source

The current behavior is explicitly asserted in `crates/noon/src/execution_session/family_transform_tests.rs::unequal_family_transform_publishes_one_identity_free_expansion_copy`; change that test from transient-retirement semantics to the intended persistent endpoint contract rather than merely adding another endpoint test.

## Non-goals

- transient renderer locality (#1466);
- `TransformMatchingShapes` cleanup ordering (#1571);
- broad animation IR redesign (#1485);
- changing `ReplacementTransform` semantics.

yongkyuns:
Fixes #1601.

## What changed
- persist unequal-family ordinary `Transform` completion into the source-owned semantic family instead of leaving padded interpolation state transient-only
- reconcile the persisted semantic endpoint through execution/runtime publication atomically
- preserve receiver identity/priority metadata and target independence while keeping interpolation padding identity-free
- cover 2→3 expansion, 3→2 contraction, target independence, completed seeks, copy/family traversal, and a second animation consuming the persisted endpoint
- keep authored aliases/shared leaves explicitly fail-closed through the existing correspondence contract rather than duplicating mutation
- fail closed on nested family completion at the shared prepared-activation boundary before publication; recursive correspondence remains compiler-owned for a future persistent nested-topology slice
- upgrade the shared `family_transform_indicate` example to a 2→3 lifecycle proof

## Validation
- focused unequal-family lifecycle suite: 14/14 passed before the shared-example upgrade
- full Noon validation and clippy passed for the core change; two unrelated assertions were confirmed already red on clean master during that run
- nested/flat activation-boundary regressions were added after review to prevent valid recursive correspondence from reaching the currently flat-only persistent completion helper
- corrected shared lifecycle proof and the review follow-up are running through normal PR CI; temporary scratch workflows are not part of this PR


yongkyuns (commented):
I found one completion-contract gap that should be fixed before treating #1601 as complete.

`stage_unequal_family_transform_completion` is flat-only: it reads direct family members and then `expect`s each aligned member through `semantic_object_state_checked`. But the existing compiler correspondence explicitly supports nested topology (`nested_alignment_recurses_after_parent_padding`) and recursively lowers family↔family / family↔object alignment. A nested unequal Transform can therefore activate successfully and then panic during completion instead of persisting the aligned endpoint or failing closed. #1601 explicitly calls out nested topology and fail-closed unsupported cases.

Please either (preferably) carry enough compiler-derived recursive completion provenance into the pending completion plan to materialize receiver-owned aligned topology without recomputing it from renderer state, or reject unsupported nested completion before publication with a typed error. Add a lifecycle regression that reaches `complete_segment`, not only correspondence lowering.

Also, the PR body says aliases/shared leaves are covered, but the unchanged correspondence currently has `unequal_authored_aliases_fail_closed` and returns `AliasedSourceLeaf` / `AliasedTargetLeaf`. If alias support remains intentionally fail-closed in this slice, please state that accurately and add/retain endpoint-level evidence; otherwise #1601's alias acceptance is still open.

yongkyuns:
Review follow-up applied in `e2d65e2`: the prepared family-Transform activation boundary now rejects nested family endpoints before any semantic/runtime publication, so recursive correspondence cannot reach the currently flat-only persistent completion helper and panic. Added focused nested fail-closed + flat-admission tests. I also corrected the PR body: authored aliases/shared leaves remain explicitly fail-closed in this slice rather than being claimed as supported.

PR #1614, opened 2026-09-15
B3: expose TransformMatchingShapes through shared session

yongkyuns:
## Scope

Public `TransformMatchingShapes` adapter for #1571, stacked on #1606. The adapter selects the canonical Rust `MatchingFamilyTransformTo` request; matching, activation-time capture, interpolation, publication, completion and cleanup remain in the shared Rust execution/session path.

Current flattened head: `fa88235c9d946ae8f458391f27ddaa45b3a6ea5b` — **1 commit / 12 changed files**.

## Product behavior owned here

- Preserve original Python wrapper identity across successful matching completion without replaying `Scene.add` or introducing Python-side scene authority.
- Matching source channels use `Release` completion; re-add restores the activation-effective source state, including prior-composition endpoints, rather than stale pre-composition geometry.
- Constructor timing/easing remains on matching children while `Scene.play(...)` timing/easing applies to an enclosing composition. Unsupported play-level matching lag/path options fail closed.
- Matching activation moves the existing source family behind surviving roots through the semantic publication transaction, preserving stable execution identity and z-index rather than applying renderer-only draw-order overrides.
- Target-leftover `LayerEnd` anchors are preflighted against the proposed reordered source layer, not the old frame order.
- Shared/aliased source descendants are rejected before semantic reorder or publication on this bounded path.
- Existing unsupported `key_map`, `transform_mismatches` and `fade_transform_mismatches` modes remain fail closed; this is not a full Manim compatibility claim.

## Focused regressions

The flattened commit carries the implementation and the regressions that belong with it:

- completion binding identity/getter/second-animation and source re-add,
- native direct-seek vs dense-forward matching phases and repeated-time stability,
- activation-time same-z painter order and exact-end target replacement,
- failed activation atomicity,
- leftover-anchor placement after source reorder,
- shared-descendant rejection before mutation/publication,
- nested child/outer timing scopes through the canonical authoring scene,
- Python transport tests separating constructor options from `Scene.play` options.

## Stack split

The timing/order/session repair that previously lived downstream has been moved here. #1616 is now restacked at `8e83df5909a87b45fe97eb2ea2bb44aa0ea47ad8` and contains only the direct-WASM/public-browser/raster qualification layer.

The combined #1616 full tree remains `af64f2133f4059951e605d6a39ef8377f5985e8a`, identical to the already fully qualified pre-split tree. The rewrite changed commit ownership only, not the final integrated source.

## Qualification state

The exact implementation bytes in this PR were exercised in the previously green combined tree, including Rust tests/lint, browser authoring, WebGPU/WebGL rendering, product gate, cross-browser matrix and Manim raster qualification. Fresh exact-head CI has been triggered for this standalone flattened head and is the current acceptance gate.

Keep this PR **draft and unmerged** until exact-head CI and the remaining stack/base closeout are reviewed. No automated Codex/security review is requested.

Refs #1571; qualification layer: #1616.


web/python/_manim_scene.py
```diff
@@ -1952,7 +1954,12 @@ def append_leaf(builder: object, animation: object, child_kwargs: dict[str, obje
                 raise NotImplementedError(
                     "canonical family Transform requires linear or smooth easing"
                 )
-            builder.appendFamilyTransformTo(
+            append_family_transform = (
+                builder.appendMatchingFamilyTransformTo
+                if type(leaf) is _animate.TransformMatchingShapes
```
yongkyuns: [P1] Reconcile wrapper associations after matching-family completion

This branch appends the matching request and returns without reserving/associating target leaf wrappers or registering the replacement family. Unlike ordinary in-place Transform, Rust's matching completion removes `source` and admits the original `target`. `_play_canonical_composition` only commits `reservations`/`family_registrations`, and its completion callback only handles `removals`; the matching branch populates none of them. The common continuation event handler returns immediately for `kind == "complete"`, so it does not repair this either.

Consequently the target leaves still have `_scene is None` after successful Rust completion, while source wrappers still point at the Scene. The next `self.play(Indicate(target))` is rejected by `_canonical_indicate_animation` with `ValueError("Indicate target must belong to this Scene")`. `_group_members` merely returns existing wrappers by Rust member key; it does not derive binding state. #1616's proposed Python example uses precisely this sequence.

Please refresh the affected derived wrapper associations/registry from the successful shared Rust replacement and its authoritative identities, at the proper completion boundary. Do not work around it by issuing an extra Python `Scene.add(target)` or adding a Python-owned replacement lifecycle. Add a real Python play -> matching completion -> membership/getter -> Indicate(target) regression, plus failure-before-completion atomicity coverage. The current constructor/string-presence tests cannot detect this.

yongkyuns (commented):
Correctness review of exact head c32666fbb7e39d10fae2e067f3ad397abe870b7f, continuing #1571 without a competing implementation branch. Found a blocking Python lifecycle-association gap below. The shared Rust matching/runtime/renderer reuse is the right boundary, but the public target must remain usable after the successful replacement. This was reproduced in an isolated CPython boundary test of the existing completion and Indicate classifier functions with a mocked successful Rust membership swap; it is not a claim of a full browser/Rust integration run.

web/python/_manim_scene.py
yongkyuns: Follow-up tracing at unchanged head `c32666fbb7e39d10fae2e067f3ad397abe870b7f`: this needs the canonical Rust binding registry as well as Python attachment reconciliation. In `canonical_authoring_scene.rs`, the `bindings()` traversal explicitly treats `MatchingFamilyTransformTo` as a no-op (around 1487), so the target is absent from both `bindings` and `identities`. `mobject_observation()` then rejects an otherwise valid published target at `!self.identities.contains_key(&handle.node_id())` with `mobject is not bound to this canonical Scene`. Merely setting Python `_scene` would move the failure from Indicate admission into effective getters rather than repair it.

The existing reusable pieces are `SceneMembershipBatch` binding reservation/validation, `_membership_leaf_bindings`, `_commit_typed_binding`, `_register_membership_wrappers`, and `_sync_membership_wrapper_attachments`. Any completion-time association should validate the entire affected binding batch against the already-published Rust membership, preserve exact semantic identities, then update derived registries only. It must not call `editMembership`/`Scene.add` to replay the replacement, relower the scene, or mark the target present before completion.

Please qualify target identity + effective getter + target-family Indicate, old-source detach/re-add, and rejected/pending completion atomicity together. The separate fixture-coordinate blocker in #1616 is now fixed by `8b1ef54886219804a497f6f3923419efeb8fd6c6`; this lifecycle thread remains unresolved.

yongkyuns: Implementing this repair against unchanged head c32666fbb7e39d10fae2e067f3ad397abe870b7f. Candidate uses the existing membership binding validator plus a bindings-only association boundary that requires completed/coherent execution and already-published membership; Python invokes it only after the successful completion receipt, then refreshes target identities and source attachments. It does not replay Scene.add/replace or modify the shared matching lifecycle. Qualification is isolated on tooling/qualify-1614-completion-bindings before applying product-only files to this existing PR branch. Local Python discovery: 256 tests, 25 skipped, passing; the new completion suite fails on the unpatched baseline and passes with the candidate. Native/Rust-WASM/browser qualification is not yet complete.

yongkyuns: Implemented the completed-binding reconciliation in `2090fe1d6a2f85a7d73e5863b0391090059b9395` and integrated that commit into the #1616 stack (the parity branch has since advanced independently, and its new commits are preserved).

Rust now exposes a binding-reservation-only association against already-published membership. It reuses membership binding preflight, rejects pending/transferred/incoherent players and unpublished handles, and commits both identity registries only after validating the complete batch. Python calls it through the existing successful completion callback, then attaches the original target wrappers and detaches only source leaves that are actually absent. No `Scene.add`, `editMembership` replay, geometry matching, or new scheduler is involved.

Qualification run 35025394410: all 8 new Python production-function boundary tests and all 3 existing adapter tests passed. These Python tests mock the Rust boundary and are not a public-browser integration claim. The native crate compiled; native pending/transferred and whole-batch rejection tests passed. The real native target-getter and subsequent-Indicate assertions also passed, but its final source re-add assertion exposed an additional native cleanup defect: matched tracks still commit ordinary Transform endpoints into the removed source.

I am qualifying a compiler-payload correction that marks matching source tracks as release-only and clears padded appearance retention, without changing interpolation. It also adds source restoration checks after a prior Succession Morph and for matched/padded/unmatched members. Run 35026376752 is the isolated qualification for that correction. Keep this PR draft; the source re-add regression must pass rather than be weakened or dismissed.

yongkyuns:
Exact-head CI update for `0619d5726f4eeee86d30f4ea115157bb711203f1`: the connector now reports all 11 returned PR-triggered workflows completed successfully, including PR Fast Gate 35027042278, Playground Product Gate 35027043167, cross-browser matrix 35027042175, authoring recovery, Native Host Smoke, Architecture Ratchets and Layer Dependency Ratchet. This supersedes the missing-origin/master setup limitation of the earlier isolated fast-gate attempt; it does not turn that historical failed run green.

The integrated #1616 native/Python/sampler run 35027195267 also completed successfully. The remaining targeted acceptance evidence is real public matching lifecycle and paired raster qualification. I found the raster driver was using a non-isolated Python HTTP server, despite synchronous construct requiring COOP/COEP. Its shared-server repair is being qualified separately in 35029510663, without changing matching semantics or tolerances. Keep this PR draft pending that evidence and completion review.

web/python/_manim_scene.py
yongkyuns: Reviewed the integrated fix at 0619d5726f4eeee86d30f4ea115157bb711203f1. The bindings-only boundary now validates the complete batch against published membership before changing either Rust registry, and the existing successful-completion callback updates only affected Python associations. The additional Release-only source-track policy also preserves source re-add instead of persisting matching target endpoints/padding into the removed source.

I verified the actual output of run 35026376752, not just the PR summary: 8 Python boundary tests and 3 existing adapter tests pass; all 3 real native completed-binding tests and all 4 matching lifecycle tests pass. The native getter/second-Indicate/source-re-add regression now completes successfully. The run identifies clean product tree ad824e37159a701e68054f4436a8c0ef059c9556; its later fast-gate failure is the explicitly reported missing origin/master setup error, not a test pass.

The implementation findings are addressed at the inspected native/boundary layers, but this thread's requested real public-Python/browser qualification is not yet established. To make that gate detect the original bug, I added identity, root-order, public getter, and restored-source assertions to #1616's paired raster fixture in 5b32772413e72e56b814a78941b7adc0b86ea3d0, after play and both existing waits, without altering animation/timing/thresholds. Keep this thread and the PR draft until those assertions and the public second-Indicate lifecycle are actually exercised on the integrated head; a native/direct-WASM-only pass is insufficient.

yongkyuns (commented):
Blocking paired-raster findings from the integrated stack, separate from the completed-binding/source-restoration fixes.

Evidence: focused run https://github.com/yongkyuns/noon/actions/runs/35028891302 checked out source `9890ad9d88f87e10f6c65287e6b857a256600f9a` (which includes this PR's `0619d57`). Artifact `b3-matching-materialized-frames`, ID `10421265816`, SHA256 `c244cca96b6e8a75cf35d95743f8092eb95bed32980cc68b9f074bb9edae9b2a`. Both backends successfully captured all six frames; sparse/dense replay has identical effective state and pixels. Strict Manim comparison fails at 0%, 25%, and 75%. This is actual browser evidence, not the earlier mocked-boundary test or the Polygon admission failure.

### [P1] Preserve constructor-child versus Scene.play group option scopes

The fixture calls `self.play(TransformMatchingShapes(source, target), run_time=2, rate_func=linear)`. In pinned Manim v0.21.0, `TransformMatchingAbstractBase.__init__` passes constructor kwargs to its Transform/Fade children, then calls `super().__init__(*anims)` with the AnimationGroup's linear outer default. `Scene.compile_animations` applies play kwargs only to the top-level animation. Therefore play-level linear does not replace the inner Transform's default smooth rate.

Noon's `_play_canonical_composition` family-transform branch instead resolves `_canonical_transform_options(leaf, child_kwargs, allow_family_lag=True)` and sends one resolved rate to `appendMatchingFamilyTransformTo`. The distinction is erased. At 25% and 75% the captured geometry and paint are linear in Noon but smooth in Manim; midpoint and endpoints agree and therefore earlier midpoint-only coverage missed it. Differing ratios are 0.0583217593 and 0.0576388889, versus the unchanged limit 0.003; mean channel errors are 5.1495408951 and 3.9108738426, versus 0.1.

Preserve the two scopes through the existing shared composition/time-map representation, rather than adding frontend easing or a renderer special case. Cover default options, constructor-level linear, play-level linear over default children, and a non-linear outer group independently. Keep this fixture's play-level call: moving the rate into the constructor would test a different contract and hide this failure.

### [P1] Admit the active matching presentation in the correct same-z painter order

This is independent of timing: at t=0 the same source geometry and paint already composite incorrectly. At image pixel (696,270), Manim RGB is (233,109,99), Noon is (234,184,180), on both backends. This pixel is inside the overlapping red kite/right white circle, not an antialiasing edge. These are respectively red-over-white versus white-over-red.

Manim's newly added animation group is after the surviving circles during the animation (`Scene.add_mobjects_from_animations` / `Scene.add`); Noon debug publication remains [before id0, triangle id2, kite id3, after id1]. The existing native regression only checks that the original target is appended at cleanup. It does not establish the active matching order. At t=1.5 the discrepancy is again visible on the moving triangle; the final target order at t=2 is correct.

Fix the shared activation/publication ordering with existing identity and locality rules, preserving authored source/target topology, then verify order at begin and interior samples plus exact cleanup/re-add. Do not fix only the final swap, change z-index, reorder the fixture's self.add call, or widen raster limits.

These findings remain in #1571/#1616's existing scope. Keep the stack draft and unmerged. The later assertion-only fixture head `5b32772` has not been qualified by this run; source provenance above is exact.

yongkyuns:
Continuing the timing-scope P1 from review 5216368077 on this existing stack. I am preparing a bounded adapter correction against `0619d5726f4eeee86d30f4ea115157bb711203f1`: preserve constructor options on the existing Rust matching child and carry Scene.play timing/rate on an enclosing existing canonical composition node. No Python interpolation, alternative matching algorithm, renderer changes, fixture call changes, or tolerance changes. Qualification will distinguish default/constructor-linear/play-linear/nonlinear-outer cases. The independent active same-z painter-order blocker remains open and is not part of this timing correction. Please preserve concurrent #1616 assertion/harness changes; neither PR is merge-ready.

yongkyuns:
Coordination: found your timing-scope candidate `443d49d303a3015c98678757349c384a85e5dfd2`; I am reusing it verbatim rather than publishing my parallel adapter/test draft. This #82 continuation owns only the shared active painter-order repair (four files under crates/noon). Combined qualification 35032939020 integrates `443d49d`, the existing harness candidate `e3fd12f`, and #1616 fixture correction `8250da3`, with no motion/tolerance edits. First isolated order candidate passed the full noon library test step in 35032644831. No public branch overwrite, PR merge, or raster acceptance has occurred.

yongkyuns:
Superseded by #1640, now squash-merged into master at b9fff0e4df2aefd2f4230fb4bd616ff5df31eb3a after all 12 exact-head CI workflows passed. The reconstruction preserves current structural-family completion, plotting, and additional source/target alias preflight. Closing this old stacked draft; its branch remains available for provenance. Qualification/demo closeout continues in #1641.

PR #1628, opened 2026-09-16
Fix playground timeline and source-edit replay lifecycle

yongkyuns:
## Summary
- keep source-owned Python continuation playback non-seekable during the first authoring pass, then reattach the same semantic context as an ordinary replay lease once source execution completes so the timeline is genuinely draggable
- keep the currently running preview completely unchanged when Python source is edited or Reset; edits only update draft/footer state and wait for explicit Run
- allow explicit Run while a Python-owned animation is still executing: invalidate that run generation, cancel its existing source continuation through the authoring protocol, retire that runtime lease, and start the newest source immediately
- if the current scene is already an ordinary replay, let it continue while the next Python source begins authoring and switch only when the new execution handoff is ready
- report actual playback phase (`Playing`, `Paused`, `Finished`) instead of leaving the UI in `Building` while execution is visibly playing
- prevent stale metrics/runtime results from overwriting the current playback state after an explicit rerun

## Architecture
The browser remains a lifecycle/control shell. The first-pass continuation still uses the canonical source-owned semantic execution path and its existing prohibition on seek/pause. Replay controls are enabled only after the same semantic context is reattached without continuation ownership; no JavaScript playback clock, scene mirror, or second scheduler is introduced. Source edits perform no execution action. Explicit Run is the sole same-example supersession boundary.

The automatic post-paint warmup still exercises the normal Run path once, but the old edit-triggered `LatestSourceRunner` hookup has been removed from the live bootstrap. The standalone coalescer remains available for other bounded consumers but no longer owns playground editing behavior.

## Tests
Adds/updates regressions for:
- same-context continuation -> seekable replay promotion
- controls unavailable during source ownership and available after replay promotion
- source edit and Reset remaining execution-inert while a preview continues
- explicit Run invalidating/cancelling an active continuation before replacement
- Run staying available once a source-owned animation is visibly running
- transactional duration commit after continuation/replay freshness checks
- stale metrics suppression and playback-aware status polling
- initial authoring warmup without edit-triggered reruns

Fresh exact-head CI is the validation gate. Keep this PR draft until the browser/recovery/product workflows qualify the revised lifecycle.


yongkyuns:
## Continuation checkpoint — `a90a044687580c279467270437c5adb29397aff6`

Fixed the failing qualification contract and strengthened the tests without changing the runtime implementation in this tranche.

### Changes
- `d8dfce1`: replace the obsolete requirement for a literal `for (const rows of [5, 7, 20])` loop with checks for the explicit 5/7/20-row edit/play/rerun sequence. Keep the canonical stress source, dense workload, and existing semantic assertions intact.
- `a7ebcf0`: extend the browser stress scenario with live runtime frame observations and run-generation checks. Exercise Reset and typing during source-owned playback, rows=7 replacement while rows=5 is active, and a subsequent keyboard-triggered rows=20 Run. Use the existing execution-metrics hook, not the DOM frame counter, which is not polled during authoring.
- `a90a044`: add four executable tests in `web/playground-rerun-runtime.test.mjs`. They execute the production handlers extracted from `main.js`, with the real `PlaygroundGeneration` and fake DOM/worker boundaries. Cover execution-inert edit/Reset; immediate freshness invalidation and cancellation before waiting for interpreter cleanup; failed-cancellation worker retirement; and late cleanup preserving replacement owners.

### Validation observed
- **Current-head Web static and unit checks: success**, including `Run web preflight`: https://github.com/yongkyuns/noon/actions/runs/35097222386/job/104797471862
- Current-head Layer Dependency Ratchet: success.
- Focused local validation: `node --check web/playground-rerun-runtime.test.mjs` and `node --test web/playground-rerun-runtime.test.mjs` passed (4 tests). The local harness used retrieved production-handler sections, not a complete repository checkout; the complete generation module matched Git blob `3da51e6e3b6d6ea08e6af48695e662d582774f61`.
- Five deliberate local mutations were rejected: edit-triggered invalidation; missing cancellation freshness invalidation; waiting for natural completion before cancellation; stale cleanup clearing the replacement runtime; retaining a dead authoring client. Restoring the original handler sections returned all four tests to green.
- Full local `scripts/check.sh` was **not** run: no complete local checkout/Rust toolchain was available, and the Git clone attempt failed on DNS resolution.

The latest-head browser, authoring-recovery, Product Gate, and remaining PR Fast Gate work had not completed at this checkpoint. Earlier-head successes are not a substitute. The strengthened real-browser scenario is added but is **not yet browser-qualified**. Keep this PR draft and unmerged until the final-head gates finish successfully.

Issue #1639, opened 2026-09-16
Agent handoff: integrated Phase B checkpoint and remaining cleanup


yongkyuns:
## Purpose and authority

Continue from current source, not obsolete draft-stack descriptions. Optimize for clean, scalable architecture, readable code, modularity/reuse, meaningful regression coverage, and feature slices that compose with the existing engine. Read `AGENTS.md`, `AGENTS.override.md`, and `docs/architecture.md` before changing code. The architecture document remains the single normative authority; this issue is an execution handoff.

## Reconciled checkpoint — September 17, 2026 (Toronto)

Reviewed master: `be26da28e370e59efeb15f021f7eaf74e7dfd984` (September 18 UTC). The earlier body described the September 16 draft-stack situation and is superseded by this checkpoint. Historical discussion and qualification evidence remain in the comments and linked PRs.

Already integrated:

- Plotting and the geometry-presence locality repair: #1637. Explicit-range linear 2D axes, bounded function/parametric/sample/synchronized/gapped plots, native-Text numeric labels, and paired executable evidence are on master. The old sample-211 all-geometry full-rebuild blocker is fixed; do not weaken or reimplement that regression.
- TransformMatchingShapes product/lifecycle/evidence: #1640/#1641, followed by duplicate-key and default-leftover work in #1647. Do not replay #1614/#1616 as though the implementation were absent.
- Immutable image resources and retained image rendering: #1642.
- Playground source-continuation/replay lifecycle: #1643, with cancellation/run-routing and center-precision repairs in #1650.
- Scene geometry publication, family/tracker/current queries, and five primitive convenience constructors: #1644/#1646/#1651.
- Arrow family-copy repair and raster qualification: #1645/#1638.
- MarkupText/text-raster work: #1648.
- The INS/GNSS tutorial was deliberately removed in #1633, including discovery/workflow/build hooks. It is not current work and must not be restored as incidental cleanup. The obsolete native static-plotting example was also removed during integration.

The reviewed source tree is identical to #1651 head `cc1488b95a6cf36b999e0b1c783baf6b63ea91b9`, for which all 14 PR workflows passed. This does not make future changes qualified automatically.

## Current audit repairs — review exact heads before landing

[...]

PR #1647, opened 2026-09-17
Fix matching leftovers and add a separate duplicate breadth demo

yongkyuns:
Advances #1639/#954 beyond unique shape signatures and equal family sizes. Shared lowering keeps unmatched target FadeIn at its authored position while source leftovers retain directional displacement. Completion retires transient display rows atomically with original-target replacement. Python remains a thin wrapper; `key_map`, `transform_mismatches`, and `fade_transform_mismatches` remain explicitly unsupported.

Adds paired native Rust/direct-WASM and Python duplicate-breadth gallery demos, midpoint/paint/lifecycle checks, and pinned Manim growth/shrink fixtures. The original lifecycle demo retains source re-add and identity checks.

The shared renderer follows Cairo source-color quantization and computes pixel-area coverage once for eligible convex triangles/quadrilaterals, including stable morph fan meshes. Whole-interval convexity checks, bounded boundary recovery, and conservative fallbacks retain ordinary paths. `docs/architecture.md` documents renderer ownership; no extra semantic or runtime authority is introduced.

Current reviewed head: `4c8f5a5db4382b51d993f23e54fd54bdc5e0be0c`.

Validation: renderer library tests and WGSL validation pass. The combined #1647 + #1648 tree passes architecture, formatting, workspace check, strict Clippy, 2,359 Rust tests (0 failed; 5 ignored), Node/Python preflight, and release WASM package validation (165 public API contracts). The changed mega-mesh test uses an unrelated pentagon to preserve its original coverage after rectangles moved to the polygon route.

The previous head passed the complete canonical raster and sparse/dense playback suites. The combined text branch exposed repeated product FPS pressure, so this head adds a guarded convex-polygon interior/exterior fast path. Ambiguous edges retain exact clipping; f32 arithmetic guards and independent f64 clipping tests cover both windings, triangles/quads, large coordinates, thin polygons, and pixel boundaries.

The exact combined tree passes `bash scripts/check.sh full`, including all 2,359 Rust tests, strict Clippy, Node/Python preflight, and release WASM validation. All 15 current-head workflows pass, including canonical raster, sparse/dense playback, all three browser galleries, plotting, arrow regression, and product qualification. Product FPS retained ~98% of baseline on the first run (31.7 vs 32.3; minimum 80%). No raster or performance thresholds changed.

#1648 includes this head and qualifies the combined 20-entry gallery; merge this PR first.


yongkyuns:
Review: the shared lowering change and completion cleanup preserve compiler/runtime ownership; I found no additional semantic defect in this diff. This is not ready to merge because the new growth raster fixture still fails the unchanged canonical gate on both backends: frame 0015 has bounds delta 2 px (limit 1), and frame 0045 has 5 px (limit 1).

I inspected the current CI artifact (run 35221474748). At frame 0045 the fading red triangle contains Cairo RGB (14, 5, 4) versus Noon (14, 5, 5). Those values lie on opposite sides of the existing foreground distance threshold, accounting for the measured extra extent. An isolated libcairo fill reproduces the reference value at alpha 0.0560829639. This points to renderer paint quantization/edge coverage rather than a reason to change matching semantics. The frame 0015 edge discrepancy also remains. No raster thresholds or fixture samples were changed; keeping the PR draft until representative raster qualification passes.

yongkyuns:
Review follow-up: the renderer already implements Cairo's 16-bit-to-8-bit solid-source quantization in analytic.wgsl, but retained paths still used nearest-UNORM conversion of full-precision paint. Published 3e79debef4ad2a68a09a86b8bd993b2bf565e32f makes paths use that same established policy, without changing raster coverage, resource ownership, or dirty work.

Added an exact interior pixel assertion to the paired direct-WASM matching smoke at 0.75s: RED at .9 × .25 alpha must be (56,22,19), matching Cairo; the old path output was (57,22,19). This tests the actual renderer output rather than a duplicate formula. Architecture gate, JS syntax, diff checks, and Naga WGSL parse/validation passed. Current-head CI will validate browser output and the existing raster fixtures; the frame-0015 edge coverage discrepancy is still under investigation and no thresholds changed.

yongkyuns:
Remaining review blocker after the paint-quantization repair: the duplicate-growth frame 15 fixture measures Cairo bounds x=143..830 versus Noon x=144..829 (width delta 2, unchanged gate 1). Independent renderer review traced both edges to faint area coverage: the left edge is an animated matched triangle, while the right is the fading four-vertex target kite. Cairo's left-edge pixel [5,9,10] crosses the foreground cutoff; GPU's corresponding edge does not. The kite's far edge similarly differs at low coverage.

The Arrow PR's exact triangle pipeline cannot safely cover this case: it intentionally excludes morphing meshes and only carries three static vertices. Animated triangle support needs source/target coverage geometry with local dirty updates; convex polygon support needs unified polygon coverage so internal triangulation edges are not overblended. An alpha nudge or weaker tolerance would conceal the rendering gap. Keep this fixture and leave this PR unmerged until the coverage issue is implemented and qualified. Semantic lifecycle review found no additional blocker. The integrated candidate also needs both compact and full path vertex entry points to use cairo_source_color after Arrow merges; the ordinary Git merge does not catch that helper-call conflict.

yongkyuns:
Final raster result for 3e79debef4ad2a68a09a86b8bd993b2bf565e32f confirms the diagnosis: paint quantization fixed both frame-45 failures. The only remaining failures are duplicate-growth frame 15 on WebGPU and WebGL, each with width delta 2 > 1. All other canonical raster checkpoints passed. This PR remains open for the area-coverage follow-up described above.

yongkyuns:
Published `a1d9ecbe2f8775a7d4c4b65a782c6cde9fcfebcb` after self-review.

The previous Plotting qualification failure was in its full-repository test gate, before plotting ran: both `complex_filled_morph_mega` cases expected an unrelated rectangle to use the mega-mesh path. Convex rectangles now use polygon coverage. The fixture now uses an unrelated five-corner path so it continues testing that sampled morph updates neither dirty nor index the unrelated mega attributes. Both focused tests pass; assertions and performance/raster thresholds are unchanged.

`docs/architecture.md` now records the renderer-owned polygon specialization, whole-interval morph eligibility, and ordinary fallback. The combined Matching + MarkupText browser package also builds and validates all 165 public API contracts. Current-head CI remains the merge gate; the prior canonical raster run was superseded before it finished, so no raster pass is claimed yet.

yongkyuns:
Current-head product gate missed only the relative FPS check (baseline 46.2, candidate 35.7, required 36.9); pixel difference was 0%, and cold/warm/edit latency checks passed. The immediately preceding head has identical runtime code—the latest commit changes only a regression-test fixture and architecture prose—and passed this same product comparison at baseline 37.0 / candidate 37.9 FPS. Given the baseline variation, I requested one rerun of the failed product job to distinguish runner variance. No performance threshold or scene has changed. A repeated failure remains blocking and will be investigated.

yongkyuns:
Holding merge for a further performance repair. Although this head passed the product rerun (baseline44.5/candidate37.0 FPS, 83%), the combined #1648 head missed again at60.2/46.0 (76%), on the same Create(circle)/Create(square) product fixture. The polygon shader's full per-pixel clipping is a plausible shared cost, so runner variance alone is not enough to dismiss this. Implementing conservative interior/exterior coverage fast paths while retaining full clipping at edges; thresholds stay unchanged. Canonical raster enforcement on this head passed; sparse/dense playback is still running.

yongkyuns:
Current-head raster qualification is now fully green, including canonical ratchets and shared sparse/dense playback (run 35250901556). I am retaining draft status while a reviewed polygon interior/exterior fast path is validated: the combined text branch missed the product FPS floor, so the prior rerun is not sufficient evidence to dismiss shader cost. The new fast path retains exact clipping at ambiguous edges and has scale-aware f32-versus-f64 coverage checks. No thresholds change.

yongkyuns:
Reviewed head 4c8f5a5db4382b51d993f23e54fd54bdc5e0be0c passed the product gate on its first run: baseline 32.3 FPS, candidate 31.7 FPS (~98% retention; required 80%). Runtime job 105324721998 also passed visual/latency comparison. This resolves the prior measured FPS pressure with the guarded shader fast path, without changing thresholds. Remaining current-head raster/cross-browser/plotting/arrow checks are still running.

PR #1652, opened 2026-09-18
fix: keep Arrow constructors on Scene-owned publication

yongkyuns:
## Review finding

The September 16–17 architecture review found that `Scene::manim_arrow` still called the raw-store constructor after the Scene owned execution, unlike the primitive constructors fixed in #1651. Detached Arrow/Vector/DoubleArrow creation could therefore leave the running publication on a stale scene revision.

## Bounded repair

- Route running Scene Arrow-family creation through the existing execution preflight and atomic semantic publication.
- Share one Arrow preparation/publication helper with raw-store integration; remove the old nested `commit_prepared`/`commit_transaction` wrappers.
- Reuse `SemanticStore::with_geometry_paths` for both tip resources. No new runtime, state mirror, context trait, or compatibility API.
- Preserve explicit membership/admission and exact cold Arrow behavior; update one caller to mutable Scene access.
- Add regressions for all three variants, detached frame locality, coherent later admission, stale rejection before resource import, and a real post-import lowering failure with both new tip resources rolled back while existing state survives.

## Validation completed before publication

Isolated qualification run 35303033421 passed:

- pinned `cargo fmt --all` and diff checks;
- `cargo test -p noon --no-default-features --lib --test manim_arrow_geometry --test arrow_tip_regression --test family_copy`;
- `cargo clippy -p noon --no-default-features --all-targets -- -D warnings`;
- negative control: restoring the old Scene raw-store route fails the exact new running-constructor regression (0 passed, 1 failed).

The publication job verified the candidate tree and an exact four-file allowlist before publishing. Diagnostic workflows/patches are NOT in this PR's tree or history.

## Completion review — September 18, 2026

Reviewed exact head `8e424f99d69f984733dd72da53ef05db178dd1a9`; no blocking findings or source amendments required. All 10 reported normal PR workflows succeeded: PR Fast Gate, Architecture Ratchets, Layer Dependency Ratchet, Native Host Smoke, Provider feature isolation, Playground Product Gate, Playground cross-browser matrix, Arrow tip Manim regression, Plotting qualification, and Image raster qualification.

Inspected execution of the relevant jobs, including Rust tests/format/lint and native Arrow geometry plus ManimCE 0.21.0 Cairo pixel qualification on WebGPU/WebGL at DPR 1 and 2. These are verified CI results, not a local Rust rerun. Detailed source review is recorded on the PR.

Refs #1603. The broader public Scene/LiveSession control migration, callback integrations, vector-field construction, and other constructor families remain outside this repair.

Issue #1603, opened 2026-09-15
[Cleanup W2] Finish Scene/live control convergence and remove residual public mode split


yongkyuns:
Parent: #1480. Follows completed ownership/routing slices #1496/#1510.

## Current checkpoint — September 17, 2026 (Toronto)

Reviewed master: `be26da28e370e59efeb15f021f7eaf74e7dfd984`.

`Scene::geometry` and the five `circle`/`square`/`rectangle`/`line`/`path` convenience constructors now support detached creation before and after owned execution bootstrap through the shared atomic publication path. Family/tracker/current-query convergence has also advanced. Do not replay the closed #1611 or claim these routes are still missing.

The audit found the remaining raw-store bypass in `Scene::manim_arrow`. Draft #1652 repairs Arrow/Vector/DoubleArrow publication with running/coherence, stale-before-import and late-lowering rollback regressions; its 444 focused/minimal-provider tests and strict all-target Clippy passed in isolated qualification. Check the PR's live head and normal CI before treating it as landed.

`LiveSession::create_manim_geometry` still has real borrowed external-session/callback callers. Those capabilities must be classified before deleting the route; a name alone is not evidence that a wrapper is redundant. Other constructor families and the broader public control surface remain open.

## Why this cleanup exists

Scene owns the running execution component and generic persistent mutation routes through Scene-owned publication, but residual authored-versus-live API mode distinctions still expose implementation details. The authored/base versus effective-runtime state distinction must remain; ordinary same-semantic operations should not require callers to choose a separate feature API merely because execution has started.

## Goal

Make Scene the ordinary durable operation surface before and after bootstrap. Keep LiveSession/ExecutionSession narrow and internal/advanced only where an irreducible execution-context capability exists.

## Scope

1. Inventory remaining public or broadly used LiveSession operations and classify each as a same-semantic persistent operation, an explicit effective/current query, or a genuinely execution/segment-only capability.
2. Migrate constructor/edit/query call sites that differ only because execution is running.
3. Delete redundant `live_session/` forwarding/feature modules when their last callers move; do not retain compatibility aliases in this greenfield project.
4. Internalize ExecutionSession exposure where ordinary callers no longer need it, consistent with #1484.
5. Retain ExecutionSegment/continuation concepts only where they model real play/wait barrier semantics.
6. Update examples/docs/tests so callers do not branch on cold-versus-running mode for the same semantic operation.

## Architecture constraints

[...]

PR #1656, opened 2026-09-18
refactor: unify Scene family translation and arrangement across bootstrap

yongkyuns:
## Scope

Refs #1603. This is a bounded Scene/live convergence slice, independent of the Arrow constructor repair #1652. It does not close the umbrella issue.

The same existing public operations now work before and after Scene-owned execution bootstrap:

- `Scene::shift_family`
- `Scene::arrange_family_with_options`
- `Scene::arrange_family_in_grid_with_options`

Cold operations prepare from authored state and commit through Scene's existing transaction route. Running operations retain the existing effective observations, stale-context preflight, active-affine-driver restrictions and coherent publication helpers. Cold edits do not bootstrap execution: `Scene::effective` / `effective_family_layout` continue to fail explicitly when no effective observation exists.

## Implementation and ownership

Extract the existing authored arrangement preparation into a private method on the existing `FamilyArrangePlan`. Both raw-family cold authoring and the durable Scene route use it; the bounds/anchor/grid algorithms are unchanged. No new context trait, public control type, semantic store, identity allocator, cache, runtime, scheduler, serialization boundary or compatibility adapter.

Migrate cold authoring in the paired family-arrangement and family-grid Rust examples to Scene. Their normal native/direct-WASM execution and Python counterparts remain the qualification path; no Python semantic implementation or gallery case changes.

Delete `LiveSession::arrange_family` and `LiveSession::arrange_family_in_grid`. They had no production callers; remaining test callers now use the existing options-based operations. Keep those options-based borrowed operations: native continuations and WASM integrations genuinely borrow an external execution/store/root. This PR does not disguise that remaining control-ownership work as completed migration.

Locality is unchanged: preparation traverses the affected family and its required request-local arrangement data; mutations use the existing per-leaf transaction/publication machinery. No whole-scene or resource-store scan/clone, relowering fallback, or resource regeneration was added.

## Regression evidence

- Identical Scene calls in cold and owned-running modes, for linear and grid placement, both detached and admitted (eight combinations).
- Stable object/family identities, exactly one semantic revision, two affected leaves, unchanged unrelated authored state/resources, and no rendered-row changes for detached edits.
- Later admission publishes the arranged positions coherently.
- Nested aliases translate each leaf exactly once in both modes.
- Invalid options, foreign families and a deliberately stale owned execution reject without changing state/revision/publication.
- Running arrangement/grid retain active-affine-driver rejection.
- Existing retained-path arrangement tests still distinguish boundary anchors from geometric dimensions.

## Validation

Production head: `0753af356c0b9509a42d28bcf85db1716a294724`.
Exact source tree: `81fda0c9d74b33e330fa5e7443f55e02db9f1ccf`.

Isolated qualification run **35343023511** passed using pinned Rust **1.98.1**:

```sh
cargo fmt --all
bash scripts/check.sh architecture HEAD
cargo test -p noon --no-default-features --lib --test family_arrangement --test family_grid --test path_family_arrangement --test family_translation --test family_callback_translation --test live_family_layout
cargo clippy -p noon --no-default-features --all-targets -- -D warnings
cargo fmt --all -- --check
git diff --check
```

**453 tests passed: 430 library + 23 focused integration tests.** Two negative-control runs restore the old Scene routes, leaving the new tests intact. Each fails its intended test with `EffectiveStateUnavailable` (0 passed, 1 failed); restoring the candidate restores the exact recorded tree. No compile failure is counted as a negative-control success.

Downloaded and SHA-256-verified artifact **10546080757**, inspected the test/Clippy/negative-control logs, and reproduced the qualified tree locally. Publication verifies the artifact digest, patch digest, exact tree, nine-file allowlist and single production parent before a non-force push. Temporary preparation workflows and payloads are excluded from this PR's tree and history.

Local `git diff --check` passed. Local `bash scripts/check.sh fast HEAD` was attempted but stops at the prerequisite check because Cargo is not installed in this container. No full local, all-feature, browser-raster or exact-head normal-PR-CI pass is claimed yet; the PR remains draft for those checks.

## Before / after inventory

- Public `LiveSession` functions across its implementation files, counting all feature configurations: **146 -> 144**; precisely the two convenience wrappers above are deleted.
- Main `live_session.rs` production section, excluding its inline test modules: **2,470 -> 2,434 raw lines** (-36).
- Public control types and production modules added/removed: **0 / 0**.
- Across the six changed production files, including documentation/examples and excluding inline test modules: **3,997 -> 4,018 raw lines** (+21). The complete diff is 9 files, +425/-100, mostly regression coverage. This is an API-mode reduction, not a claim of net production-LOC reduction or completion of #1603's wider deletion target.

Remaining #1603 work includes other constructor families, other running-only placement operations, ordinary control/continuation ownership, and classification/migration of genuine borrowed callback capabilities. No new temporary engine seam was introduced.


yongkyuns:
Normal PR CI found one integration gap outside the isolated candidate allowlist: `fixtures/provider-consumer/tests/object_authoring_errors.rs` still called the two deleted `LiveSession::arrange_family` conveniences. That made every provider-consumer matrix entry fail to compile, while the other eight workflows on the first head passed.

Follow-up `a767449` migrates those two fixture calls to the retained explicit `arrange_family_with_options` route; no production code or behavior changed. This is the intended classification: borrowed external-session tests keep the explicit options-based execution capability, while the redundant convenience wrapper remains deleted. A fresh full PR workflow set is running on the repaired head. I am not counting the previous provider failures as flaky or rerunning them unchanged.

PR #1669, opened 2026-09-18
Add static implicit curves through shared retained geometry

yongkyuns:
Static implicit curves now work through the public Rust facade and Python `ImplicitFunction` / `Axes.plot_implicit_curve`. The existing adaptive isoline planner produces ordinary retained paths; scalar callbacks run only during bounded preparation, before scene identity or resources are published.

Smoothing reuses the shared spline solver with an explicit pinned-Manim boundary policy. Coordinate mapping uses one captured Axes frame, including transformed bases. Python owns only argument/callable adaptation and releases its callback proxy on success or failure. No new renderer primitive, runtime subsystem, crate, or dependency is introduced.

Includes paired native Rust/Python gallery examples, a captured thumbnail, and extensions to the existing semantic, raster and plotting harnesses. Dynamic field resampling, nonlinear coordinates and retained callable metadata remain unsupported.

Validation:
- 96/96 semantic fixtures match pinned ManimCE 0.21.0, including 10 implicit-curve cases and actual Pyodide callback/error/snapshot checks.
- Rust/Python demo pixels are identical on WebGPU and WebGL2; sparse/dense playback is identical on both.
- The focused implicit raster fixture passes the unchanged 0.6% gate at 0.35% on both backends. It draws mapped contours; coordinate shaft/tick rendering remains separately qualified.
- Focused native spline/implicit tests, Python adapter tests, gallery checks and browser preflight pass.
- Full repository gate passed: 2,395 Rust tests, architecture constraints, formatting/clippy, browser preflight and 232 WASM API contracts (`NOON_WASM_PROFILE=dev NOON_RENDERER_SMOKE=1 bash scripts/check.sh full`, shared Cargo cache, two compiler jobs). All active GitHub checks have now passed, including raster/playback, performance, provider matrix, native presentation and Chromium/Firefox/WebKit. Human approval is still required by repository rules; auto-merge is disabled.

Refs #856, #85, #954.


yongkyuns:
Superseded by #1668, which includes the same implicit-curve implementation alongside NumberPlane and has passed all required CI checks. Implementation and qualification history remain in both PRs.

PR #1672, opened 2026-09-18
Add animated plotting and number-line demos

yongkyuns:
## Current status

Repair head **4d69b7a18a01e90382493c17318609d576a124ea**, tree **2d1ca64176476ec07caa0a0fb930d3e9cf1a633c**. The branch advanced without force from 3321111. The dedicated static, animated and live-coordinate browser checks pass with the corrected source. **Fresh normal PR CI is still required before merge; no deployment is claimed.**

## Actual cause of the repeated failure

`Python source suspended without a semantic continuation consumer` came from **scripts/plotting-qualification.mjs loading coordinate_plotting.py as a zero-duration static scene**, using `client.run(source)` without `onSemanticContinuation`. This PR had changed that shared input into an animation. The failed report from run 35403861147/artifact 10571991529 identifies this capture call on both backends.

It was **not** a NumberLine admission or post-play n2p limitation. The earlier re-add/precomputed-target workarounds did not address this error and have been removed. The corrected real Pyodide sequence performs n2p and p2n after the shifted NumberLine finishes, successfully. Comment 5737451592 records the diagnosis correction.

## Delivered examples and shared implementation

- **CoordinatePlotting**: Write title, Create real axes, draw the sine curve, reveal the sampled-data path, then hold. Total **5.1 seconds**, **17 objects**. Curves still use the existing Rust plotting/sampling API.
- **AnimatedNumberLine**: construct the real NumberLine and retained numeric labels including zero; Create it; move a marker -4 → 0 → 3; shift the line and marker together; query the new live frame with n2p/p2n and move to -2. Total **8.8 seconds**, **22 objects**. No cached targets, hand-drawn coordinate substitute or Python coordinate arithmetic.
- Both have **typed Rust LiveProgram examples shared unchanged by native and direct WASM**. Native entry points are `cargo run -p noon-native --example coordinate_plotting` and `--example animated_number_line`.
- The static coordinate scene uses the same Rust object builder and remains explicitly qualified against its original Python source, now named `coordinate_plotting_static.py`. It is not substituted for the public animated gallery entry.

No runtime semantics, renderer, dependency, semantic store, scheduler or frontend clock was added. Parallel marker FadeIn and caption Write have explicitly equal half-second durations in both examples, rather than different omitted leaf-duration defaults. Gallery metadata is updated without the earlier whole-file formatting churn.

## Qualification repair

The original static comparison and resumed-authoring checks are preserved. Both public animations are additionally registered with the **existing continuation-aware time-series plotting harness**, not a new host execution path. The normal plotting workflow now runs both new modes and every existing static/single/synchronized/gapped/live case after its full native gate.

New checks cover partial curve/data reveal, absence before admission, marker positions at intermediate times, transformed live coordinate round trips, exact total duration, and unchanged pictures during waits. Renderer timestamps remain separate from engine acknowledgements: a quiet interval can retain its last visual timestamp, but the start of the next animation must publish its new boundary timestamp. Neither value is replaced with the request.

The renderer-comparison threshold remains **zero differing pixels**. No original fixture, wait, timing tolerance, argument validation or existing assertion was removed. Static and animated cases are tested separately because their continuation requirements are genuinely different.

## Completed tests and independently decoded images

- **395 web tests passed** locally and in the isolated candidate runs. All four new harness-policy tests pass; the old harness fails three of those four with zero cancelled tests.
- **15 plotting/numeric-label Python adapter tests passed** locally.
- Pinned Rust Actions runs **35406008188**, **35406403477**, and **35407035487** passed both new native integration tests plus the existing static-coordinate test. These exercise ordered object admission, intermediate marker positions and the completed transformed coordinate frame. The full normal plotting gate explicitly includes integration tests.
- Final browser verification **35407486920** passed original static/Rust-Python equality and resumed authoring, **24 animated-coordinate frame pairs**, **42 animated-NumberLine frame pairs**, and **18 existing live-coordinate frame pairs** on **WebGPU and WebGL2**.
- Downloaded final artifact **10573236847**, verified ZIP SHA256 **c9a6eeb4c79ef702c97175ea61edacbc37dd13698c181b3b4386c196b2245cba**, and independently decoded every one of those **84 PNG pairs: all are exactly RGBA-identical**. Inspected the partial plot and shifted NumberLine frames visually. Both new examples finish with the expected counts and durations, and error-free source execution.

This is **Python versus shared Rust/direct-WASM parity within each backend**, not a new blanket claim of full upstream Manim parity or identical native-window pixels. The gallery remains a parity candidate; broader UnitInterval/coordinate options are not claimed by this PR.

## Provenance and remaining gate

The native/WASM runtime was built from isolated candidate **b7586c64d14af584fbf80b35b9ee063701170fde**, tree **b64e755891b8010be44691874fbca152857c2f67**, in 35407035487. That run intentionally retained the two differing fade-midpoint images rather than weakening the comparison. The final diagnostic 35407486920 reused its hash-verified unchanged Rust/worker/glue and changed **only the Python example's explicit parallel leaf durations**. It recorded test source **2b4921a79409a25e58ac795ae47fc26fd71d9c24** and the intended tree **2d1ca641…**, without relabeling the retained binary's build identity. Normal current-head CI recompiles and qualifies the final PR independently.

Publication restored the required plotting workflow through the authorized connector; the complete GitHub tree exactly matches the independently reconstructed/tested tree. Temporary prep workflows, staging JSON, downloaded artifacts and generated packages are excluded from the PR.

Local `scripts/check.sh fast HEAD` was attempted but Cargo is unavailable. Full local web preflight attempts timed out during the pre-existing Python subprocess suite, so no complete local preflight or local Rust/browser pass is claimed. Node/adapter tests, source verification and image comparisons were local; native/browser execution evidence is from the recorded Actions runs. Earlier preparation errors (event-checkout provenance mismatch and the test's wait-boundary expectation) are retained in the discussion/artifacts and are not presented as passing.

Merge only after the final-head required checks and review are green, as authorized by the user.


yongkyuns:
Root cause correction: the repeated `Python source suspended without a semantic continuation consumer` is raised by `scripts/plotting-qualification.mjs` while loading **coordinate_plotting.py**, not by the NumberLine's n2p query. That harness calls `client.run(source)` without `onSemanticContinuation` and explicitly requires zero duration; the PR changed its shared source into a five-second animation. The two backend failure stacks in artifact 10571991529 identify `capture` at this call. My earlier live-query diagnoses were not supported and the subsequent NumberLine workarounds did not address this failure.

The correction preserves the exact original static scene as a separate static qualification fixture, and adds both actual animated public examples to the existing continuation-aware time-series raster harness. Native/direct-WASM counterparts are included, using the same typed Rust live-program path, with native tests for ordered plot admission and NumberLine post-transform coordinates. The NumberLine's transformed live n2p/p2n queries are restored, rather than deleting coverage to accommodate a static harness. Gallery registration assertions that previous string replacements did not actually add are now present and locally executed. Metadata formatting is reduced back toward the base format.

Source changes are being checked on isolated prep/1672-animated-qualification before advancing this PR. No release-ready or merge claim yet; all original static/lifecycle/raster assertions remain, and new animated tests cover partial reveals, waits and the moving marker.

PR #1673, opened 2026-09-19
B5: share retained area and Riemann helper semantics

yongkyuns:
Axes area and Riemann helpers now build ordinary retained paths and rectangle families from shared Rust coordinate semantics. Python and direct Rust use the same bounded partitions, signed-area rules, and atomic cold/live publication. Retained function-plot metadata supports the usual helper call shape without a Python geometry cache.

The paired native, direct-WASM, and Python area examples cover shaded regions and signed midpoint rectangles. Coordinate gallery thumbnails now show actual rendered examples.

Validation completed:
- 31 focused coordinate tests and 10 Python transport tests.
- 121 actual Rust/WASM/Python semantic probes against pinned ManimCE 0.21.
- Exact paired pixels and live plotting assertions on WebGPU and WebGL2.
- Pinned area raster comparison: 0.01% worst differing pixels on both backends, within the strict fixture policy.

Stacked on #1671. Full integrated validation and required CI remain merge gates. This completes the area-helper batch under #85 and #954; it does not close Phase B.


yongkyuns:
Follow-up #1677 is stacked on this PR's exact `5c71588e8acf56b6ffaf4b8ee469831aae732e36` head and changes only the Riemann preparation loop plus regression tests. All nine new Rust regressions pass; `cargo test --workspace --all-features --lib` completed successfully (1,932 passed, three existing display-dependent tests ignored). The real browser Manim-authoring step also passed.

Two inherited stack blockers were exposed by #1677's PR Fast Gate run `35450761656` (merge checkout `38ebe6677effb2080d72da6adf40a7325437324a`):
1. Strict lint: `crates/noon/src/text_authoring/semantic/objects.rs:178`, `text_artifact_state` has eight arguments, triggering `clippy::too_many_arguments` under `-D warnings`. The same eight-argument definition is present at this PR's base snapshot (file blob `757dd1744bf06d2363533d885ff64bac7da09244`). Job `105917382153`.
2. Web preflight: `web/manim-compatibility-gallery.test.mjs:16` still expects 23 gallery entries, but the manifest now has 24. Its explicit ID inventory also lacks the area example. The unchanged parent test blob is `a9dd9e6744eb9e2c3890d8e14f064500fb04b19b`. Job `105917382176`.

These files are outside #1677's three-file diff. Leaving the fixes with the owning text/area stack rather than editing active parent branches or folding unrelated changes into the partition patch. #1677 remains draft; no integrated gate pass or merge-readiness is claimed.

Issue #778, opened 2026-09-01
[Roadmap B5] Axes area and Riemann geometry


yongkyuns:
## Roadmap

Phase B5 child of #85/#954. Architecture authority: `docs/architecture.md`.

## Goal

Implement shared `Axes.get_area` / `Axes.get_riemann_rectangles` geometry using the same coordinate/path state and explicit authored-vs-effective query semantics as #695.

## Query/snapshot contract

These helpers create persistent geometry from the state they are invoked against. When invoked during a live execution session, source Axes/graph coordinate/path queries use one coherent **effective Runtime snapshot** (`FrameEpoch`) so active transform/drivers are visible consistently. The produced area's/rectangles' geometry then becomes authored/base content for the newly created semantic objects.

Do not mix authored Axes values with effective graph values (or vice versa) inside one helper call.

## `get_area`

- resolve requested interval against graph range metadata;
- clip against optional bounded graph using pinned v0.21 overlap rules;
- include exact endpoints and retained graph samples within the interval;
- close to x-axis baseline when unbounded;
- trace bounded graph B in reverse when bounded;
- output ordinary closed path/polygon content, not a renderer Area primitive.

## `get_riemann_rectangles`

- deterministic half-open partition equivalent to upstream range behavior;
- left/right/center sampling;
- optional bounded-graph baseline;
- `width_scale_factor` behavior;
- ordinary rectangle/path leaves;
- enough metadata for shared style/color behavior without frontend geometry recomputation.

## Architecture

- Rust owns interval clipping, graph-proportion lookup, coordinate conversions, baseline coordinates and closed geometry;
- live helper evaluation uses a coherent effective source snapshot;
- persistent authored source edits use normal semantic mutations; active transform/reactive motion does not rewrite source semantic declarations every frame;
- reuse #695/#696 coordinate/path state and B2 geometry;
- no renderer-specific area/Riemann types;
- Python performs argument/style adaptation only.

## Acceptance

- exact area point order for bounded/unbounded graphs;
- interval clipping/non-overlap behavior;
- helper results under authored transforms and under active effective Axes/graph transforms;
- no mixed-revision snapshot within one helper call;
- left/right/center sample positions and deterministic partition counts;
- signed/bounded baseline/width scaling behavior;
- static results require no per-frame host work;
- promoted cases pass B7/#185 qualification.

## Dependencies

#695 coordinate/live-query semantics, #696 retained graph/path sampling, B2 geometry and B4 text only for labels.

[...]

PR #1677, opened 2026-09-19
B5: unify bounded Riemann partition sampling

yongkyuns:
## Scope

Advances #778 / #85 / #954. Stacked on #1673 at `5c71588e8acf56b6ffaf4b8ee469831aae732e36`; this is a bounded follow-up, not a replacement for that implementation or closure of Phase B.

## Change

The retained Riemann path builder duplicated a sampling rule already implemented in `noon-geometry::PlotSamplingOptions`. Replace that second count/increment loop with the existing bounded planner. Exclude only the planner's separately appended terminal endpoint, rather than filtering regular samples by their rounded value.

This fixes three reproduced defects:
- A nonempty span whose positive length/step quotient underflows incorrectly created zero rectangles.
- Large offsets used the requested increment rather than NumPy's representable `(start + step) - start` increment, shifting rectangle positions.
- Repeated samples or regular samples rounded onto the interval end could be truncated.

The existing rectangle limit remains inclusive and unchanged. The planner's separately appended endpoint is not charged as a rectangle.

## Ownership and complexity

All production changes are in shared Rust authoring preparation. Native, direct WASM, and Python still converge on the same coordinate snapshots, retained-path lookup, paint preparation, and atomic path-family publication. No new public API, renderer primitive, runtime, language-side numerical policy, dependency, or migration seam.

Preparation remains O(requested samples) plus existing retained-path query costs, independent of unrelated scene size. Reusing the planner adds temporary bounded f64 parameter storage (about 8 MB at the one-million-rectangle ceiling); it creates no retained cache or per-frame work.

## Regression coverage added

Nine Rust tests cover exact half-open samples, quotient underflow, unused-increment overflow, positive/negative large offsets, repeated samples, exact admission limit, cold/live published rectangle geometry, distinct duplicate family members, and atomic rejection both before preparation and after a later sample exceeds the retained domain.

The existing real Manim/Pyodide smoke gate now exercises the same large-offset/repeated-sample family counts and physical rectangle starts through direct WASM and actual Python wrappers. Existing paired native/direct-WASM/Python area examples and raster assertions remain unchanged.

## Validation actually performed

- Reproduced the old arithmetic against installed NumPy: tiny span produced 0 vs 1 sample; +/-1e16 offsets produced [0,4,6] vs [0,4,8]; a sub-ULP step produced 3 vs 4 repeated samples.
- `node --input-type=module --check < web/plotting-smoke.js` passed.
- Compiled the embedded Python fixture with Python `compile(..., 'exec')`; syntax passed.
- Verified reconstructed parent file bytes against their exact Git blob SHAs and published file blobs against the locally checked bytes.
- Self-reviewed the production diff and unchanged publication/resource ownership.

This container has no Cargo/Rust toolchain. Rust compilation/tests, rustfmt, `bash scripts/check.sh fast`, `bash scripts/check.sh full`, actual WASM/Pyodide execution, and required GitHub CI are **not yet claimed as passed**. Keep draft until current-head qualification completes.

## Still open

The earlier #778 authored-function fast-path requirement remains unimplemented in the retained-only foundation: sampled-path interpolation is not exact evaluation of an authored function. This patch does not silently replace that requirement or claim its parity. Broader retained graph query/discontinuity behavior and extrapolation also remain separate follow-ups. Dynamic resampling stays Phase C-owned.

No automated reviewer was requested.


yongkyuns:
Validation update for head `955a570c77df1518216bd4911d70ba85f922093e`, tested by PR Fast Gate run `35450761656` on synthetic merge `38ebe6677effb2080d72da6adf40a7325437324a` against parent `5c71588e8acf56b6ffaf4b8ee469831aae732e36`.

Passed:
- `cargo test --workspace --all-features --lib`: 1,932 passed, 0 failed, three existing native display tests ignored. All nine `coordinate_authoring::area::tests` regressions are individually `ok` in job `105917382037`.
- `cargo fmt --all -- --check`.
- Browser fast-check job `105917382049` completed successfully: actual Manim-style authoring, typed WASM/Python error boundary, deterministic playback, SVG authoring and primary WebGPU rendering. The added direct-WASM/Python partition cases are in the existing plotting qualification called by that Manim-authoring gate.

The overall gate is still red due to two inherited, untouched parent-stack files:
- `text_authoring/semantic/objects.rs:178`: `text_artifact_state` has 8 arguments and fails strict Clippy's 7-argument limit. Job `105917382153`.
- `web/manim-compatibility-gallery.test.mjs:16`: stale expected gallery size 23 versus actual 24; the exact ID inventory also needs the area case. Job `105917382176`.

Verified both offending definitions at the pinned parent head and reported them on #1673 (comment 5742948414). No lint suppression, test weakening, parent-branch edit, or unrelated production refactor was added to this patch. The fast gate skipped its renderer-critical WebGL/recovery checks, so it is not full dual-backend qualification. Local full scripts and the wider required integration suite remain unclaimed. Keep this PR draft and unmerged until the owning stack is repaired and full current-head qualification is complete.

yongkyuns:
The next step is implemented as draft #1678, stacked on this exact `955a570c77df1518216bd4911d70ba85f922093e` head. Repair commit `f0c120e916ddde2714ed785b6c4114e6fb04ce83` changes four files (+23/-25): deletes the redundant text forwarding helper and borrows compiled font/geometry arenas, gates the two browser-only path-family adapter methods out of native builds, and repairs the exact gallery inventory with explicit area-example assertions.

The gallery test and repository-configured formatting passed before publication. The actual published repair is now running `scripts/check.sh fast` plus the normal PR checks, including plotting/full integrated validation and shared text qualification. No full gate pass is claimed yet. Parent branches remain unchanged; no workflow, dependency, lint exemption, weakened assertion, parity promotion or migration seam is included in the repair.

PR #1680, opened 2026-09-19
B6: harden camera viewport projection and qualify pan/zoom

yongkyuns:
## Scope

Advances B6.3 / #89 under #954 as an independent parallel slice, based on master `b7258363817b0c4d31f4becf8a5cc17b6605c139`. Head `fd35ba096a8991d20a88bf39e4197fdaf999dd1c`: **one code-only commit, two files, +273/-6**. It is not stacked on the active B4/B5/Riemann or C2/C5 work, and does not edit their execution-session, input, binding or gallery-registration files.

## Implementation

`Camera2DState::viewport_bounds` now widens intermediate arithmetic before multiplication, halving and translation, then validates the actual narrowed f32 endpoints. It rejects infinite endpoints and positive mathematical spans that collapse to one representable coordinate. Representable subnormal bounds and large-but-resolvable translations remain accepted; there is no arbitrary epsilon cutoff, clamping or silent fallback.

This remains a constant-time, allocation-free projection in the existing shared core camera module. No public signature, semantic identity, camera declaration, runtime, renderer, dependency, host authority or serialization boundary is added. There is no temporary migration seam. The temporary validation workflow is outside this PR's tree and ancestry.

Nine new numerical tests cover intermediate product overflow, translated endpoint overflow on both axes/signs, unrepresentable horizontal extent, horizontal/vertical underflow, translated precision collapse, valid subnormal bounds, valid large-center spans, and invalid camera/aspect inputs.

Two typed integration tests reuse ordinary `Scene::camera_frame`, target transforms, `LiveSession`, effective layout queries and `ExecutionSession::camera`. They qualify pan plus uniform zoom at 0, 25%, 50%, 75%, 59/60 and exact end, with four viewport aspects. A 1,000-unrelated-object fixture verifies unchanged authored state during interpolation, unchanged unrelated effective object state, camera-only dirty rows, normal endpoint reconciliation and read-only repeated camera observations. Reverse and forward direct seeks reproduce the forward camera and viewport samples exactly. This is dirty-state evidence, not a CPU-time or GPU-upload benchmark.

## Executed qualification

Isolated Actions run [35456559698](https://github.com/yongkyuns/noon/actions/runs/35456559698), job **105932769094**, succeeded. Artifact **10588675081**, `b6-camera-viewport-proof`, retains the full source, patch, identity and executable logs.

[...]

PR #1683, opened 2026-09-19
B6: preserve foreground ordering across scene edits

yongkyuns:
## Scope

Advances B6.1 / #89 under #954 with the still-missing Manim-compatible foreground persistence surface: `Scene.add_foreground_mobject(s)`, `remove_foreground_mobject(s)`, and persistence across ordinary scene edits.

This branch started from master `29b5b159b1d5842b54c82124089160cd07001bc0`, after #1680. It is independent of the active B4/B5 and C2/C5 stacks.

## Design

`foreground_mobjects` is compatibility metadata only. It never becomes a renderer list, z-index authority, semantic scene mirror, or alternate painter-order implementation. Every actual ordering change still goes through the existing canonical Rust scene-membership batch.

Ordinary `Scene.add()` submits newly added wrappers followed by the current foreground wrappers in one membership batch, matching Manim's Cairo ordering rule while letting shared Rust retain family-aware restructuring and authoritative root order. Exact duplicate request wrappers are removed before crossing the typed boundary.

Foreground metadata is committed only after an authoritative add succeeds. Ordinary remove, clear, and bring-to-back retire foreground status after their Rust membership edit succeeds. Animation completion reconciliation also retires foreground metadata for wrappers that the shared Rust scene actually removed, preventing a later add from resurrecting a completed FadeOut/Uncreate target. Removing foreground status alone does not remove or reorder the object in the semantic scene. Group-aware metadata restructuring recursively promotes unaffected siblings rather than dropping an entire foreground group.

Keyed single-object adds remain supported: the public Scene validates the one keyed caller object while the internal membership adapter may carry already-bound foreground wrappers after it; only the first reservation receives the explicit key.

No renderer/runtime primitive, dependency, per-frame work, host-side scene graph, or new ordering state is added.

## Tests

Adds focused Python facade regressions for:
- foreground remaining last when later objects are added;
- stable multi-foreground order and re-adding a foreground wrapper;
- rollback of compatibility metadata when authoritative add fails;
- remove/clear/bring-to-back retiring foreground status;\n- completion reconciliation retiring foreground wrappers removed by animation cleanup;
- keyed single-object adds while foreground ordering is projected.

Normal exact-head CI is starting. No raster parity or full B6 completion is claimed yet.

## Still open

Sections/metadata, broader moving-camera/auto-zoom APIs, and ZoomedScene/compositing remain separate B6 work.


yongkyuns:
CI repair update: the first exact-head run exposed two test-integration defects, not Rust/runtime failures. The new facade tests used `object()` values even though `Scene.add()` correctly validates Mobject/Group leaves, and completion-binding unit fixtures use `SimpleNamespace` scene doubles without the new helper method. Both are fixed on new one-commit head `d7b89496cf67d800f9dbc5c9936ce1ad9281aa57`: foreground tests now use real detached `Circle` wrappers, and completion reconciliation calls the class helper only when foreground metadata exists. Rust unit, format/lint, browser-fast, architecture/layer and agent checks had already passed on the prior head; the downstream build failures all inherited the same Python preflight failure. Fresh exact-head CI is now running.

Issue #87, opened 2026-08-24
[Roadmap B5] Graph/DiGraph topology, layouts, and mutation


yongkyuns:
## Roadmap

Phase B5 Graph/DiGraph parent under #954. Architecture authority: `docs/architecture.md`.

## Goal

Support Graph/DiGraph as retained semantic topology plus local runtime dependencies—never a graph renderer primitive or Python per-frame edge-following engine.

## State/dependency contract

Topology (vertices/edges/keys/order/configuration) is authored semantic state. Current vertex transforms are effective Runtime state when driven by timeline/reactive/host/editor behavior.

Incident-edge following is therefore a **runtime dependency**: moving one vertex through an active driver updates that vertex and incident edge effective geometry/state locally without committing a semantic topology/property mutation every frame. Persistent topology/position edits use normal semantic mutation.

## Scope

- Graph/DiGraph construction from vertices/edges;
- custom vertex/edge mobjects/config;
- labels/partitions/root helpers where supported;
- explicit positions and later common author/edit-time layout algorithms;
- add/remove vertices/edges and persistent layout changes;
- effective edge following under live vertex motion;
- directed edges reuse B2/#77 Arrow semantics;
- stable topology/source identity under mutation/hot reload.

## Architecture

- topology is semantic family/dependency state associated with ordinary semantic children;
- execution lowering derives local incident-edge dependency relationships;
- active vertex motion propagates through effective runtime dependencies rather than Python updater loops or per-frame semantic mutation;
- topology edits use shared semantic transactions/local lowering;
- layout solving is author/edit-time work unless a specific native dynamic layout is explicitly designed;
- frontend keys adapt to semantic handles but do not own topology/edge-following semantics;
- renderer receives normal retained geometry/resources.

## Acceptance

- representative Graph/DiGraph explicit-position, movement and mutation cases work through shared semantics;
- one driven vertex dirties only its effective state plus incident-edge dependency state;
- no semantic revision is required merely because a timeline/reactive driver moves a vertex;
- adding/removing topology preserves unrelated semantic/execution/resource identity;
- directed edges reuse common Arrow behavior;
- layout randomness/seed behavior is explicit/deterministic where required;
- no Python frame updater/full-graph scan/graph-specific renderer primitive;
- promoted cases pass B7 qualification.

## Dependencies

B1/#74, B2/#77, B4/#83 and Phase A/C mutation/effective-state/locality machinery. #697 is the first implementation child.

[...]

PR #1688, opened 2026-09-20
B5: reconcile retained Graph topology identity core

yongkyuns:
## Scope

Advances #697 / #87 / #954 by reconciling the architecture-only retained Graph topology core from historical #866 onto current master.

This is deliberately topology-only:
- monotonic vertex/edge identities;
- deterministic insertion order;
- undirected edge-key normalization;
- per-vertex incident-edge indexes for O(degree) dependency lookup;
- local add/remove validation and identity-preservation tests.

It does **not** claim public Graph/DiGraph support and adds no renderer/runtime primitive, Python facade, layout engine, edge geometry, or per-frame updater. The next integration boundary remains semantic-family binding with ordinary vertex mobjects and Line/Arrow edges plus effective incident-edge following.

Historical #866 is ~2,745 master commits behind, so only its isolated topology module is rebuilt here rather than reviving stale history. Current-head CI is required; keep draft until qualified.


yongkyuns:
Review hardening on current head `26aee37e6b3df19650ac3ff5b88d919fe9258ca2`: monotonic vertex/edge ID allocation now uses checked increments and returns explicit `VertexIdExhausted` / `EdgeIdExhausted` before mutating topology. Added regression coverage proving exhaustion leaves vertex/edge and incident indexes unchanged. This prevents release-mode wraparound from aliasing retained identities. API change is confined to this new unlanded topology core (`add_vertex` now returns Result); no existing public Graph facade exists yet. Fresh current-head CI required.

yongkyuns:
First current-head Fast Gate exposed a reconstruction-only crate-root mistake: the new `pub use graph_topology::*` had been inserted between the crate's inner `//!` documentation lines, so Rust parsed the remaining crate docs/attribute as invalid outer documentation (E0753). That single parse failure cascaded into Rust tests and browser compilation; it was not a topology test failure. Current head `a90c5e04397424558f2e12d7742e9c7be3b06ff0` moves the export beside the `mod graph_topology` declaration after crate attributes. Fresh CI required.

yongkyuns:
Follow-up self-review on `a90c5e04` found the only remaining obvious rustfmt delta in the new module: missing blank separation before the exhaustion test and a trailing blank before the test module close. Current head `97de3d7d8e623083175a05931460262a140f5c35` applies that formatting only; topology behavior is unchanged. Fresh current-head CI required.

yongkyuns:
Architecture gate on `97de3d7d` reached the actual ratchet and rejected the historical type name `RetainedGraphTopology`: that exact symbol is intentionally banned as a retired pre-consolidation model by `scripts/architecture_retired_models.py`. The module itself is semantic-authoring topology data, not a scene/runtime wrapper, so the correct reconciliation is to adopt the current architecture vocabulary rather than weaken the ratchet. Current head `f62f3c98e1efe0dfd10c61a628a9cff57b04fd34` renames the new authority to `GraphTopology` throughout; behavior/tests are otherwise unchanged. Fresh CI required.

yongkyuns:
Core qualification is now complete on exact head `f62f3c98e1efe0dfd10c61a628a9cff57b04fd34`: PR Fast Gate, architecture ratchets, layer-dependency ratchet, Native Host, Product Gate, plotting, playback and image qualification all passed. The architecture guardrail specifically accepts the reconciled `GraphTopology` authority after rejecting the historical retired type name. Remaining provider/cross-browser jobs are queued and do not exercise this Rust-only topology module; PR is mergeable and is now marked ready for review. The dependent semantic-binding branch remains pinned to this exact head.

yongkyuns:
Architecture-review scalability follow-up: incident-edge lookup is O(degree), but vertex deletion is not local. remove_vertex calls remove_edge for every incident edge; each remove_edge does Vec::remove plus suffix reindexing, yielding quadratic work for a high-degree star (10k central edges implies ~49,995,000 position rewrites before other movement). Please batch the removal set and compact/reindex edge storage once while preserving deterministic edge order. Add a high-degree regression/operation-count or benchmark-style qualification that distinguishes one-pass compaction from repeated suffix rewrites. Do not use swap_remove if it violates deterministic iteration order.

PR #1687, opened 2026-09-20
B6: add camera auto-frame target authoring

yongkyuns:
## Scope

Advances B6.4 / #89 under #954 with a bounded shared 2D camera auto-frame primitive.

Pinned base: `8b3286475869960e59928372cfb47662f0a28b11`. Exact head: `ce69b5dc14c3d40066f09396dd312e7ae5d383f7`. One commit / four files, independent of #1686 section metadata and active B4/B5/C2/C5 work.

## Design

`Scene::camera_auto_frame` observes existing authored boundary bounds once, unions the supplied objects, applies an explicit world-space margin, and expands the fitted height as needed to preserve the requested viewport aspect. It is a pure authored query: no semantic mutation, renderer state, execution scan, or per-frame work.

`Scene::camera_auto_frame_target` validates the ordinary semantic Camera2D frame, computes the same fit, and creates a normal detached target editor with translation plus uniform scale. The caller then uses the existing TransformTo/animate lifecycle; there is no camera-specific animation path.

Validation is typed through `AuthoringError::InvalidCameraAutoFrame`. Empty targets, invalid aspect/margin, missing/non-finite bounds, unrepresentable center/height, foreign handles, and non-camera target frames fail explicitly.

## Qualification

Four Rust integration tests cover:
- transformed multi-object union, margin, and 16:9 fit;
- width-dominated aspect expansion;
- pure fit observation leaving revision/camera/source state unchanged;
- invalid and foreign inputs failing before mutation.

Exact-head CI is starting.

## Deliberate limits

This is authored/static auto-framing, not reactive camera tracking. It does not add ZoomedScene/compositing, 3D camera behavior, renderer-side camera ownership, Python facade, or full B6 completion.

PR #1686, opened 2026-09-20
B6: add Scene section timeline metadata

yongkyuns:
## Scope

Advances B6.2 / #89 under #954 with shared authored section timeline metadata and the Manim-facing `Scene.next_section(name=..., type=...)` surface.

Pinned base: `8b3286475869960e59928372cfb47662f0a28b11`. Exact head: `750a60ac89ea31b2e295a876f8855cb039de38e3`. One commit, independent of active B4/B5 and C2/C5 stacks.

## Design

Rust `Scene` owns an ordered `SceneSection` list beside its existing authored cursor. A section carries name, `Normal`/`Skip` intent, and authored time. Repeated declarations at the same time remain distinct and ordered.

Section declarations are intentionally non-semantic/non-rendering metadata. They do not create semantic nodes, advance revisions, dirty execution objects, relower geometry, or introduce a second clock. Before execution they use `Scene::time()`; after bootstrap the canonical host synchronizes the Scene metadata cursor to the local execution handoff duration before recording the boundary. Declaration is rejected while execution ownership is transferred.

Python exposes `SectionType.NORMAL`, `SectionType.SKIP`, `Scene.next_section`, and read-only projected `Scene.sections`. The host serializes only this metadata projection; Python owns no duplicate timeline.

Names reject newline/CR/NUL before mutation. Failed declarations do not change section metadata or authored time.

## Qualification added

Rust integration tests cover:
- ordered same-time declarations and later authored time;
- atomic invalid-name rejection without clock/revision mutation;
- no semantic revision, publication-context, frame, or dirty-state change when declaring a section after execution bootstrap.

Normal exact-head CI is starting.

## Deliberate limits

This does not implement video splitting/export orchestration, presentation controls, renderer behavior, camera/auto-zoom, or ZoomedScene/compositing. It does not claim full B6 completion or pixel parity.

PR #1689, opened 2026-09-20
B6: define bounded secondary viewport composition descriptor

yongkyuns:
## Scope

Advances the B6 ZoomedScene/compositing foundation under #89 with only the renderer-neutral descriptor boundary needed before any second-view encoding exists.

Pinned base: `8b3286475869960e59928372cfb47662f0a28b11`. Exact head: `86d619e5faac753e00cde6220b6f1da43443f5f7`. One commit / two files. It does not overlap #1686 section metadata, #1687 camera auto-frame, or active B4/B5/C2/C5 work.

## Design boundary

`SecondaryViewport` is a validated value containing:
- an ordinary existing `Camera2D`;
- one physical-pixel destination rectangle.

It deliberately owns no Scene, semantic store, ExecutionSession, FrameState, PreparedFrame, GPU buffers, texture, renderer, clock, or input state. This makes the intended next renderer step explicit: encode another bounded projection of the same retained prepared frame/resources, rather than instantiate a second renderer/runtime authority.

Destination validation rejects empty rectangles, checked-add overflow, and output overflow before any GPU work.

## Qualification

Two renderer-crate tests cover a valid camera/destination descriptor plus empty, overflowing and out-of-output rectangles.

Exact-head CI is starting.

## Deliberate limits

This PR does **not** render a second viewport yet and makes no ZoomedScene, raster, WebGPU/WebGL parity, or Manim support claim. It intentionally stops at the ownership boundary so the subsequent encoding slice can be reviewed independently for retained-resource reuse and presentation correctness.


yongkyuns:
Qualification status: all completed correctness/architecture gates are green (Fast Gate, plotting, native, image raster, arrow parity, authoring recovery, architecture/layer). Firefox and Chromium full-gallery controls passed; WebKit is still running. Product Gate alone sampled effective FPS 29.4 vs baseline 37.7 / floor 30.2. This PR adds only a passive value descriptor and validation with no render/runtime call site, so there is no executable hot-path change that explains the sample; nevertheless I am leaving the performance failure visible and will not promote until the remaining matrix completes and we decide whether to rerun/compare the performance gate.

PR #1690, opened 2026-09-20
B6: encode retained secondary analytic viewport

yongkyuns:
## Scope

Stacked follow-up to #1689. Adds the first actual secondary-view encoding while preserving the ownership boundary established there.

Exact head: `40add0a849fd7a9264269995ad7450c95db757e8`. One commit on #1689; two production files changed.

## Architecture

The existing `GpuRenderer` gains one small secondary camera uniform/bind group. Secondary encoding:
- consumes the same already-prepared `PreparedFrame`;
- reuses all existing resident stable geometry buffers, transient-presentation buffers, and render pipelines;
- performs no upload, reprepare, semantic/runtime mutation, or second renderer construction;
- loads the existing scene target and clips drawing to the validated physical-pixel destination with viewport + scissor;
- binds the independent secondary camera uniform so primary-camera commands are not affected by queue-write ordering.

The ordinary `draw_ordered` and transient-presentation draw paths now delegate to camera-bind-group-parametrized helpers; primary behavior remains the same. Secondary encoding has a matching entry point for already-uploaded analytic transient presentations, preserving the same painter-splice authority rather than dropping animation-local occurrences.

## Explicit fail-closed boundary

The current secondary pass accepts only the single-sample analytic lane. Vector-path content requires a destination-sized MSAA target/resolve contract, so it returns `MultisampledContentUnsupported` instead of silently drawing with incorrect sampling. That follow-up remains separate.

## Deliberate limits

No ZoomedScene/Python facade, border/frame decoration, image/text special handling, vector-path MSAA, presentation-transfer finalization, or Manim parity claim yet. This is only the retained analytic composition primitive.

Exact-head CI is starting.


yongkyuns:
Source review found one failure-atomicity issue: the secondary camera uniform write occurred before rejecting unsupported path/MSAA content. That would leave renderer-owned secondary state changed even though encoding returned `Err`. Validation now completes before any queue write, so rejected secondary encodes are side-effect free. Re-flattened one-commit head: `9228589bf66f9e868101f64fde88c2ec47bd2e2b`; fresh exact-head CI is starting.

PR #1691, opened 2026-09-20
B6: bound secondary viewport MSAA ownership

yongkyuns:
## Scope

Stacked follow-up to #1690. Investigates the retained secondary viewport vector-path/MSAA lane and records the safe fail-closed boundary until composition can preserve primary pixels outside the secondary destination.

Exact head: `50c2a04736cbfbc219f7e52b63036d792f0ed920`. One commit / one production file.

## Design

The same `GpuRenderer`, `PreparedFrame`, stable GPU buffers and transient-presentation buffers remain authoritative. A second renderer or geometry upload is not introduced.

The renderer retains one additional full-output-size secondary MSAA scratch target, recreated only with ordinary viewport resize. A destination-sized target cannot directly resolve into the full scene target, while a full-size resolve would overwrite pixels outside the scissored destination unless the scratch target is first seeded with the existing primary scene.

For now, path/MSAA secondary content remains explicitly rejected before rendering. Analytic-only secondary rendering continues to use #1690. This avoids publishing a composition path that can erase the primary scene outside the destination.

## Important correction during implementation

Two unsafe approaches were rejected during review: destination-sized MSAA cannot resolve directly into the full scene target; full-size MSAA with a clear+resolve overwrites the entire resolve attachment, including primary pixels outside the scissor. The remaining implementation deliberately fails closed until a seed/copy or equivalent compositing step is added.

## Deliberate limits

No vector-path support claim, ZoomedScene/Python facade, border decoration, independent scene/session, or Manim parity claim. Exact-head CI will qualify only this ownership/fail-closed boundary.


yongkyuns:
CI repair: after the stack rebase, the MSAA constructor/resize assignments survived but the two corresponding `GpuRenderer` scratch fields were accidentally lost, causing compile failures. Restored `secondary_path_msaa_texture` / `secondary_path_msaa_view` and re-flattened directly on current #1690. New exact head `8edb3f0f1aa5a4c1de7122f64b0db7eb4ba90a3b`; fresh CI will run.

yongkyuns:
Architecture-review blockers for the secondary viewport stack: (1) current head initializes/uses secondary_path_msaa_texture/view without declaring those fields on GpuRenderer; fix the build before extending the path. (2) More importantly, one shared secondary_camera_buffer is queue-written per encode. Multiple secondary passes encoded before one submission can all observe the last write; a later rejected operation can also overwrite the camera for an earlier valid encoded pass because the write currently happens before the unsupported-MSAA return. Validate/reject before any queue write and use a submission-safe camera-uniform strategy (e.g. distinct immutable/dynamic-offset regions), or explicitly enforce a single-secondary-per-submission contract. Add actual GPU regressions for two distinct secondary cameras in one submission and valid A -> rejected B -> submit A. (3) Do not retain a full-output secondary MSAA allocation while the path only rejects MSAA unless there is an executable consumer.

yongkyuns:
Re-review of new head 285a5be7: the missing GpuRenderer fields are now declared, but the substantive blockers remain. The secondary camera queue.write_buffer still occurs before the multisampled rejection, so a rejected B can mutate the camera buffer referenced by an already-encoded valid A; multiple secondary views before one submission still share one mutable uniform. Also the full-output secondary MSAA texture/view are now allocated/recreated but still have no rendering consumer because multisampled content immediately returns Unsupported. Please move all rejection before writes, make camera data submission-safe (or enforce one-secondary-per-submission explicitly), add the two GPU sequence regressions, and remove/defer unused MSAA allocation until consumed. Keeping draft is correct.

PR #1692, opened 2026-09-20
B6: retain a sampleable scene target on every backend

yongkyuns:
Closed after architecture review. A universal intermediate scene target would impose an extra full-frame texture and presentation pass on ordinary Direct-mode rendering even when no secondary composition is used. The secondary-view stack now keeps Direct output on its existing fast path and handles presentation at the secondary composition boundary instead. This avoids making a global renderer cost a prerequisite for the bounded B6 feature.


yongkyuns:
Architecture-review composition/performance follow-up: please treat this together with #1690/#1691 rather than as an isolated renderer prerequisite. The universal intermediate scene texture + final fullscreen pass makes Direct-mode frames pay an extra full-size texture/pass even when no secondary composition is used (~31.6 MiB for one RGBA8 3840x2160 texture, before other costs). Either preserve a direct fast path when composition is unnecessary or provide explicit whole-frame performance/raster evidence supporting the universal path. Also consolidate frame finalization so ordinary scene/transients, secondary views, overlays, then exactly one final presentation have one owner; callers should not need to reconstruct encode ordering. If PresentationBridge resources are now invariantly present, remove Option/assert churn rather than preserving a dead optional representation.

